PR #10, opened 2023-02-28
Add sweeper utility

andrewkmin:
Simple script to sweep ones funds to a new address.

### DIFF
--- a/examples/sweeper/src/scripts/index.ts
+++ b/examples/sweeper/src/scripts/index.ts
@@ -1,0 +1,152 @@
+import * as path from "path";
+import * as dotenv from "dotenv";
+
+// Load environment variables from `.env.local`
+dotenv.config({ path: path.resolve(process.cwd(), ".env.local") });
+
+import { ethers } from "ethers";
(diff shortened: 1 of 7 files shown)


examples/sweeper/src/scripts/index.ts
```diff
@@ -0,0 +1,152 @@
...
+import { toReadableAmount } from "../utils";
+import { getProvider, getTurnkeySigner } from "../provider";
+import {
+  ERC20_ABI,
+  UNI_TOKEN_GOERLI,
+  USDC_TOKEN_GOERLI,
+  WETH_TOKEN_GOERLI,
+} from "../utils";
+
+const prompts = require("prompts");
```
keyz, line 18: Need `prompts` in package.json, and `@types/prompts` as a devDependency

keyz: Then you can convert this to an `import`

andrewkmin: 💯 thanks

keyz (approved):

PR #22, opened 2023-04-07
`signMessage(...)`: move encoding and hashing logic to client side

keyz-tk:
## Summary & Motivation
For raw payload signing, our configuration is too inflexible -- currently we require the message to be UTF8 encoded, and we hash the payload via `KECCAK256` on the remote signer. 

This PR changes `signMessage(...)` to conform to what `Ethers` expects, `eth_sign` style:
- The method accepts either a UTF8-message (`string`) or a binary message (`Bytes`)
- Hashing is done in JS land and conforms to [EIP-191](https://eips.ethereum.org/EIPS/eip-191) (see https://docs.ethers.org/v5/api/utils/hashing/#utils-hashMessage)
- Remote signer always receive hex-encoded payload (`PAYLOAD_ENCODING_HEXADECIMAL`), and does not apply extra hashing (`HASH_FUNCTION_NO_OP`)

## How I Tested These Changes
CI and running internal E2E tests manually

### DIFF
--- a/packages/ethers/src/__tests__/index-test.ts
+++ b/packages/ethers/src/__tests__/index-test.ts
@@ -102,37 +102,12 @@ test("TurnkeySigner", async () => {
     `);
   }
 
-  // Test message (raw payload) signing
+  // Test message (raw payload) signing, `eth_sign` style
   const message = "Hello Turnkey";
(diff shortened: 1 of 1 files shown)


r-n-o (approved):
Neat!

PR #23, opened 2023-04-11
`@turnkey/http`: add `withAsyncPolling(...)` helper

keyz-tk:
## Summary & Motivation

All Turnkey mutation endpoints are asynchronous (with the exception of signing endpoints) because we support consensus via policy on all write actions. To simplify async mutations, this PR adds a `withAsyncPolling(...)` wrapper. Here's a quick example:

```typescript
import {
  PublicApiService,
  withAsyncPolling,
  TurnkeyActivityError,
} from "@turnkey/http";

// Use `withAsyncPolling(...)` to wrap & create a fetcher with built-in async polling support
const fetcher = withAsyncPolling({
  request: PublicApiService.postCreatePrivateKeys,
  refreshIntervalMs: 500,
});

// The fetcher remains fully typed. After submitting the request,
// it'll poll until the activity reaches a terminal state.
try {
  const activity = await fetcher({
    body: {
      /* ... */
    },
  });

  // Success!
  console.log(activity.result.createPrivateKeysResult?.privateKeyIds?.[0]);
} catch (error) {
  if (error instanceof TurnkeyActivityError) {
    // In case the activity is rejected, failed, or requires consensus,
    // a rich `TurnkeyActivityError` will be thrown. You can read from
    // `TurnkeyActivityError` to find out why the activity didn't succeed.
    //
    // For instance, if your activity requires consensus and doesn't have
    // enough approvals, you can get the `activityId` from `TurnkeyActivityError`,
    // store it somewhere, then re-fetch the activity via `.postGetActivity(...)`
    // when the required approvals/rejections are in place.
  }
}
```

### Note to reviewer
Feel free to review commit by commit (they are stacked). See changes in `createNewEthereumPrivateKey.ts` as a real word example.

## How I Tested These Changes
- CI is green
- Ran the `with-ethers` example locally

### DIFF
--- a/packages/http/README.md
+++ b/packages/http/README.md
@@ -39,7 +39,7 @@ The OpenAPI spec that generates all fetchers is also [included](./src/__generate
 
 ## `withAsyncPolling(...)` helper
 
-All Turnkey mutation endpoints are asynchronous (with the exception of signing endpoints) because we support consensus via policy on all write actions. To help you simplify async mutations, `@turnkey/http` provides a `withAsyncPolling(...)` wrapper. Here's a quick example:
+All Turnkey mutation endpoints are asynchronous (with the exception of signing endpoints). To help you simplify async mutations, `@turnkey/http` provides a `withAsyncPolling(...)` wrapper. Here's a quick example:
 
 ```typescript
 import {


examples/deployer/src/createNewEthereumPrivateKey.ts
```diff
@@ -14,12 +18,37 @@ export async function createNewEthereumPrivateKey() {
...
 
-  const privateKeyName = `ETH Key ${String(
-    Math.floor(Math.random() * 10000)
-  ).padStart(4, "0")}`;
+  // Use `withAsyncPolling` to handle async activity polling.
+  // In this example, it polls every 250ms until the activity reaches a terminal state.
+  const mutation = withAsyncPolling({
+    request: PublicApiService.postCreatePrivateKeys,
+    refreshIntervalMs: 250,
+  });
```
r-n-o, line 26: Might be worth mentioning in this example that the default is 500?

examples/with-gnosis/src/createNewEthereumPrivateKey.ts
```diff
@@ -14,12 +18,37 @@ export async function createNewEthereumPrivateKey() {
...
   );
 
-  const privateKeyName = `ETH Key ${String(
-    Math.floor(Math.random() * 10000)
-  ).padStart(4, "0")}`;
+  // Use `withAsyncPolling` to handle async activity polling.
+  // In this example, it polls every 250ms until the activity reaches a terminal state.
+  const mutation = withAsyncPolling({
+    request: PublicApiService.postCreatePrivateKeys,
+    refreshIntervalMs: 250,
```
r-n-o, line 25: I think one of these examples should omit the `refreshIntervalMs`, to show that this doesn't have to be specified.

packages/ethers/README.md
```diff
@@ -76,3 +76,13 @@ main().catch((error) => {
   process.exit(1);
 });
 ```
+
+## More Examples
```
r-n-o, line 80: ❤️

packages/http/README.md
```diff
@@ -30,3 +30,55 @@ const data = await PublicApiService.postGetWhoami({
...
+
+## HTTP fetchers
+
+`@turnkey/http` provides fully typed http fetchers for interacting with the Turnkey API. You can find all available methods [here](./src/__generated__/services/coordinator/public/v1/public_api.fetcher.ts). The types of input parameters and output responses are also exported for convenience.
+
+The OpenAPI spec that generates all fetchers is also [included](./src/__generated__/services/coordinator/public/v1/public_api.swagger.json) in the package.
+
+## `withAsyncPolling(...)` helper
+
+All Turnkey mutation endpoints are asynchronous (with the exception of signing endpoints) because we support consensus via policy on all write actions. To help you simplify async mutations, `@turnkey/http` provides a `withAsyncPolling(...)` wrapper. Here's a quick example:
```
r-n-o, line 42: `because we support consensus via policy on all write actions` is misleading.

We support consensus on signing actions yet this activity type is synchronous. Why? Organization data writes! We have a queue for writes, which forces us to be asynchronous (our updater picks up tasks from the queue, decoupled from end-user requests).

Signatures don't require data changes so they can be synchronous. Whereas all other actions require a change in organization data (create user, delete policy, etc etc)

packages/http/README.md
```diff
@@ -30,3 +30,55 @@ const data = await PublicApiService.postGetWhoami({
...
+import {
+  PublicApiService,
+  withAsyncPolling,
+  TurnkeyActivityError,
+} from "@turnkey/http";
+
+// Use `withAsyncPolling(...)` to wrap & create a fetcher with built-in async polling support
+const fetcher = withAsyncPolling({
+  request: PublicApiService.postCreatePrivateKeys,
+  refreshIntervalMs: 500,
```
r-n-o, line 54: again, a comment with the fact that this is the default could help.

packages/http/src/__tests__/async-test.ts
```diff
@@ -0,0 +1,330 @@
...
+            privateKeyTags: [],
+          },
+        ],
+      },
+      organizationId: "89881fc7-6ff3-4b43-b962-916698f8ff58",
+      timestampMs: String(Date.now()),
+    },
+  };
+
+test("`withAsyncPolling` should return data after activity competition", async () => {
```
r-n-o, line 44: you mean completion, not competition right? 🏆

r-n-o (commented):
Nice!!

packages/http/README.md
r-n-o: I'd say something along the lines of: `All Turnkey mutation endpoints are asynchronous because we process updates to organization data asynchronously, with a queuing system. Signing activities are a notable exception: they do not require organization data updates, hence they are synchronous.`

andrewkmin: > I'm gonna further simplify this because it's mostly implementation details (in our current system)

any more context on this? by current system, do you mean the current state of the SDK, or the turnkey backend?

PR #25, opened 2023-04-12
[easy] Explain the difference between two public packages

keyz-tk:
## Summary & Motivation
On npm you can only see a package's own README, so we need to make it more clear everywhere

## How I Tested These Changes
Rendered:
- https://github.com/tkhq/sdk/tree/keyan/readme/packages/ethers
- https://github.com/tkhq/sdk/tree/keyan/readme/packages/http
- https://github.com/tkhq/sdk/tree/keyan/readme/

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,10 +6,10 @@ API Docs: https://turnkey.readme.io/
 
 ## Packages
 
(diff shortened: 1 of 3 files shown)


andrewkmin (approved):
💯

PR #26, opened 2023-04-13
[easy] Switch from `node-fetch` to `isomorphic-unfetch`

keyz-tk:
## Summary & Motivation
Such that the packages can be used in non-node environments. We don't recommend using the SDK directly in browsers as-is today (because you'll be exposing your API key to a client), but this is an infra change that unlocks further improvements.

## How I Tested These Changes
- CI
- Ran a few examples
- Will publish a beta


andrewkmin (approved):
😎

PR #27, opened 2023-04-14
Switch from Node's `crypto` to `WebCrypto.subtle`

keyz-tk:
## Summary & Motivation
Spiritual successor to #26. See inline comments re: the _s u b t l e_ difference in signature encoding

## How I Tested These Changes
- CI
- Ran a few examples

### DIFF
--- a/packages/http/src/stamp.ts
+++ b/packages/http/src/stamp.ts
@@ -1,4 +1,4 @@
-import * as crypto from "crypto";
+import { webcrypto } from "crypto";
 
 // Specific byte-sequence for curve prime256v1 (DER encoding)
 const PRIVATE_KEY_PREFIX = Buffer.from(
@@ -25,14 +25,14 @@ export async function stamp(input: {
 
 async function importPrivateKey(
   privateKeyHex: string
-): Promise<crypto.webcrypto.CryptoKey> {
+): Promise<webcrypto.CryptoKey> {
   const privateKeyBuffer = Buffer.from(privateKeyHex, "hex");
   const privateKeyPkcs8Der = Buffer.concat([
     PRIVATE_KEY_PREFIX,
     privateKeyBuffer,
   ]);
 
-  return await crypto.webcrypto.subtle.importKey(
+  return await webcrypto.subtle.importKey(
     "pkcs8",
     privateKeyPkcs8Der,
     {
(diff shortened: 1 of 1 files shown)


MaxBlaushild (approved):

packages/http/src/stamp.ts
```diff
@@ -1,38 +1,137 @@
...
     privateKeyBuffer,
   ]);
 
-  const privateKeyObject = crypto.createPrivateKey({
-    type: "pkcs8",
-    format: "der",
-    key: privateKeyPkcs8Der,
-  });
+  return await webcrypto.subtle.importKey(
+    "pkcs8",
```
r-n-o, line 36: `webcrypto.subtle.importKey` supports `raw` as encoding. Can we use this instead of pkcs8? That'd be easier to deal with if it works!

packages/http/src/stamp.ts
```diff
@@ -1,38 +1,137 @@
...
+}
+
+/**
+ * `SubtleCrypto.sign(...)` outputs signature in IEEE P1363 format:
+ * - https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/sign#ecdsa
+ *
+ * Turnkey expects the signature encoding to be DER-encoded ASN.1:
+ * - https://github.com/tkhq/tkcli/blob/7f0159af5a73387ff050647180d1db4d3a3aa033/src/internal/apikey/apikey.go#L149
+ *
+ * Code modified from https://github.com/google/tink/blob/6f74b99a2bfe6677e3670799116a57268fd067fa/javascript/subtle/elliptic_curves.ts#L114
```
r-n-o, line 74: Oh man. Painful. Thanks for going through this 🎖️

packages/http/src/stamp.ts
```diff
@@ -1,38 +1,137 @@
...
+ *
+ * Transform an ECDSA signature in IEEE 1363 encoding to DER encoding.
+ *
+ * @param ieee the ECDSA signature in IEEE encoding
+ * @return ECDSA signature in DER encoding
+ */
+function convertEcdsaIeee1363ToDer(ieee: Uint8Array): Uint8Array {
```
r-n-o, line 81: I think some basic tests are in order for this function? I know it's indirectly tested by the high-level "signature is valid" test, but I'd feel better with a few unit tests given this function has pretty complex, hard-to-review-at-a-glance logic.

packages/http/src/stamp.ts
r-n-o: 😿 

Dang it. Well, thanks for checking!

PR #30, opened 2023-04-19
Make `@turnkey/http` available in browser runtimes

keyz-tk:
## Summary & Motivation
This PR does a few things:
- (easy) Switch from `isomorphic-unfetch` to ~~`cross-fetch`~~ `undici`. ~~In a later PR I might switch that to Node's official `undici` package such that we can drop all external dependencies~~ Done in https://github.com/tkhq/sdk/pull/30/commits/fd91a81819fcd8704b991a0f9d353786da949014. `undici` is an official Node.js library that's already included in Node v18; we're including it for v16.
- (easy) Change `webcrypto.subtle` and `TextEncoder` to be isomorphic imports, i.e. in Node.js it'll import from the standard library, and in browser runtimes it'll use the globally available version
- (hard) Change Turnkey's API format from PKCS#8 to PWK, because the DER format (that we currently have) doesn't get recognized by browsers. Because we expose the public key in compressed format, we need to restore the `x` and `y` -- this is done via vendoring google's `tink` library (`tink` doesn't have an up-to-date `npm` package)

Very helpful message from [Chrome](https://www.chromium.org/blink/webcrypto/#supported-key-formats):
> There are differences in DER key format handling between Web Crypto implementations. Where possible for compatibility prefer using "raw" keys or "jwk" which have better interoperability.

Will follow up with another PR that adds tests for the new utilities (and the ones in #27 and #29)

## How I Tested These Changes
- CI is green
- Ran a few examples

### DIFF
--- a/packages/http/src/stamp.ts
+++ b/packages/http/src/stamp.ts
@@ -4,7 +4,7 @@ import {
   hexStringToUint8Array,
   hexStringToBase64urlString,
 } from "./encoding";
-import { pointDecode, PointFormatType } from "./tink/elliptic_curves";
+import { pointDecode } from "./tink/elliptic_curves";
 
 export async function stamp(input: {
   content: string;
@@ -77,11 +77,7 @@ function convertTurnkeyApiKeyToJwk(input: {
 }): JsonWebKey {
   const { uncompressedPrivateKeyHex, compressedPublicKeyHex } = input;
 
-  const jwk = pointDecode(
-    "P-256",
-    PointFormatType.COMPRESSED,
-    hexStringToUint8Array(compressedPublicKeyHex)
-  );
+  const jwk = pointDecode(hexStringToUint8Array(compressedPublicKeyHex));
 
   jwk.d = hexStringToBase64urlString(uncompressedPrivateKeyHex);
 
--- a/packages/http/src/tink/elliptic_curves.ts
+++ b/packages/http/src/tink/elliptic_curves.ts
@@ -9,83 +9,24 @@
 import * as Bytes from "./bytes";
 
 /**
- * Supported elliptic curves.
+ * P-256 only
  */
-enum CurveType {
-  P256 = 1,
-  P384,
-  P521,
+function getModulus(): bigint {
+  // https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.186-4.pdf (Appendix D).
+  return BigInt(
+    "115792089210356248762697446949407573530086143415290314195533631308" +
+      "867097853951"
+  );
 }
 
 /**
- * Supported point format.
+ * P-256 only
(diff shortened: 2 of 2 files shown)


packages/http/src/stamp.ts
```diff
@@ -23,16 +26,20 @@ export async function stamp(input: {
   };
 }
 
-async function importPrivateKey(
-  privateKeyHex: string
-): Promise<webcrypto.CryptoKey> {
-  const privateKeyPkcs8Der = hexStringToUint8Array(
-    PRIVATE_KEY_PREFIX + privateKeyHex
-  );
+async function importPrivateKey(input: {
```
r-n-o, line 29: Now that this function takes both public and private key, I'd call it `importTurnkeyApiKey`

packages/http/src/stamp.ts
```diff
@@ -23,16 +26,20 @@ export async function stamp(input: {
...
+  uncompressedPrivateKeyHex: string;
+  compressedPublicKeyHex: string;
+}): Promise<CryptoKey> {
+  const { uncompressedPrivateKeyHex, compressedPublicKeyHex } = input;
+
+  const jwk = convertTurnkeyPrivateKeyToJwk({
```
r-n-o, line 35: Same here: `convertTurnkeyApiKeyToJwk`

packages/http/src/stamp.ts
```diff
@@ -62,6 +69,23 @@ async function signMessage(
   return uint8ArrayToHexString(signatureDer);
 }
 
+function convertTurnkeyPrivateKeyToJwk(input: {
+  uncompressedPrivateKeyHex: string;
+  compressedPublicKeyHex: string;
+}): JsonWebKey {
```
r-n-o, line 75: We should contemplate changing the API key format if jwk is easier to deal with. We could change the CLI so that it outputs new API keys in JWK, and have a feature to migrate from one format to another. We're still early enough that it'd be doable!

The checked-in `tink` dependency is definitely reasonable though. The only "bad" thing about it is it's a tiny bit too broad. It supports P256, P384, and P521 but we only need P256. Even with that, the size isn't bad and the code is readable so I'm not opposed to taking it on.

packages/http/src/isomorphic.ts
```diff
@@ -0,0 +1,9 @@
+/// <reference lib="dom" />
```
r-n-o, line 1: Probably a lack of knowledge on my part, but why is this file named "isomorphic"? It seems like it's a polyfill, or a list of namespaces that map to either web-browser-maintained packages, or node packages.

Is "isomorphic" the standard JS term for this? I don't know that there's a better name necessarily, I'm just curious.

r-n-o (approved):
👏

PR #35, opened 2023-04-21
Switch from `undici` to `cross-fetch` to improve compatibility

keyz-tk:
## Summary & Motivation

Apparently there are bundlers that don't like the `undici` dependency

If we run into problems again we can have a new separate `@turnkey/universal` package that provides different exports per environment, but for now it doesn't seem to be necessary yet 

## How I Tested These Changes
- CI
- Ran examples locally
- Tested with webpack

### DIFF
--- a/packages/ethers/CHANGELOG.md
+++ b/packages/ethers/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @turnkey/ethers
 
+## 0.8.1
+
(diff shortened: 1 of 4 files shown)


andrewkmin (approved):
> Apparently there are bundlers that don't like the undici dependency

Any in particular/how'd you discover this?

PR #37, opened 2023-04-28
Split stamping into two code paths: node / browser

keyz-tk:
## Summary & Motivation
The previous implementation was ok for supporting browsers and node runtimes, but it doesn't work for React Native.

Generally, when you use RN you need to provide polyfills for `node:crypto`, but no popular implementations support `crypto.webcrypto`.

Solution: let's separate the two code paths. If WebCrypto is available (likely in a browser, or Vercel edge runtime), we use `crypto.subtle`. If WebCrypto isn't available, stop trying to use `WebCrypto`; just use node's native `crypto` API.

## How I Tested These Changes
- CI
- Ran a few examples manually

### DIFF
--- a/packages/ethers/CHANGELOG.md
+++ b/packages/ethers/CHANGELOG.md
@@ -1,5 +1,16 @@
 # @turnkey/ethers
 
+## 0.9.0
+
+### Minor Changes
+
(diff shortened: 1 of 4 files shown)


MaxBlaushild (approved):

andrewkmin (approved):
💯

r-n-o (approved):

PR #39, opened 2023-04-28
add signTypedData to ethers implementation

andrewkmin:
## Summary & Motivation
- Add support for signing typed data (per EIP-712) to Turnkey Ethers implementation
- Update Gnosis example to make use of offchain signatures, with two flavors: (1) using signed typed data, and (2) using raw message signing, in addition to the standard onchain flow

## How I Tested These Changes
- Unit
- Testnet

Some helpful documentation/references:
- https://github.com/ethers-io/ethers.js/issues/830
- https://ethereum.stackexchange.com/questions/117631/as-in-web3-js-get-the-same-signature-as-the-signature-from-the-signtypeddata-v4
- Not relevant for JS/TS implementations, but noting some gotchas around encoding:
  - https://ethereum.stackexchange.com/questions/137491/inconsistency-in-recovered-account-when-verifying-eip-712-javascript-python
  - https://ethereum.stackexchange.com/questions/125425/solidity-cannot-verify-ethers-js-signed-data-signtypeddata

### DIFF
--- a/packages/ethers/src/__tests__/index-test.ts
+++ b/packages/ethers/src/__tests__/index-test.ts
@@ -104,10 +104,35 @@ test("TurnkeySigner", async () => {
 
   // Test message (raw payload) signing, `eth_sign` style
   const message = "Hello Turnkey";
-  const sig = await signer.signMessage(message);
+  const signMessageSignature = await signer.signMessage(message);
 
-  expect(sig).toMatch(/^0x/);
-  expect(ethers.utils.verifyMessage(message, sig)).toEqual(expectedEthAddress);
+  expect(signMessageSignature).toMatch(/^0x/);
+  expect(ethers.utils.verifyMessage(message, signMessageSignature)).toEqual(expectedEthAddress);
+
+  // Test typed data signing (EIP-712)
+  const typedData = {
+    types: {
+      // Note that we do not need to include `EIP712Domain` as a type here, as Ethers will automatically inject it for us
+      Person: [
+        { name: 'name', type: 'string' },
+        { name: 'wallet', type: 'address' },
+      ],
+    },
+    domain: {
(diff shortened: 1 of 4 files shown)


packages/ethers/src/index.ts
```diff
@@ -199,6 +234,37 @@ export class TurnkeySigner extends ethers.Signer {
...
+    return this._signMessageWithErrorWrapping(
+      ethers.utils._TypedDataEncoder.hash(
+        populated.domain,
+        types,
+        populated.value
+      )
+    );
+  }
+
+  _signTypedData = this.signTypedData.bind(this);
```
andrewkmin, line 267: While most clients use `signTypedData`, some (like Gnosis SDK) expect to see `_signTypedData` (see https://github.com/safe-global/safe-core-sdk/blob/7959821ab08c96cf3babb9ed906c01d644ac49f4/packages/protocol-kit/src/adapters/ethers/utils/index.ts#L21-L22)

packages/ethers/src/__tests__/index-test.ts
```diff
@@ -104,10 +104,48 @@ test("TurnkeySigner", async () => {
...
-  expect(ethers.utils.verifyMessage(message, sig)).toEqual(expectedEthAddress);
+  expect(signMessageSignature).toMatch(/^0x/);
+  expect(ethers.utils.verifyMessage(message, signMessageSignature)).toEqual(
+    expectedEthAddress
+  );
+
+  // Test typed data signing (EIP-712)
+  const typedData = {
+    types: {
+      // Note that we do not need to include `EIP712Domain` as a type here, as Ethers will automatically inject it for us
```
andrewkmin, line 117: Reference: https://github.com/ethers-io/ethers.js/discussions/3163

PR #50, opened 2023-05-12
Build tooling: enable TypeScript mono project references / composite build

keyz-tk:
## Summary & Motivation
Without using project references, we'll soon run into the infamous "double compilation needed" problem https://www.typescriptlang.org/docs/handbook/project-references.html#overall-structure

## How I Tested These Changes
CI

### DIFF
--- a/examples/deployer/package.json
+++ b/examples/deployer/package.json
@@ -5,6 +5,7 @@
   "scripts": {
     "start": "tsx src/index.ts",
     "compile": "tsx src/compile.ts",
+    "clean": "rimraf ./dist ./.cache",
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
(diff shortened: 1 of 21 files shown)


examples/deployer/package.json
```diff
@@ -3,8 +3,9 @@
   "version": "0.1.3",
   "private": true,
   "scripts": {
-    "start": "ts-node src/index.ts",
-    "compile": "ts-node src/compile.ts",
+    "start": "tsx src/index.ts",
+    "compile": "tsx src/compile.ts",
+    "clean": "rimraf ./dist ./.cache",
```
r-n-o, line 8: Nit: can we use `rm -rf` directly here? Why do we need `rimraf` as a dependency?

r-n-o (approved):
Looks good to me!

PR #80, opened 2023-06-22
add example with managed nonce

andrewkmin:
## Summary & Motivation
$title

A simple, low lift POC that can be used to demonstrate how to handle delayed or failed transactions.

## How I Tested These Changes
Ran the scripts!


MaxBlaushild (commented):
I like this in concept, but I think most people using this will need to persist the nonce in some sort of state, so it would be nifty if we could create some sort of state that would persist between invocations of the script, and abstract that storage piece out into its own adapter. It could be something super naive like writing to a file

andrewkmin:
> I like this in concept, but I think most people using this will need to persist the nonce in some sort of state, so it would be nifty if we could create some sort of state that would persist between invocations of the script, and abstract that storage piece out into its own adapter. It could be something super naive like writing to a file

I like the idea! Let me know what you think of https://github.com/tkhq/sdk/pull/80/commits/135b05eb31dca81d83ca1d763a32abc7c2705fe2. I broke it up into `initiate()` and `monitor()` so that users can just call `monitor()` if they already have transactions out in the wild.

MaxBlaushild:
I love it! 

> > I like this in concept, but I think most people using this will need to persist the nonce in some sort of state, so it would be nifty if we could create some sort of state that would persist between invocations of the script, and abstract that storage piece out into its own adapter. It could be something super naive like writing to a file
> 
> I like the idea! Let me know what you think of [135b05e](https://github.com/tkhq/sdk/commit/135b05eb31dca81d83ca1d763a32abc7c2705fe2). I broke it up into `initiate()` and `monitor()` so that users can just call `monitor()` if they already have transactions out in the wild.

I love it!

MaxBlaushild (approved):

PR #82, opened 2023-06-22
Implements federated request (WebAuthn Assertion) SDK

MaxBlaushild:
## Summary & Motivation

Propose a framework for exposing what I'm calling federated requests. These requests collect signatures from the browser and provide enough context for the client to proxy to the backend to completely the request.

To send the request, the backend service needs the url to send the request to, the stamp to put in the header and the body to post, so the SDK provides this wrapped in a `FederatedRequest` struct.

Clientside code would be like this:

```typescript
// this prompts the user to sign with their passkey
const federatedRequest = await TurnkeyApi.federatedPostCreatePrivateKey({
  body: ...
});

await myOwnApiClient.post('/proxy_turnkey_request', federatedRequest);
```

Then on the backend, things could be handled like this (example in express):

```typescript
app.post('/proxy_turnkey_request', async (req, res) => {
  await axios.post(req.body.url, req.body.body, {
    headers: {
      "X-Stamp-WebAuthn": req.body.stamp,
    }
  })
})
```


packages/http/src/webauthn.ts
```diff
@@ -0,0 +1,43 @@
+import type { definitions } from "./__generated__/services/coordinator/public/v1/public_api.types";
+import { get } from "@github/webauthn-json/browser-ponyfill";
+import { base64StringToBase64UrlEncodedString } from "./encoding";
+
+type TWebAuthnStamp = definitions["v1WebAuthnStamp"];
+
+const timeout = 5 * 60 * 1000;
```
r-n-o, line 7: That's a 5mins timeout?

packages/http/src/webauthn.ts
```diff
@@ -0,0 +1,43 @@
...
+async function getCredentialRequestOptions(
+  payload: string
+): Promise<CredentialRequestOptions> {
+  const challenge = await getChallengeFromPayload(payload);
+
+  return {
+    publicKey: {
+      challenge: await new TextEncoder().encode(challenge),
+      timeout,
+      userVerification: "discouraged",
```
r-n-o, line 18: This probably needs to be configurable. In the case of passkeys I don't think it hurts to have `userVerification` set to "preferred" or even "required". It's only in the case of yubikeys that UV is problematic (users may not want to go through the hassle of an extra pin code to remember)

packages/http/src/webauthn.ts
```diff
@@ -0,0 +1,43 @@
...
+async function getChallengeFromPayload(payload: string): Promise<string> {
+  const messageBuffer = new TextEncoder().encode(payload);
+  const hashBuffer = await crypto.subtle.digest("SHA-256", messageBuffer);
+  const base64String = Buffer.from(hashBuffer).toString("base64");
+  return base64StringToBase64UrlEncodedString(base64String);
+}
+
+export async function getWebAuthnAssertion(payload: string): Promise<string> {
+  const signingOptions = await getCredentialRequestOptions(payload);
+  const clientGetResult = await get(signingOptions);
```
r-n-o, line 32: nit: let's import this as "webauthnGet"? I had a brief moment of confusion reading this.

r-n-o (commented):
This looks good! I don't have any major comments aside from the missing "create" functionality (registration flow). It's important to write both `create` and `get` so we have confidence that things can work end-to-end.

The comments below are fine to ignore if you're still in the prototyping phase!

packages/http/src/webauthn.ts
MaxBlaushild: I was planning on adding an options param that allows users to add their own signing options. Do you think "preferred" is a sane default?

packages/http/src/webauthn.ts
MaxBlaushild: Yep! Timeouts are in milliseconds

README.md
```diff
@@ -14,15 +14,16 @@ API Docs: https://turnkey.readme.io/
...
+| [`with-nonce-manager`](/examples/with-nonce-manager/)                       | Create a new Ethereum address, then sign and broadcast multiple transactions in a sequential or optimistic manner. |
+| [`sweeper`](/examples/sweeper/)                                             | Sweep funds from one address to a different address                                                                |
+| [`deployer`](/examples/deployer/)                                           | Compile and deploy a smart contract                                                                                |
+| [`with-offline`](/examples/with-offline/)                                   | Sign a Turnkey request in offline context                                                                          |
+| [`demo-consumer-wallet`](https://github.com/tkhq/demo-consumer-wallet)      | A minimal consumer wallet app powered by Turnkey and WalletConnect                                                 |
+| [`with-federated-authentication`](/examples/with-federated-authentication/) | A nextjs app that demonstrates how to use Turnkey to build a federated, webauthn powered authentication flow       |
```
r-n-o, line 29: Nit: let's rename this to `with-federated-passkeys` or `with-end-user-passkeys`? It'll be more concrete. When I read "federated authentication" I don't really have a good sense of what this example is about.

examples/with-federated-authentication/README.md
```diff
@@ -0,0 +1,49 @@
...
+```
+
+Now open `.env.local` and add the missing environment variables:
+
+- `API_PUBLIC_KEY`
+- `API_PRIVATE_KEY`
+- `BASE_URL`
+- `ORGANIZATION_ID`
+- `PRIVATE_KEY_ID` -- if you leave it blank, we'll create one for you via calling the Turnkey API
+- `INFURA_KEY` -- if this is not set, it will default to using the Community Infura key
```
r-n-o, line 41: I don't think this is relevant for this example?

examples/with-federated-authentication/README.md
```diff
@@ -0,0 +1,49 @@
...
+$ cd examples/with-federated-authentication/
+```
+
+### 2/ Setting up Turnkey
+
+The first step is to set up your Turnkey organization and account. By following the [Quickstart](https://turnkey.readme.io/docs/quickstart) guide, you should have:
+
+- A public/private API key pair for Turnkey
+- An organization ID
+- A (crypto) private key ID
```
r-n-o, line 26: let's remove this

examples/with-federated-authentication/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+Once you've gathered these values, add them to a new `.env.local` file. Notice that your private key should be securely managed and **_never_** be committed to git.
```
r-n-o, line 28: Just to make sure there's no ambiguity: "Notice that your API private key should be...etc..." (to minimize confusion with crypto private keys)

examples/with-federated-authentication/src/pages/_document.tsx
```diff
@@ -0,0 +1,19 @@
+import Document, { Html, Head, Main, NextScript } from "next/document";
+
+class MyDocument extends Document {
```
r-n-o, line 3: Extremely minor: rename this to "TurnkeyDemo" or something custom?

examples/with-federated-authentication/src/pages/api/subOrg.ts
```diff
@@ -0,0 +1,85 @@
...
+  message: string;
+};
+
+httpInit({
+  apiPublicKey: process.env.API_PUBLIC_KEY!,
+  apiPrivateKey: process.env.API_PRIVATE_KEY!,
+  baseUrl: process.env.BASE_URL!,
+});
+
+export default async function createUser(
```
r-n-o, line 33: let's rename this: `createSubOrganization`

examples/with-federated-authentication/src/pages/api/subOrg.ts
```diff
@@ -0,0 +1,85 @@
...
+  req: NextApiRequest,
+  res: NextApiResponse<CreateSubOrgResponse | ErrorMessage>
+) {
+  const createUserRequest = req.body as CreateSubOrgRequest;
```
r-n-o, line 37: same here: `createSubOrganizationRequest`

examples/with-federated-authentication/src/pages/api/subOrg.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  const createUserMutation = withAsyncPolling({
```
r-n-o, line 39: same here: `createSubOrganizationMutation`

examples/with-federated-authentication/src/pages/api/subOrg.ts
```diff
@@ -0,0 +1,85 @@
...
+    request: TurnkeyApi.postCreateSubOrganization,
+  });
+
+  try {
+    const createUserActivity = await createUserMutation({
```
r-n-o, line 44: and here: `createSubOrganizationActivity` / `createSubOrganizationMutation`

examples/with-federated-authentication/src/pages/index.module.css
```diff
@@ -0,0 +1,221 @@
+@font-face {
```
r-n-o, line 1: this file feels really long for the content of this demo. Are you sure we need all these fonts and all the variants? If not let's trim to only what is used.

examples/with-federated-authentication/src/pages/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+type subOrgFormData = {
+  orgName: string;
+};
+
+type privateKeyFormData = {
+  privateKeyName: string;
+};
+
+const es256 = -7;
```
r-n-o, line 17: this probably deserves a comment + a link to the registry which defines the algorithm constants (IANA Cose identifier): https://www.iana.org/assignments/cose/cose.xhtml

examples/with-federated-authentication/src/pages/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+const publicKey = "public-key";
```
r-n-o, line 18: Likewise. A link to the spec + a mention of the fact that we're always creating public key credential would be nice. The reason why it's good to be verbose here is that this example is meant to be copy/pasted; the more explicit we are on the "why" the better.

examples/with-federated-authentication/src/pages/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+  const { register: subOrgFormRegister, handleSubmit: subOrgFormSubmit } =
+    useForm<subOrgFormData>();
+  const {
+    register: privateKeyFormRegister,
+    handleSubmit: privateKeyFormSubmit,
+  } = useForm<privateKeyFormData>();
+
+  const createPrivateKey = async (data: privateKeyFormData) => {
+    if (!subOrgId) {
+      throw new Error("sub org id id not found");
```
r-n-o, line 37: "sub-org id" (one too many "id")

examples/with-federated-authentication/src/pages/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  const createSubOrg = async (data: subOrgFormData) => {
+    const challenge = generateRandomBuffer();
+    const authenticatorUserId = generateRandomBuffer();
+
+    const attestation = await getWebAuthnAttestation({
+      publicKey: {
+        rp: {
+          id: "localhost",
+          name: "Turnkey WebaAuthn Demo",
```
r-n-o, line 71: WebAuthn. Maybe reuse the example name here? `Turnkey Federated Passkey Demo` for example.

examples/with-federated-authentication/src/pages/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+        pubKeyCredParams: [
+          {
+            type: publicKey,
+            alg: es256,
+          },
+        ],
+        user: {
+          id: authenticatorUserId,
+          name: data.orgName,
+          displayName: data.orgName,
```
r-n-o, line 83: Why is this "orgName"? Shouldn't this be "subOrgName"?

examples/with-federated-authentication/src/pages/index.tsx
```diff
@@ -0,0 +1,162 @@
...
+    const res = await axios.post("/api/subOrg", {
+      orgName: data.orgName,
+      attestation,
+      // The challenge must be base64 url encoded
+      challenge: Buffer.from(challenge)
+        .toString("base64")
+        .replace(/\+/g, "-")
+        .replace(/\//g, "_")
+        .replace(/=/g, ""),
+    });
```
r-n-o, line 97: let's extract this into a function for readability: `challenge: base64urlencode(challenge)`, and then you can remove the comment because what's going on is pretty clear.

packages/http/src/config.ts
```diff
@@ -36,9 +36,9 @@ export function init(value: TConfig): void {
 
 export function getConfig(): TConfig {
   return {
-    apiPublicKey: assertNonEmptyString(config.apiPublicKey, "apiPublicKey"),
-    apiPrivateKey: assertNonEmptyString(config.apiPrivateKey, "apiPrivateKey"),
-    baseUrl: assertNonEmptyString(config.baseUrl, "baseUrl"),
+    apiPublicKey: config.apiPublicKey || "",
+    apiPrivateKey: config.apiPrivateKey || "",
+    baseUrl: config.baseUrl || "https://coordinator-beta.turnkey.io",
```
r-n-o, line 41: let's add the `assertNonEmptyString` back? Seems useful.

I'm also not a fan of hardcoding `https://coordinator-beta.turnkey.io` at the core of our SDK. We may change domains or subdomains at some point.

packages/http/src/shared.ts
```diff
@@ -64,3 +64,9 @@ export class TurnkeyRequestError extends Error {
     this.code = input.code;
   }
 }
+
+export type FederatedRequest = {
```
r-n-o, line 68: Let's insert a comment at the top to explain why this might be useful

packages/http/src/webauthn-json/convert.ts
```diff
@@ -0,0 +1,80 @@
+// We export these values in order so that they can be used to deduplicate
+// schema definitions in minified JS code.
+
+import type { Schema, SchemaProperty } from "./schema-format";
+
+// TODO: Parcel isn't deduplicating these values.
+export const copyValue = "copy";
+export const convertValue = "convert";
```
r-n-o, line 8: I'm not sure I understand what's going on here

MaxBlaushild: This is vendored code, so I'm not sure either! I'll move it to a vendor folder to make that explicit.

packages/http/src/config.ts
MaxBlaushild: Fair enough. I'll figure out a good pattern for initializing this in the browser

packages/http/src/webauthn.ts
```diff
@@ -0,0 +1,81 @@
+import type { definitions } from "./__generated__/services/coordinator/public/v1/public_api.types";
+import { base64StringToBase64UrlEncodedString } from "./encoding";
+
+type TWebAuthnStamp = definitions["v1WebAuthnStamp"];
+
+const defaultTimeout = 5 * 60 * 1000; // five minutes
+const defaultUserVerification = "preferred";
+
+// We generate challenge, so user supplies everything else
+export type TurnkeyPublicKeyCredentialRequestOptions = {
```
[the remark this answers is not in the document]
MaxBlaushild: I think we need some sort of prefix, since the unfixed names already exist on the window, which could lead to some nasty, unintentional shadowing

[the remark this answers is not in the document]
timurnkey: ohhh! interesting! could we hook into `window.Turnkey`? not necessary... just exploring ways we could namespace here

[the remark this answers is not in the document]
MaxBlaushild: I don't think we need to attach anything to the window ourselves, but definitely interesting to think about

examples/with-federated-passkeys/README.md
```diff
@@ -0,0 +1,45 @@
+# Example: `with-federated-passkeys`
+
+This example shows how to create a sub organization and faciliate creating a private key for the user's sub organization with Turnkey.
```
r-n-o, line 3: facilitate

examples/with-federated-passkeys/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+```bash
+$ cp .env.local.example .env.local
+```
+
+Now open `.env.local` and add the missing environment variables:
+
+- `API_PUBLIC_KEY`
+- `API_PRIVATE_KEY`
+- `BASE_URL`
```
r-n-o, line 37: `ORGANIZATION_ID` is missing from this list

examples/with-federated-passkeys/src/pages/_document.tsx
```diff
@@ -0,0 +1,19 @@
+import Document, { Html, Head, Main, NextScript } from "next/document";
+
+class MyDocument extends Document {
```
r-n-o, line 3: I think I commented on this previously, but it's a minor thing: let's rename this "FederatedPasskeysDemo", "TurnkeyDemo" or something similar?

examples/with-federated-passkeys/src/pages/api/subOrg.ts
```diff
@@ -0,0 +1,85 @@
...
+  res: NextApiResponse<CreateSubOrgResponse | ErrorMessage>
+) {
+  const createUserRequest = req.body as CreateSubOrgRequest;
+
+  const createUserMutation = withAsyncPolling({
+    request: TurnkeyApi.postCreateSubOrganization,
+  });
+
+  try {
+    const createUserActivity = await createUserMutation({
```
r-n-o, line 44: All of these should be renamed to "createSubOrganization" instead of "createUser"

r-n-o (commented):
Looking good!

Re: vendored webauthn-json dep: I think it's worth adding a README (`packages/http/src/webauthn-json/README.md`) with a quick explainer covering:
* how these files were fetched (which version/tag/commit sha)
* how upgrade this dependency in the future, when needed

r-n-o (approved):
👏

PR #87, opened 2023-07-21
Update types for http package

emostov:
## Summary & Motivation

Updates the types for the http package.

### DIFF
--- a/.changeset/late-pumas-buy.md
+++ b/.changeset/late-pumas-buy.md
@@ -1,0 +1,5 @@
+---
+"@turnkey/http": minor
+---
+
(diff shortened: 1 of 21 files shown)


.changeset/late-pumas-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@turnkey/http": minor
```
andrewkmin, line 2: I don't think we need to commit this changeset unless we're doing a version bump on one of the packages. The other 3 file updates look great tho!

timurnkey (approved):

PR #90, opened 2023-07-28
Version bump and CHANGELOG

emostov:
## Summary & Motivation

Sync public types


examples/rebalancer/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# @turnkey/rebalancer
+
+## 0.0.2
+
+### Patch Changes
+
+- Updated dependencies [46473ec]
+- Updated dependencies [38b424f]
+  - @turnkey/http@1.0.0
+  - @turnkey/ethers@0.16.4
```
r-n-o, line 10: I don't think we should have a changelog for an example 🤔

r-n-o (approved):

examples/rebalancer/CHANGELOG.md
r-n-o: The reason why is package name:
https://github.com/tkhq/sdk/blob/f72e9fd40ef1bfe766eb1fd403e4ffbeaaada742/examples/rebalancer/package.json#L2

This should be `@turnkey/example-with-rebalancer`. Then the changelog / release tooling will ignore.

emostov: fixed in [c68a008](https://github.com/tkhq/sdk/pull/90/commits/c68a0080366eb92843f986c531a0b49baf2c2eee)

PR #93, opened 2023-08-07
New viem package + example

r-n-o:
## Summary & Motivation

This PR introduces a new `@turnkey/viem` package with (for now) a single helper function to create a Viem Custom Account backed by a Turnkey API key. I'll work on making a helper to work with passkeys next, this is a warmup!

The associated `with-viem` example demonstrates sending a transaction on Sepolia using this new helper. It works 🎉 

A problem I ran into: `viem` expects typescript >5.1 (for cutting edge typing features) and node >18 (for `fetch`). Does that mean we need to upgrade all of our packages? I tried to mandate this only in the new packages, but there are warnings when running `typecheck` now:
```
examples/with-viem                       |  WARN  Unsupported engine: wanted: {"node":">=18.0.0"} (current: {"node":"v16.16.0","pnpm":"8.4.0"})
packages/viem                            |  WARN  Unsupported engine: wanted: {"node":">=18.0.0"} (current: {"node":"v16.16.0","pnpm":"8.4.0"})
```
I put 2 `.npmrc` files in both `with-viem` and `viem` packages so `pnpm run <foo>` does the right thing and uses node 18, but still uses node 16 everywhere else. Not sure if it's worth supporting 2 versions of node?

### DIFF
--- a/examples/with-viem/src/index.ts
+++ b/examples/with-viem/src/index.ts
@@ -1,0 +1,49 @@
+import * as path from "path";
+import * as dotenv from "dotenv";
+
+import { createApiKeyAccount } from "@turnkey/viem";
+import { createWalletClient, http } from "viem";
+import { sepolia } from "viem/chains";
+
+// Load environment variables from `.env.local`
+dotenv.config({ path: path.resolve(process.cwd(), ".env.local") });
+
+async function main() {
+  const turnkeyAccount = await createApiKeyAccount({
+    apiPublicKey: process.env.API_PUBLIC_KEY!,
+    apiPrivateKey: process.env.API_PRIVATE_KEY!,
+    baseUrl: process.env.BASE_URL!,
+    organizationId: process.env.ORGANIZATION_ID!,
+    privateKeyId: process.env.PRIVATE_KEY_ID!,
+  });
+
+  const client = createWalletClient({
+    account: turnkeyAccount,
+    chain: sepolia,
+    transport: http(
+      `https://sepolia.infura.io/v3/${process.env.INFURA_API_KEY!}`
+    ),
+  });
+
+  // This demo sends ETH back to our faucet (we keep a bunch of Sepolia ETH at this address)
+  const turnkeyFaucet = "0x08d2b0a37F869FF76BACB5Bab3278E26ab7067B7";
+
+  const transactionRequest = {
+    to: turnkeyFaucet as `0x${string}`,
+    value: 1000000000000000n,
+  };
+
+  const txHash = await client.sendTransaction(transactionRequest);
+
+  print("Source address", client.account.address);
+  print("Transaction", `https://sepolia.etherscan.io/tx/${txHash}`);
+}
+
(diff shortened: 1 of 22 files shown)


packages/viem/src/index.ts
```diff
@@ -0,0 +1,291 @@
...
+/**
+ * Creates a new Custom Account backed by a Turnkey API key.
+ */
+export async function createApiKeyAccount(
+  config: TApiKeyAccountConfig
+): Promise<LocalAccount> {
+  const { apiPublicKey, apiPrivateKey, baseUrl, organizationId, privateKeyId } =
+    config;
+
+  httpInit({
```
MaxBlaushild, line 53: If we use this global init, and users create multiple api key accounts, are the newer instances going to clobber the configs of the older instances?

packages/viem/src/index.ts
```diff
@@ -0,0 +1,291 @@
...
+
+async function signTransaction<
+  TTransactionSerializable extends TransactionSerializable
+>(
+  transaction: TTransactionSerializable,
+  serializer: SerializeTransactionFn<TTransactionSerializable>,
+  organizationId: string,
+  privateKeyId: string
+): Promise<`0x${string}`> {
+  const serializedTx = serializer(transaction);
```
MaxBlaushild, line 134: It looks like these serializers may be user supplied. If thats the case, are we sure that these will produce a format that our API understands? Is it worth it to do some validation on the payload itself clientside if thats the case. If my assumption is wrong, please ignore!

MaxBlaushild (approved):
The only thing that really concerns me here is that global config, but otherwise looks good! The config seems like a separate issue, so feel free to ship as is

examples/with-viem/src/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    baseUrl: process.env.BASE_URL!,
+    organizationId: process.env.ORGANIZATION_ID!,
+    privateKeyId: process.env.PRIVATE_KEY_ID!,
+  });
+
+  const client = createWalletClient({
+    account: turnkeyAccount,
+    chain: sepolia,
+    transport: http(
+      `https://sepolia.infura.io/v3/${process.env.INFURA_API_KEY!}`
```
andrewkmin, line 24: should we include the default community API key somewhere? 
`84842078b09946638c03157f83405213`

examples/with-viem/src/index.ts
```diff
@@ -0,0 +1,49 @@
...
+    ),
+  });
+
+  // This demo sends ETH back to our faucet (we keep a bunch of Sepolia ETH at this address)
```
andrewkmin, line 28: 🧠 nice! we can/should do this with all our examples (later)

examples/with-viem/src/index.ts
r-n-o: Oh yeah. I can include it in the example env file 👍

andrewkmin (approved):

PR #98, opened 2023-08-14
Minor fixes to env files and Solana example

jordansexton:
## Summary & Motivation
There are two small fixes in this PR.

1. [Recent examples](https://github.com/tkhq/sdk/blob/bbbe31ff764cc356f6af357de5c343db27467f8a/examples/with-viem/.env.local.example#L3) use `https://api.turnkey.com` as the API base URL. 965c8897200beebf155c55b99c4de3576c6208f3 updates all the example env files to use it.

2. The Solana example contains a bug that causes it to fail after creating a private key. It's caused by the type of call being `ACTIVITY_TYPE_CREATE_PRIVATE_KEYS_V2` and the result that is read from being `createPrivateKeysResult`, presumbly the V1 API. 076c201fcc8f10fc7868087ac9e94b8505e6fbe2 fixes this.

## How I Tested These Changes
I ran the Solana example project, but otherwise haven't tested this.

### DIFF
--- a/examples/deployer/.env.local.example
+++ b/examples/deployer/.env.local.example
@@ -1,6 +1,6 @@
 API_PUBLIC_KEY="<Turnkey API Public Key (that starts with 02 or 03)>"
 API_PRIVATE_KEY="<Turnkey API Private Key>"
-BASE_URL="https://coordinator-beta.turnkey.io"
+BASE_URL="https://api.turnkey.com"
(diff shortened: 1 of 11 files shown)


andrewkmin (approved):
🐐 thanks so much for these contributions! also, awesome work with solana-passkeys — really appreciate the dogfooding

PR #103, opened 2023-08-19
`@turnkey/api-key-stamper`: web bundlers should not try to polyfill `crypto`

r-n-o:
## Summary & Motivation
Webpack (< 5) defaults to polyfilling `crypto` if it sees it imported. After a bit of research on the topic (reading [this discussion](https://github.com/webpack/webpack/issues/8826) in particular), providing a separate browser-specific endpoint is the preferred, most widely supported way to declare an alternate file for browser bundlers. See [package.json's `browser` doc](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#browser).

The "separate entrypoint" approach certainly works but requires a bit of duplication to have two "main" files (`index-browser.ts` and `index.ts` -- see [`1868493`](https://github.com/tkhq/sdk/pull/103/commits/186849364c1f43cd098deaf3be597ab78a2a701d) for what this looks like).

I've pivoted to a simpler approach: include `"crypto": false` in package.json. No code change needed!

## How I Tested These Changes
See this on StackBlitz: https://stackblitz.com/edit/github-wosujp?file=README.md 

This contains a toy Webpack project which loads the codesandbox release from this PR: `https://pkg.csb.dev/tkhq/sdk/commit/0d6633e5/@turnkey/api-key-stamper`. It works! No errors 🎉 

If I load the latest `@turnkey/api-key-stamper` instead, I get the following error:
```
BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "crypto": require.resolve("crypto-browserify") }'
        - install 'crypto-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "crypto": false }
```

To work around this the simplest solution is to add:
```
  resolve: {
    fallback: {
      crypto: false,
    }
  },
```
to `webpack.config.js`. But that won't be needed once this PR gets merged!


james-callahan (approved):
Looks suitable according to this doc I found: https://github.com/defunctzombie/package-browser-field-spec#ignore-a-module assuming that you've tested that webpack follows it.

Would an alternative solution would be to throw away the current node implementation entirely? Doesn't node provide the brower's `crypto.subtle` apis? https://nodejs.org/api/webcrypto.html#cryptosubtle

r-n-o:
> Would an alternative solution would be to throw away the current node implementation entirely? Doesn't node provide the brower's crypto.subtle apis? https://nodejs.org/api/webcrypto.html#cryptosubtle

Interesting! Yeah I guess so! However, the node implementation is pretty compact and simple (~10 lines) so I'm inclined to leave it as-is for now.

PR #104, opened 2023-08-22
Migrate the remaining examples to use new `TurnkeyClient`

andrewkmin:
## Summary & Motivation
Counterpart for https://github.com/tkhq/sdk/pull/105.
- $title, for the following:
  - deployer
  - rebalancer
  - sweeper
  - trading-runner
  - with-cosmjs
  - with-gnosis
  - with-nonce-manager
- Fixes some misc typos
- Fix wrapping/unwrapping behavior in `trading-runner`
- Updates `withcosmjs` to use current Celestia testnet + RPC

## How I Tested These Changes
Ran each locally!

### DIFF
--- a/examples/deployer/src/createNewEthereumPrivateKey.ts
+++ b/examples/deployer/src/createNewEthereumPrivateKey.ts
@@ -41,16 +41,11 @@ export async function createNewEthereumPrivateKey() {
       timestampMs: String(Date.now()), // millisecond timestamp
     });
 
-    const privateKeyId = refineNonNull(
-      activity.result.createPrivateKeysResultV2?.privateKeys?.[0]?.privateKeyId
+    const privateKeys = refineNonNull(
+      activity.result.createPrivateKeysResultV2?.privateKeys
     );
-
-    const keyInfo = await turnkeyClient.getPrivateKey({
-      organizationId: process.env.ORGANIZATION_ID!,
-      privateKeyId,
-    });
-
-    const address = refineNonNull(keyInfo.privateKey.addresses[0]?.address);
+    const privateKeyId = refineNonNull(privateKeys?.[0]?.privateKeyId);
+    const address = refineNonNull(privateKeys?.[0]?.addresses?.[0]?.address);
 
     // Success!
     console.log(
--- a/examples/deployer/src/createNewEthereumPrivateKey.ts
+++ b/examples/deployer/src/createNewEthereumPrivateKey.ts
@@ -1,56 +1,53 @@
-import { TurnkeyApi, init as httpInit, withAsyncPolling } from "@turnkey/http";
+import { TurnkeyClient } from "@turnkey/http";
+import { createActivityPoller } from "@turnkey/http/dist/async";
+import { ApiKeyStamper } from "@turnkey/api-key-stamper";
 import { TurnkeyActivityError } from "@turnkey/ethers";
 import * as crypto from "crypto";
 
 export async function createNewEthereumPrivateKey() {
-  // Initialize `@turnkey/http` with your credentials
-  httpInit({
-    apiPublicKey: process.env.API_PUBLIC_KEY!,
-    apiPrivateKey: process.env.API_PRIVATE_KEY!,
-    baseUrl: process.env.BASE_URL!,
-  });
+  const turnkeyClient = new TurnkeyClient(
+    { baseUrl: process.env.BASE_URL! },
+    new ApiKeyStamper({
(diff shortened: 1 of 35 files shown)


examples/deployer/src/createNewEthereumPrivateKey.ts
```diff
@@ -1,56 +1,53 @@
...
     );
 
-    const keyInfo = await TurnkeyApi.getPrivateKey({
-      body: {
-        organizationId: process.env.ORGANIZATION_ID!,
-        privateKeyId,
-      },
+    const keyInfo = await turnkeyClient.getPrivateKey({
+      organizationId: process.env.ORGANIZATION_ID!,
+      privateKeyId,
```
r-n-o, line 50: nit: technically we don't have to make that call: the v2 result already contains the address!

examples/rebalancer/src/requests/createActivityApproval.ts
```diff
@@ -1,35 +1,34 @@
...
+import { TurnkeyClient } from "@turnkey/http";
+import { createActivityPoller } from "@turnkey/http/dist/async";
+import { ApiKeyStamper } from "@turnkey/api-key-stamper";
 import { TurnkeyActivityError } from "@turnkey/ethers";
 import { refineNonNull } from "./utils";
 
 export default async function approveActivity(
   activityId: string,
   activityFingerprint: string
 ): Promise<string> {
```
r-n-o, line 10: Note: for all these files (`in src/requests/*.ts`), I think we should pass the TurnkeyClient as an argument potentially? That'd avoid having to read process.env in every file.

Not a big deal though, it's an example; not a production-ready app!

examples/rebalancer/src/requests/createPrivateKey.ts
```diff
@@ -1,49 +1,56 @@
...
     );
-    const privateKeyId = refineNonNull(privateKey.privateKeyId);
-    const address = refineNonNull(privateKey.addresses?.[0]?.address);
+
+    const keyInfo = await turnkeyClient.getPrivateKey({
+      organizationId: process.env.ORGANIZATION_ID!,
+      privateKeyId,
+    });
+
+    const address = refineNonNull(keyInfo.privateKey.addresses[0]?.address);
```
r-n-o, line 53: Same here, we can save a call to `getPrivateKey` now that V2 results contain the address!

examples/trading-runner/src/requests/createPrivateKey.ts
```diff
@@ -1,49 +1,56 @@
...
+    const privateKeyId = refineNonNull(
+      activity.result.createPrivateKeysResultV2?.privateKeys?.[0]?.privateKeyId
     );
-    const privateKeyId = refineNonNull(privateKey.privateKeyId);
-    const address = refineNonNull(privateKey.addresses?.[0]?.address);
+
+    const keyInfo = await turnkeyClient.getPrivateKey({
+      organizationId: process.env.ORGANIZATION_ID!,
+      privateKeyId,
+    });
```
r-n-o, line 51: And here as well. We can get the address from `createPrivateKeysResultV2?.privateKeys?.[0]?.addresses?.[0]?.address`

r-n-o (commented):
Minor things below: overall looking great!

I didn't realize how many files there was in our rebalancer example 😱

examples/rebalancer/src/index.ts
```diff
@@ -29,6 +31,15 @@ const TRANSFER_GAS_LIMIT = 21000;
...
 const ACTIVITIES_LIMIT = "100";
 
+// For demonstration purposes, create a globally accessible TurnkeyClient
+const turnkeyClient = new TurnkeyClient(
+  { baseUrl: process.env.BASE_URL! },
+  new ApiKeyStamper({
+    apiPublicKey: process.env.API_PUBLIC_KEY!,
+    apiPrivateKey: process.env.API_PRIVATE_KEY!,
+  })
+);
```
r-n-o, line 41: I think this should be created inside of `main`? It's not very global if you have to pass it into each function anyway ;)

EDIT: oh I see. Actually it _is_ kind of global because it's not passed explicitly to `fundImpl` and `sweepImpl`

r-n-o (approved):
💯 / 💯

examples/rebalancer/src/index.ts
andrewkmin: Haha yeah maybe `global` wasn't the accurate term here but we can roll with it for now for the examples' sake!

PR #108, opened 2023-08-25
federated/viem passkey: Add login button leveraging updated Whoami endpoint

emostov:
## Summary & Motivation

This updates the federated passkey example to include a login button. Despite the example being stateless, we login by using the updated `whoami` endpoint which now allows lookups for sub-orgs using the parent-org ID + WebAuthN stamp.

## How I Tested These Changes

I developed this against staging. Note the `whoami` endpoint updates have not been deployed to production yet so we should wait until that happens.

### DIFF
--- a/examples/with-federated-passkeys/src/pages/index.tsx
+++ b/examples/with-federated-passkeys/src/pages/index.tsx
@@ -45,18 +45,20 @@ export default function Home() {
     register: privateKeyFormRegister,
     handleSubmit: privateKeyFormSubmit,
   } = useForm<privateKeyFormData>();
+  const { register: _loginFormRegister, handleSubmit: loginFormSubmit } =
+    useForm();
+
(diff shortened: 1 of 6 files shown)


examples/with-federated-passkeys/src/pages/index.tsx
```diff
@@ -150,6 +167,17 @@ export default function Home() {
               value="Create new sub-organization"
             />
           </form>
+          <div></div>
+          <h2 className={styles.prompt}>
+            OR already created a sub-org? Login!
+          </h2>
+          <form className={styles.form} onSubmit={loginFormSubmit(login)}>
```
emostov, line 174: Maybe an abuse of the form element, but this was the most obvious way for me (a helpless frontend dev) to create a button

MaxBlaushild (approved):

andrewkmin (approved):

PR #110, opened 2023-09-01
Sync protos; Bump http package

emostov:

### DIFF
--- a/packages/http/CHANGELOG.md
+++ b/packages/http/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @turnkey/http
 
+## 1.1.2
+
+### Patch Changes
+
+- Sync protos
+
 ## 1.1.1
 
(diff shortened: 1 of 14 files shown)


emostov:
cc @andrewkmin

packages/http/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @turnkey/http
 
+## 1.1.2
+
+### Patch Changes
+
+- Sync protos
```
r-n-o, line 7: I think this should be a minor release because we're introducing a new capability: create sub-org with keys. Also think we need a more detailed changelog entry to let people know that they'll need to update their code to look for createSubOrganization results in `createSubOrganizationV3` going forward.

emostov:
closing in favor of #111

PR #111, opened 2023-09-01
Bump http package version

emostov:

### DIFF
--- a/packages/http/CHANGELOG.md
+++ b/packages/http/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @turnkey/http
 
+## 1.2.0
+
+### Minor Changes
+
+- The return type for `createSubOrganization` has been updated to include a `privateKeys` field to reflect that it now calls createSubOrganizationV3
+
 ## 1.1.1
 
 ### Patch Changes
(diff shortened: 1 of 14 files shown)


packages/http/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @turnkey/http
 
+## 1.2.0
+
+### Minor Changes
+
+- The return type for `createSubOrganization` has been updated to include a `privateKeys` field to reflect that it now calls createSubOrganizationV3
```
emostov, line 7: @r-n-o is this correct?

r-n-o: Kind of! I think we should be more concrete/explicit about what's changing:
* calling `.createSubOrganization` on our HTTP client will trigger an activity of type `CREATE_SUB_ORGANIZATION_V3` instead of `CREATE_SUB_ORGANIZATION_V2` previously
* If there are any policies referencing `CREATE_SUB_ORGANIZATION_V2` specifically, they won't work out of the box. These policies will need to be updated to allow `CREATE_SUB_ORGANIZATION_V3`
* `CREATE_SUB_ORGANIZATION_V3` supports everything `CREATE_SUB_ORGANIZATION_V2` supports, with the addition of a `privateKeys` field to atomically create a sub-org and private keys
* When reading `createSubOrganization` results, SDK users need to look at `activity.result.createSubOrganizationResultV3` instead of `activity.result.createSubOrganizationResult` previously

andrewkmin: will update the readme in my PR

andrewkmin (approved):

PR #118, opened 2023-09-15
Update federated passkey demo

octavien-mural:
## Summary & Motivation
Added an additional algorithm that is now supported, and added new authenticatorSelection parameters to support android based passkeys that do not support discoverable credentials 

## How I Tested These Changes
Discussed with Max and tested locally that I am able to create and use a passkey with an android phone

### DIFF
--- a/examples/with-federated-passkeys/src/pages/index.tsx
+++ b/examples/with-federated-passkeys/src/pages/index.tsx
@@ -113,7 +113,7 @@ export default function Home() {
     // https://www.w3.org/TR/webauthn-2/#sctn-sample-registration
     const attestation = await getWebAuthnAttestation({
       publicKey: {
-         authenticatorSelection: {
+        authenticatorSelection: {
           residentKey: "preferred",
           requireResidentKey: false,
           userVerification: "preferred",
--- a/examples/with-federated-passkeys/src/pages/index.tsx
+++ b/examples/with-federated-passkeys/src/pages/index.tsx
@@ -21,8 +21,9 @@ type privateKeyResult = {
 };
 
 // All algorithms can be found here: https://www.iana.org/assignments/cose/cose.xhtml#algorithms
-// We only support ES256, which is listed here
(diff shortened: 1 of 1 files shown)


examples/with-federated-passkeys/src/pages/index.tsx
```diff
@@ -122,6 +128,10 @@ export default function Home() {
             type: publicKey,
             alg: es256,
           },
+          {
+            type: publicKey,
+            alg: rs256,
+          }
```
andrewkmin, line 134: I think prettier wants a trailing comma here. `pnpm -w prettier-all:write`

andrewkmin (changes requested):
🎉🎉 thanks for updating this example! looks great; see comment about `prettier`

examples/with-federated-passkeys/src/pages/index.tsx
octavien-mural: Just pushed again!

andrewkmin (approved):

examples/with-federated-passkeys/src/pages/index.tsx
```diff
@@ -112,6 +113,11 @@ export default function Home() {
     // https://www.w3.org/TR/webauthn-2/#sctn-sample-registration
     const attestation = await getWebAuthnAttestation({
       publicKey: {
+         authenticatorSelection: {
+          residentKey: "preferred",
+          requireResidentKey: false,
+          userVerification: "preferred"
```
andrewkmin, line 119: hey @octavien-mural sorry for our linter being so pushy, but we may need another trailing comma here :) 

to be safe, you can also try running `pnpm -w prettier-all:write`, in case anything else is missed (doesn't look like it though)

octavien-mural: Apologies, just pushed again! I believe I have different prettier settings so I'm unable to run the command with the correct styling

PR #119, opened 2023-09-17
set up viem tests

andrewkmin:


packages/viem/src/__tests__/index-test.ts
```diff
@@ -1,15 +1,48 @@
...
+import Test721 from "./contracts/artifacts/src/__tests__/contracts/source/Test721.sol/Test721.json";
+import { expect, beforeEach, describe, test } from "@jest/globals";
 
-import { createApiKeyAccount } from "../";
+// @ts-expect-error
+const testCase: typeof test = (...argList) => {
+  if (!process.env.BANNED_TO_ADDRESS) {
+    // For now, this test requires certain environment variables to be injected (from Turnkey's internal environment)
+    return test.skip(...argList);
+  }
```
r-n-o, line 28: This means the test won't be run in CI right? I don't think we inject this env variable anywhere right now.

packages/viem/src/__tests__/contracts/artifacts/@openzeppelin/contracts/access/Ownable.sol/Ownable.dbg.json
r-n-o, line 1: Let's ignore all files in `artifacts`?

packages/viem/src/__tests__/index-test.ts
andrewkmin: It's injected here: https://github.com/tkhq/mono/blob/2fac89233c77dfd86d64094537703b31bd4271eb/src/go/e2e/e2e_viem_test.go#L103

r-n-o: right right. But I think we should run this in _this repo_'s CI as well no? That would be useful to detect bad changes to the viem package itself without running tests through our private monorepo.

andrewkmin: ah i see what you're saying. i'll mess around with CI!

r-n-o: If it's too hard don't worry about it though: can be its own issue/PR. Not running tests in the status quo (unfortunately); ethers tests are in the same boat!

https://github.com/tkhq/sdk/blob/397b6b5fb1a17d5a88c372e97c1485c9abacb419/packages/ethers/src/__tests__/index-test.ts#L15

packages/viem/src/__tests__/contracts/artifacts/src/__tests__/contracts/source/Test721.sol/Test721.json
r-n-o, line 1: Looks like there's one more file to remove? (should be gitignored as well)

packages/viem/src/__tests__/index-test.ts
```diff
@@ -34,35 +67,265 @@ describe("createApiKeyAccount", () => {
...
+      value: parseEther("1.0"),
+      chain,
+      nonce: transactionCount,
+      gas: 21000n,
+      maxFeePerGas: parseUnits("2", 9),
+      maxPriorityFeePerGas: parseUnits("2", 9),
+      type: "eip1559",
+    });
+
+    expect(txHash).toMatch(/^0x/);
```
r-n-o, line 140: something that could be stronger than "txHash starts with 0x" could be: 1) fetch the tx details from hash, then 2) assert against the tx details ("to" should be the address we want, amount should be 1 ETH, etc)

Definitely nice-to-have though!

packages/viem/src/__tests__/index-test.ts
```diff
@@ -34,35 +67,265 @@ describe("createApiKeyAccount", () => {
...
+          },
+        ],
+      };
+
+      // Looks like here, too, we are subject to similar issues as described below:
+      // See https://github.com/ethers-io/ethers.js/issues/1683
```
r-n-o, line 273: is this still relevant given we're in viem tests, rather than ethers?

packages/viem/src/__tests__/contracts/artifacts/src/__tests__/contracts/source/Test721.sol/Test721.json
andrewkmin: this one is actually required for one of the unit tests. i moved it to a more surface-level location to avoid confusion :)

packages/viem/src/__tests__/index-test.ts
andrewkmin: i'll do these as a followup for both viem and ethers

packages/viem/src/__tests__/index-test.ts
andrewkmin: i think so, in a certain way given there are some weird behaviors with anvil <> viem. viem expects a signer if the `from` isn't provided; in a real setting, this would be provided via Metamask/a client-side wallet.

if anything, i can comment this out until we're ready to have an example for react-native-viem

.github/workflows/js-build.yml
```diff
@@ -26,8 +26,21 @@ jobs:
       - name: Typecheck
         run: pnpm run typecheck-all
 
+      # Run Anvil node in background for e2e tests
+      # - name: Anvil
+      #   run: |
+      #     anvil &
```
r-n-o, line 32: Remove if not used?

.github/workflows/js-build.yml
```diff
@@ -26,8 +26,21 @@ jobs:
...
-        run: pnpm run test-all
+        run: anvil & pnpm run test-all
+        env: # Or as an environment variable
+          API_PUBLIC_KEY: ${{ secrets.API_PUBLIC_KEY }}
+          API_PRIVATE_KEY: ${{ secrets.API_PRIVATE_KEY }}
+          BASE_URL: ${{ secrets.BASE_URL }}
+          ORGANIZATION_ID: ${{ secrets.ORGANIZATION_ID }}
+          PRIVATE_KEY_ID: ${{ secrets.PRIVATE_KEY_ID }}
+          EXPECTED_ETH_ADDRESS: ${{ secrets.EXPECTED_ETH_ADDRESS }}
+          BANNED_TO_ADDRESS: ${{ secrets.BANNED_TO_ADDRESS }}
```
r-n-o, line 43: I think some of these could be public?

r-n-o (approved):
Minor comments below; this PR looks good to go otherwise 🚀

PR #125, opened 2023-10-04
add buffer polyfill for webauthn-stamper

andrewkmin:

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -619,7 +619,11 @@ importers:
         specifier: ^1.10.0
         version: 1.10.0(typescript@5.1.3)
 
-  packages/webauthn-stamper: {}
+  packages/webauthn-stamper:
+    dependencies:
+      buffer:
+        specifier: ^6.0.3
+        version: 6.0.3
 
 packages:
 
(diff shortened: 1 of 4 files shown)


r-n-o:
CI: `TurnkeyActivityError: Failed to sign: Turnkey error 7: You have reached the alloted amount of signatures for your subscription. Please upgrade to continue.`. Uh oh!

pnpm-lock.yaml
```diff
@@ -4480,7 +4484,7 @@ packages:
   /@types/bn.js@5.1.2:
     resolution: {integrity: sha512-dkpZu0szUtn9UXTmw+e0AJFd4D2XAxDnsCLdc05SfqpqzPEBft8eQr8uaFitfo/dUUOZERaLec2hHMG87A4Dxg==}
     dependencies:
-      '@types/node': 20.6.2
+      '@types/node': 16.18.12
```
r-n-o, line 4487: why is that happening? Is this on purpose?

andrewkmin: was just trying to figure this out - after doing some testing, it does not look like local node version (toggled using `nvm`) is the cause. i think it could be the installation of buffer that introduces this deep dependency change

r-n-o (approved):

PR #129, opened 2023-10-10
New `@turnkey/iframe-stamper` package and email recovery example

r-n-o:
## Summary & Motivation
This PR adds a new package: `@turnkey/iframe-stamper`. This is an abstraction to work with iframes **as signers** (aka stampers).

Right now I envision `@turnkey/iframe-stamper` to be used mostly via our upcoming email recovery flow, but I can see this "iframe as stampers" idea being useful for other flows in the future: iframes and localStorage could be used to hold standard API keys rather than recovery credentials.

The new example covers the recovery flow.

## How I Tested These Changes

https://github.com/tkhq/sdk/assets/104520680/296f7879-610f-4d81-a173-7de94e46efa7

Local testing for now. There's still a bunch of work that needs to be done to complete the example, but having this PR out in the open will be useful to get feedback!


packages/iframe-stamper/src/index.ts
```diff
@@ -0,0 +1,192 @@
...
+  iframeOrigin: string;
+  iframePublicKey: string | null;
+
+  /**
+   * Creates a new iframe stamper. This function _does not_ insert the iframe in the DOM.
+   * Call `.init()` to insert the iframe element in the DOM.
+   */
+  constructor(config: TIframeStamperConfig) {
+    if (typeof window === "undefined") {
+      throw new Error("cannot initialize iframe in non-browser environment");
```
MaxBlaushild, line 33: Nit: caps

packages/iframe-stamper/src/index.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  /**
+   * Inserts the iframe on the page and returns a promise resolving to the iframe's public key
+   */
+  async init(): Promise<string> {
+    this.container.appendChild(this.iframe);
+    const iframeOrigin = this.iframeOrigin;
+
+    // TODO: undo this hack.
+    const that = this;
```
MaxBlaushild, line 70: I think you could use arrow functions to get rid of this

packages/iframe-stamper/src/index.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    return new Promise(function (resolve, _reject) {
+      window.addEventListener(
+        "message",
+        (event) => {
+          console.log("window received event", event);
```
MaxBlaushild, line 76: Nit: remove

packages/iframe-stamper/src/index.ts
```diff
@@ -0,0 +1,192 @@
...
+          if (event.origin !== iframeOrigin) {
+            // There might be other things going on in the window, for example: react dev tools, other extensions, etc.
+            // Instead of erroring out
+            return;
+          }
+          if (event.data && event.data["type"] == "PUBLIC_KEY_READY") {
+            console.log(`iframe ready. Public key ${event.data["value"]}`);
```
MaxBlaushild, line 83: Same

packages/iframe-stamper/src/index.ts
r-n-o: Yep! Good call.

MaxBlaushild (approved):
Looks great!

examples/email-recovery/src/pages/index.tsx
```diff
@@ -0,0 +1,239 @@
...
+  const initRecovery = async (data: InitRecoveryFormData) => {
+    const response = await axios.post("/api/initRecovery", {
+      email: data.email,
+      targetPublicKey: data.targetPublicKey,
+    });
+    setInitRecoveryResponse(response.data);
+  };
+
+  const recoverUser = async (data: RecoverUserFormData) => {
+    if (iframeStamper === null) {
```
andrewkmin, line 79: if we want to cover the `undefined` case, we can do a `falsy` check here (and in the line to follow for `initRecoveryResponse`)

examples/email-recovery/src/pages/index.tsx
```diff
@@ -0,0 +1,239 @@
...
+            challenge: base64UrlEncode(challenge),
+            attestation: attestation,
+          },
+        ],
+      },
+    });
+
+    // TODO: instead of simply alerting, POST the request to Turnkey to complete recovery.
+    console.log("SUCCESS", signedRequest);
+    alert("SUCCESS! Request SIGNED!");
```
andrewkmin, line 149: should we direct attention in the alert to the console?

examples/email-recovery/src/pages/index.tsx
```diff
@@ -0,0 +1,239 @@
...
+            <label className={styles.label}>
+              Email
+              <input
+                className={styles.input}
+                {...initRecoveryFormRegister("email")}
+                placeholder="Email"
+              />
+            </label>
+            <label className={styles.label}>
+              Target Public Key (FYI)
```
andrewkmin, line 193: naive q: what does `FYI` refer to here?

packages/iframe-stamper/README.md
```diff
@@ -0,0 +1,33 @@
...
+// This inserts the iframe in the DOM and returns the public key
+const publicKey = await iframeStamper.init();
+
+// Injects a new credential in the iframe
+const injected = await iframeStamper.injectRecoveryBundle(recoveryBundle);
+
+// New HTTP client able to sign with the credentials inside of the iframe
+const httpClient = new TurnkeyClient(
+  { baseUrl: "https://api.turnkey.com" },
+  stamper
```
andrewkmin, line 31: nit: `stamper: iframeStamper`?

packages/iframe-stamper/src/index.ts
```diff
@@ -0,0 +1,182 @@
...
+    this.iframeOrigin = iframeUrl.origin;
+
+    // This is populated once the iframe is ready. Call `.init()` to kick off DOM insertion!
+    this.iframePublicKey = null;
+  }
+
+  /**
+   * Inserts the iframe on the page and returns a promise resolving to the iframe's public key
+   */
+  async init(): Promise<string> {
```
andrewkmin, line 65: 🏅 nice!!!

packages/iframe-stamper/src/index.ts
```diff
@@ -0,0 +1,182 @@
...
+    return new Promise((resolve, _reject) => {
+      window.addEventListener(
+        "message",
+        (event) => {
+          if (event.origin !== this.iframeOrigin) {
+            // There might be other things going on in the window, for example: react dev tools, other extensions, etc.
+            // Instead of erroring out
+            return;
+          }
+          if (event.data && event.data["type"] == "PUBLIC_KEY_READY") {
```
andrewkmin, line 76: tiny nit: can use `?.` [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here

packages/iframe-stamper/src/index.ts
```diff
@@ -0,0 +1,182 @@
...
+    return new Promise((resolve, _reject) => {
+      window.addEventListener(
+        "message",
+        (event) => {
+          if (event.origin !== this.iframeOrigin) {
+            // There might be other things going on in the window, for example: react dev tools, other extensions, etc.
+            // Instead of erroring out we simply return. Not our event!
+            return;
+          }
+          if (event.data && event.data["type"] == "BUNDLE_INJECTED") {
```
andrewkmin, line 124: tiny nit - same optional chaining pattern here (don't feel strongly).

could be useful to throw some of these types into consts in a single place

andrewkmin (commented):
very minor nit: let's show this off in our top level README 🤩 

but otherwise. this. looks. great!!!

examples/email-recovery/src/pages/index.tsx
r-n-o: Oh that's just "For You Information". Because it's not really needed in the UI. But I wanted to show it in the demo because it's interesting to see I think?

examples/email-recovery/src/pages/index.tsx
r-n-o: I've reworked this part to actually POST. Resolved :)

examples/email-recovery/src/pages/index.tsx
r-n-o: Sounds good to me!

packages/iframe-stamper/README.md
r-n-o: Oh right! Good catch.

examples/email-recovery/email_recovery_steps.png
emostov, line 1: "encryption key pair" -> "encryption target key" or "encryption target key pair"

examples/email-recovery/next.config.js
emostov, line 1: Is this def necessary?

examples/email-recovery/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "eslint-config-next": "13.4.7",
+    "esm": "^3.2.25",
+    "install": "^0.13.0",
+    "next": "13.4.7",
+    "npm": "^9.7.2",
+    "react": "18.2.0",
+    "react-dom": "18.2.0",
+    "react-hook-form": "^7.45.1",
+    "typescript": "5.1.3"
+  }
```
emostov, line 31: Do we know that we need all these deps?

r-n-o: All these dependencies are for the example app. I don't think we need to be particularly strict here. Clients building email recovery will use our `iframe-stamper` package in a variety of contexts so I expect dependencies to be significantly different.

That being said: for the ones you highlighted the answer is yes: we need react-dom (base lib for react), react-hook-form (to power forms with `useForm`) and typescript because...we're writing typescript!

examples/email-recovery/email_recovery_steps.png
r-n-o: Good call, I'll rename 👍

examples/email-recovery/src/components/Recovery.tsx
```diff
@@ -0,0 +1,42 @@
...
+  );
+
+  useEffect(() => {
+    if (!iframeStamper) {
+      const iframeStamper = new IframeStamper({
+        iframeUrl: props.iframeUrl,
+        iframeContainerId: TurnkeyIframeContainerId,
+        iframeElementId: TurnkeyIframeElementId,
+      });
+      iframeStamper.init().then(function () {
```
emostov, line 27: any reason not to use an fat arrow function here for consistency?

examples/email-recovery/src/pages/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+const rs256 = -257;
+
+// This constant designates the type of credential we want to create.
+// The enum only supports one value, "public-key"
+// https://www.w3.org/TR/webauthn-2/#enumdef-publickeycredentialtype
+const publicKey = "public-key";
+
+const generateRandomBuffer = (): ArrayBuffer => {
+  const arr = new Uint8Array(32);
+  crypto.getRandomValues(arr);
```
emostov, line 43: where is this `crypto` import coming from?

examples/email-recovery/src/pages/index.tsx
```diff
@@ -0,0 +1,243 @@
...
+    });
+
+    const client = new TurnkeyClient(
+      {
+        baseUrl: process.env.NEXT_PUBLIC_BASE_URL!,
+      },
+      iframeStamper
+    );
+
+    // TODO: switch this to be `client.recoverUser`!
```
emostov, line 129: should be able to switch this now for staging :)

examples/email-recovery/src/components/Recovery.tsx
r-n-o: no reason in particular. We could use an arrow function 👍

packages/iframe-stamper/src/index.ts
```diff
@@ -0,0 +1,201 @@
...
+
+    const container = document.getElementById(config.iframeContainerId);
+    if (!container) {
+      throw new Error(
+        `Cannot create iframe stamper: no container with ID ${config.iframeContainerId} exists in the current document`
+      );
+    }
+    this.container = container;
+
+    var iframe = window.document.createElement("iframe");
```
emostov, line 69:
```suggestion
    let iframe = window.document.createElement("iframe");
```

packages/iframe-stamper/src/index.ts
```diff
@@ -0,0 +1,201 @@
...
+
+    // This is populated once the iframe is ready. Call `.init()` to kick off DOM insertion!
+    this.iframePublicKey = null;
+  }
+
+  /**
+   * Inserts the iframe on the page and returns a promise resolving to the iframe's public key
+   */
+  async init(): Promise<string> {
+    this.container.appendChild(this.iframe);
```
emostov, line 85: should we check that `this.container` is non null?

packages/iframe-stamper/src/index.ts
```diff
@@ -0,0 +1,201 @@
...
+    return new Promise((resolve, _reject) => {
+      window.addEventListener(
+        "message",
+        (event) => {
+          if (event.origin !== this.iframeOrigin) {
+            // There might be other things going on in the window, for example: react dev tools, other extensions, etc.
+            // Instead of erroring out
+            return;
+          }
+          if (event.data?.type == IframeEventType.PublicKeyReady) {
```
emostov, line 95: do we need `===` since this js?

examples/email-recovery/src/pages/index.tsx
r-n-o: This is the web crypto namespace, automatically available on the window object in browser environments (same as `document` or `navigator`)

examples/email-recovery/src/pages/index.tsx
r-n-o: Not quite since we don't have the updated SDK released yet. I'll see what I can do to pull an early version in. Otherwise we should be able to do this sometime this week when the proper release goes out.

packages/iframe-stamper/src/index.ts
r-n-o: `===` vs `==` makes a big difference when testing against null or undefined or false / truthy value. But if we're testing `3 == 4` or `"foo" == "bar"` I don't think it makes a big difference.

That being said. Doesn't hurt to be conservative!

packages/iframe-stamper/src/index.ts
```diff
@@ -0,0 +1,201 @@
...
+            resolve(event.data["value"]);
+            this.iframePublicKey = event.data["value"];
```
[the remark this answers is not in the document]
r-n-o: This will be reached, but doesn't hurt to switch the ordering for clarity! I'll swap line 96 and 97.

packages/iframe-stamper/src/index.ts
r-n-o: yep yep! Good catch.

r-n-o: Oh actually nevermind. This is already enforced by the non-null type. We already check for this in the constructor:
```ts
    const container = document.getElementById(config.iframeContainerId);
    if (!container) {
      throw new Error(
        `Cannot create iframe stamper: no container with ID ${config.iframeContainerId} exists in the current document`
      );
    }
    this.container = container;
```

examples/email-recovery/src/pages/index.tsx
r-n-o: Done!

Issue #123, opened 2023-09-27
@turnkey/http `getWebAuthnAttestation` / `getWebAuthnAssertion` should perform feature detection


r-n-o:
Not all browsers support passkeys and webauthn. It'd be good to do this check and return an error when passkey or webauthn isn't supported. See https://web.dev/passkey-form-autofill/#feature-detection

PR #130, opened 2023-10-11
add conditional check

andrewkmin:
## Summary & Motivation
https://github.com/tkhq/sdk/issues/123

## How I Tested These Changes
Examples

### DIFF
--- a/packages/http/src/webauthn.ts
+++ b/packages/http/src/webauthn.ts
@@ -114,10 +114,15 @@ export async function getWebAuthnAssertion(
   payload: string,
   options?: TurnkeyCredentialRequestOptions
 ): Promise<string> {
+  const webAuthnSupported = await hasWebAuthnSupport();
+
+  if (!webAuthnSupported) {
+    throw new Error("webauthn is not supported by this browser");
+  }
+
   const signingOptions = await getCredentialRequestOptions(payload, options);
   const clientGetResult = await webauthnCredentialGet(signingOptions);
   const assertion = clientGetResult.toJSON();
-
   const stamp: TWebAuthnStamp = {
     authenticatorData: assertion.response.authenticatorData,
     clientDataJson: assertion.response.clientDataJSON,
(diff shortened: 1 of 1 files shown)


packages/http/src/webauthn.ts
```diff
@@ -131,7 +136,27 @@ export async function getWebAuthnAssertion(
...
+// For additional details see https://web.dev/articles/passkey-form-autofill#feature-detection, https://github.com/w3c/webauthn/wiki/Explainer:-WebAuthn-Conditional-UI
+async function hasWebAuthnSupport(): Promise<boolean> {
+  if (
+    window.PublicKeyCredential &&
+    PublicKeyCredential.isConditionalMediationAvailable
+  ) {
+    const isCMA = await PublicKeyCredential.isConditionalMediationAvailable();
+
+    return isCMA;
+  }
```
r-n-o, line 159: I think this is too restrictive? `isConditionalMediationAvailable` is specific to the autofill feature, which isn't needed for barebone webauthn/passkey support. In our case it's probably enough to return `window.PublicKeyCredential !== undefined`?

packages/http/src/webauthn.ts
```diff
@@ -131,7 +136,20 @@ export async function getWebAuthnAssertion(
...
+
   const res = await webauthnCredentialCreate(options);
 
   return toInternalAttestation(res.toJSON());
 }
+
+// `hasWebAuthnSupport` checks for barebones webauthn support.
+// For additional details and granular settings, see:
+// https://web.dev/articles/passkey-form-autofill#feature-detection, https://developer.mozilla.org/en-US/docs/Web/API/PublicKeyCredential
+function hasWebAuthnSupport(): boolean {
```
andrewkmin, line 153: a little more verbose, but choosing to have this separate function for now in case we want to modify this shared logic

packages/http/src/webauthn.ts
```diff
@@ -114,10 +114,15 @@ export async function getWebAuthnAssertion(
   payload: string,
   options?: TurnkeyCredentialRequestOptions
 ): Promise<string> {
+  const webAuthnSupported = hasWebAuthnSupport();
+
+  if (!webAuthnSupported) {
+    throw new Error("webauthn is not supported by this browser");
+  }
```
r-n-o, line 121: Nit: you can shorten this to:

```ts
  if (!hasWebAuthnSupport()) {
    throw new Error("webauthn is not supported by this browser");
  }
```

r-n-o (approved):

PR #131, opened 2023-10-11
ethers v5 --> v6

andrewkmin:
## Summary & Motivation
$title! The upgrade isn't quite complete until we have an Eip1193Bridge for Ethers v6. This is on the Ethers teams' radar; for now, if that's a crucial piece of functionality for our users, they can resort to using our Ethers v5 implementation.

## How I Tested These Changes
- Unit
- Examples
  - [x] with-gnosis
  - [ ] hi


packages/ethers/src/index.ts
```diff
@@ -1,10 +1,10 @@
 import { ethers } from "ethers";
 import { TurnkeyActivityError, TurnkeyRequestError } from "@turnkey/http";
 import type { TurnkeyClient } from "@turnkey/http";
-import type { TypedDataSigner } from "@ethersproject/abstract-signer";
+// import type { TypedDataSigner } from "@ethersproject/abstract-signer";
```
r-n-o, line 4: remove?

packages/ethers/src/index.ts
```diff
@@ -130,18 +135,17 @@ export class TurnkeySigner extends ethers.Signer implements TypedDataSigner {
...
-      if (ethers.utils.getAddress(unsignedTx.from) !== selfAddress) {
+      if (ethers.getAddress(unsignedTx.from.toString()) !== selfAddress) {
         throw new Error(
           `Transaction \`tx.from\` address mismatch. Self address: ${selfAddress}; \`tx.from\` address: ${unsignedTx.from}`
         );
       }
 
-      delete unsignedTx.from;
+      // maybe we no longer need to drop this
+      // delete unsignedTx.from;
```
r-n-o, line 145: Can you explain your reasoning? Generally, when it doubt: ✂️ -- we can always add this back later if needed.

r-n-o (commented):
v6 reads much better than v5 👏 

Do you want to have a changeset here? I think you'll need to release a major version bump for `@turnkey/ethers`, and then see which packages depend on it. Might be good to spot check the examples using ethers but looks like you've done this already!

packages/ethers/src/index.ts
andrewkmin: I believe ethers v6 now has native support for eip1193 (no `Eip1193Bridge` required). However, I'm not sure this is currently unit testable in this way (node environment). I've reached out to the Ethers team to see if there is a new, Ethers v6-specific way of creating an eip1193 provider.

If they don't come up with a solution in a timely manner or say it's currently not supported, I'm also down to remove this test for now and revisit this topic altogether once we upgrade demo-consumer-wallet as a whole

andrewkmin:
closing in favor of https://github.com/tkhq/sdk/pull/200

PR #142, opened 2023-10-19
sdk release v2023.10.2

andrewkmin:
## Summary & Motivation
Another day another release!

Main updates:

@turnkey/http
- Consolidated signing routes:
  - We now have a single route, `/api/v1/submit/sign_raw_payload` for `SignRawPayload` activities (previously this was split into `SignRawPayload` and `SignRawPayloadV2`)
  - This extends to `/api/v1/submit/sign_transaction` as well. `SignTransaction` and `SignTransactionV2` now have a unified activity: `SignTransaction`
  - Under the hood, these routes stay largely the same: you can sign raw payloads and transactions with an address (from a wallet account), or a private key ID

@turnkey/viem
- 59dcd2f: Unpin typescript
- da7c960: Bump Viem dependency to fix `getAddresses()` for LocalAccount

Also, deleted some examples as they've been moved to https://github.com/tkhq/demo-viem-passkeys and https://github.com/tkhq/demo-ethers-passkeys, respectively


packages/http/CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # @turnkey/http
 
+## 1.4.0
+
+### Minor Changes
```
r-n-o, line 5: Because we've changed the shape of the params for signing (`privateKeyId` => `signWith`) it's technically a breaking change.

Let's mention this in its own line item or section?
```md
### Upgrade notes
* `signRawPayload` and `signTransaction` now expect a `signWith` param instead of `privateKeyId` previously
* `signRawPayload` and `signTransaction` have been updated to expect a new type: `ACTIVITY_TYPE_SIGN_RAW_PAYLOAD_V2` and `ACTIVITY_TYPE_SIGN_TRANSACTION_V2`, respectively
* if you have policies authorizing `ACTIVITY_TYPE_SIGN_RAW_PAYLOAD` or `ACTIVITY_TYPE_SIGN_TRANSACTION` specifically, they will need to be updated to authorize `ACTIVITY_TYPE_SIGN_RAW_PAYLOAD_V2` and `ACTIVITY_TYPE_SIGN_TRANSACTION_V2` (or better yet, update your policies to allow all signing actions using policy resources and actions, see https://docs.turnkey.com/managing-policies/examples
```

After writing this up I think a breaking change is warranted because of the policy implications.

andrewkmin: 👍 makes sense. Think it warrants a major change? https://github.com/tkhq/sdk/pull/142/commits/b3d31ef5a12d219703324fac88fa2e9105fbe031 I can go either way on this

r-n-o: Yeah I think that's safer. I don't think I'd want a major change if it was another activity type; but given this is signing let's make sure people have a chance to look at the notes when they upgrade. Thanks for putting this together!

r-n-o (approved):

PR #145, opened 2023-10-24
New export example

oliviathet:
## Summary & Motivation
This PR demonstrates export functionality for `Wallet` resources.
* `wallet-export`: get a list of your organization's wallets and export a wallet as a mnemonic.

This also adds `iframeStyle` as a configuration option for the iframe-stamper since we display the iframe if the export bundle is successfully injected.

### Important Note on Security:
Turnkey's key export service will allow end-users to export their raw private key or wallet mnemonic by leveraging a secure channel directly between a Turnkey secure enclave and the user's device. This is achieved by the parent page embedding the Turnkey export page as an iframe using the `@turnkey/iframe-stamper` package introduced in https://github.com/tkhq/sdk/pull/129. In the iframe, the end-user's keypair is generated and only the public key is ever shared with the parent page or Turnkey. A Turnkey secure enclave receives the public key, generates its own keypair, computes the shared secret, and encrypts the private key material to the shared secret. The resulting encrypted "export bundle" and public key from the Turnkey enclave are returned to the parent page. The parent page injects the "export bundle" in the Turnkey iframe stamper, which decrypts the bundle in the iframe. More information will be available soon in the official Turnkey docs.


examples/key-export/src/pages/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+    null
+  );
+
+  useEffect(() => {
+    const instantiateIframeStamper = async () => {
+      const stamper = new IframeStamper({
+        iframeUrl: process.env.NEXT_PUBLIC_EXPORT_IFRAME_URL!,
+        iframeContainerId: TurnkeyIframeContainerId,
+        iframeElementId: TurnkeyIframeElementId,
+        iframeStyle: "border: none; width: 600px; height: 200px;",
```
r-n-o, line 30: I'm wondering if this styling can be achieved by targeting the iframe element ID (instead of having a new `iframeStyle` property):

```css
#turnkey-iframe-element-id {
  border: none;
  width: 600px;
  height: 200px;
}
```

examples/key-export/src/pages/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+        return "SECP256K1";
+      }
+      case "CURVE_ED25519": {
+        return "ED25519";
+      }
+      default: {
+        assertNever(curve);
+      }
+    }
+  };
```
r-n-o, line 81: This could be pulled into its own file or function...but heh. Doesn't really hurt or help one way or the other since it's an example.

examples/key-export/src/pages/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+    <svg viewBox="0 0 24 24" fill="none" width="12" height="12">
+      <path
+        d="M12 2L12 16M12 16L16 12M12 16L8 12M2 20H22"
+        stroke="#3f464b"
+        strokeWidth="2"
+        strokeLinecap="round"
+        strokeLinejoin="round"
+      />
+    </svg>
+  );
```
r-n-o, line 67: Should this be its own component?

examples/key-export/src/pages/index.tsx
```diff
@@ -0,0 +1,213 @@
...
+      )}
+      <div
+        style={{ display: showPrivateKey ? "block" : "none" }}
+        id={TurnkeyIframeContainerId}
+      ></div>
+    </main>
+  );
+}
+function assertNever(curve: string) {
+  throw new Error("Function not implemented.");
```
r-n-o, line 212: given this is used in the case where we have an unknown curve, the message should be "Unknown curve: ${curve}" or if you want to keep it generic "case not implemented"

examples/wallet-export/README.md
```diff
@@ -0,0 +1,52 @@
+# Example: `key-export`
+
+This example shows a private key export flow. It contains a NextJS app with:
+
```
r-n-o, line 4: Do you think there's a way to consolidate both wallet and key export in a single example? I worry about having to maintain 2 demos doing roughly the same thing

(the title should say `wallet-export` and "shows a wallet export flow" if we wanted to stick with separate demos!)

What I have in mind is 2 tables instead of one, or 2 pages even. Most of the plumbing should be the same so it'll be nice to have a consolidated example.

r-n-o (commented):
Nice work overall! My main concern is 1 vs 2 examples for export: I think we can get away with a single one? I don't have a super strong opinion about that though, so if you want to stick with 2 I'm definitely fine with it!

examples/key-export/src/pages/index.tsx
oliviathet: i tried this a few different ways before having a new `iframeStyle` property and i might be missing something, but it didn't work :(

oliviathet: wait a second... somehow i got something ELSE to work lol.... will push it in a second! but alas `iframeStyle` _can_ be removed after all

oliviathet:
```
.walletIframe iframe {
  border: none;
  width: 600px;
  height: 600px;
}
```
and
```
<div
  style={{ display: showWallet ? "block" : "none" }}
  id={TurnkeyIframeContainerId}
  className={styles.walletIframe}
/>
```

packages/iframe-stamper/src/index.ts
```diff
@@ -149,6 +155,72 @@ export class IframeStamper {
     });
   }
 
+  /**
+   * Function to inject an export bundle into the iframe
+   * The bundle should be encrypted to the iframe's initial public key
+   * Encryption should be performed with HPKE (RFC 9180).
+   * This is used during export flows.
+   */
+  async injectKeyExportBundle(bundle: string): Promise<boolean> {
```
r-n-o, line 164: Just a note: it'd be nice to have some validation at this level (synchronous) to make sure the bundle has the right length / looks like a good bundle. Doesn't have to be part of this PR because this also affects injectRecoveryBundle :)

r-n-o (approved):
Awesome!

packages/iframe-stamper/src/index.ts
oliviathet: thanks for pointing that out! will add a reminder to self to add validation across the 3 injection functions

PR #148, opened 2023-10-30
update viem to be wallet-compatible

andrewkmin:
## Summary & Motivation
$title

## How I Tested These Changes
- unit
- examples


examples/with-viem/src/createNewWallet.ts
```diff
@@ -0,0 +1,72 @@
...
+      organizationId: process.env.ORGANIZATION_ID!,
+      parameters: {
+        walletName,
+        accounts: [
+          {
+            curve: "CURVE_SECP256K1",
+            pathFormat: "PATH_FORMAT_BIP32",
+            path: "m/44'/60'/0'/0/0",
+            addressFormat: "ADDRESS_FORMAT_ETHEREUM",
+          },
```
r-n-o, line 41: we should probably have SDK-level helpers for these constants at some point:
```ts
accounts: [EthereumAccountParam({"index": 0}), EthereumAccountParam({"index": 1}]
```
Would be pretty sweet!

r-n-o (approved):

examples/with-viem/src/createNewWallet.ts
andrewkmin: absolutely... the more we can abstract some of these details away, the better!

PR #149, opened 2023-10-30
update ethers to be wallet-compatible

andrewkmin:
## Summary & Motivation
$title. also refactored the ethers tests to use 3 potential configurations (wallet account address, private key address, private key ID). also updated many (but not all!) of the examples to use wallet accounts

Examples inventory:
- [x] rebalancer: left as is with private keys -- makes the most sense for now given the need for private key tags, etc
- [x] deployer
- [x] sweeper
- [x] trading-runner: left as is with private keys -- makes the most sense for now given the need for private key tags, etc
- [x] with-ethers: created new examples using wallets, but left private key examples under `/legacy`
- [x] with-nonce-manager
- [x] with-gnosis: create a wallet with 3 wallet accounts 

## How I Tested These Changes
- unit
- examples


examples/with-gnosis/src/createNewWallet.ts
```diff
@@ -0,0 +1,84 @@
...
+          {
+            curve: "CURVE_SECP256K1",
+            pathFormat: "PATH_FORMAT_BIP32",
+            path: "m/44'/60'/0'/0/0",
+            addressFormat: "ADDRESS_FORMAT_ETHEREUM",
+          },
+          {
+            curve: "CURVE_SECP256K1",
+            pathFormat: "PATH_FORMAT_BIP32",
+            path: "m/44'/60'/0'/0/1",
```
timurnkey, line 45: nit: since this is supposed to demonstrate multisig - probably better to update the account index instead of address index, but not a huge deal.

```
path: "m/44'/60'/0'/0/0"
path: "m/44'/60'/1'/0/0"
path: "m/44'/60'/2'/0/0"
```

source: https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#user-content-Account

timurnkey (approved):
🔥 🔥 🔥

examples/with-gnosis/src/createNewWallet.ts
andrewkmin: thanks for this callout!

PR #154, opened 2023-11-01
chore: add esm to package dist

jessecollier:


r-n-o (approved):
Ok finally! Managed to retry enough times for our CI to pass. Still not sure what's going on with these integration tests but we're actively looking. In the meantime, your PR looks good so let's merge this in! Thank you for taking the time to open it / fix it! 🙌

PR #67, opened 2023-05-28
fix get_private_key route

timurnkey:
## Summary & Motivation
beta user found this bug 🐛 

## How I Tested These Changes
e2e

PR #169, opened 2023-11-22
update iframestamper and add an example for email auth

andrewkmin:
## Summary & Motivation
$title

Related: https://github.com/tkhq/frames/pull/19

## Did you add a changeset?

To add a changeset for your pr run [`pnpm changeset`](https://pnpm.io/using-changesets#adding-new-changesets). `pnpm changest` will generate a file where you should write a human friendly message about the changes. Note how this ([example](https://github.com/tkhq/sdk/blob/b409cd06790f011bf939adcf0755499b8e7497ae/.changeset/extra-http-exports.md?plain=1#L1)) includes the package name (should be auto added by the command) along with the type of [semver change (major.minor.patch)](https://semver.org/) (which you should set).


packages/iframe-stamper/src/index.ts
```diff
@@ -135,7 +138,7 @@ export class IframeStamper {
     return new Promise((resolve, reject) => {
       this.iframe.contentWindow?.postMessage(
         {
-          type: IframeEventType.InjectRecoveryBundle,
+          type: IframeEventType.InjectAuthBundle,
```
oliviathet, line 141: did you mean to make this change?

andrewkmin: definitely not! thanks for the catch :)

packages/iframe-stamper/src/index.ts
```diff
@@ -233,6 +236,42 @@ export class IframeStamper {
     });
   }
 
+  /**
+   * Function to inject a new credential into the iframe
+   * The bundle should be encrypted to the iframe's initial public key
+   * Encryption should be performed with HPKE (RFC 9180).
+   * This is used during auth flows.
+   */
+  async injectAuthBundle(bundle: string): Promise<boolean> {
```
r-n-o, line 245: except for the name / purpose, is there a difference between recovery and auth bundles? I'm thinking we should rename the event to "INJECT_CREDENTIAL_BUNDLE" and use it in both recovery and auth use cases?

andrewkmin: yep great idea! made the relevant changes here and in frames here: https://github.com/tkhq/frames/pull/19

still begs the question: got any suggestions for a good name?

r-n-o: What do you think about `InjectCredentialBundle`? Are you happy with it or do you think we should try to find something better?

packages/iframe-stamper/README.md
```diff
@@ -55,3 +55,31 @@ const publicKey = await iframeStamper.init();
 // Injects a new private key in the iframe
 const injected = await iframeStamper.injectKeyExportBundle(exportBundle);
 ```
+
+Auth
+
```
r-n-o, line 60: I think this can be deleted/consolidated with the first section. The code snippet is exactly the same :)

examples/email-auth/src/pages/index.tsx
```diff
@@ -0,0 +1,223 @@
...
+          },
+        ],
+      },
+    });
+
+    const wallet = refineNonNull(completedActivity.result.createWalletResult);
+    const walletId = refineNonNull(wallet.walletId);
+    const address = refineNonNull(wallet.addresses[0]);
+
+    // Instead of simply alerting, redirect the user to your app's login page.
```
r-n-o, line 130: this comment should be deleted. Redirecting is a recovery-only thing

r-n-o (commented):
Looking good with the consolidation!

Let's add a changeset for iframe-stamper since we're breaking its API (`injectRecoveryBundle` won't exist anymore)

packages/iframe-stamper/src/index.ts
andrewkmin: I love `InjectCredentialBundle` ! Sorry, meant a name for the consolidated bundle -- and seems like we're good with `auth`

andrewkmin:
> Looking good with the consolidation!
> 
> Let's add a changeset for iframe-stamper since we're breaking its API (`injectRecoveryBundle` won't exist anymore)

good call! ultimately i'll be holding off on merging/releasing until the next release that ships the email auth activity, just so that protos/etc are consistent

r-n-o (approved):
Looking great! Agree that it's probably best to wait until next release so we can pull in the exact protos!

.changeset/quick-buttons-cross.md
```diff
@@ -0,0 +1,6 @@
+---
+"@turnkey/iframe-stamper": minor
```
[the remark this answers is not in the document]
andrewkmin: no strong feelings on this one. maybe leave as minor/deprecation for now, and then have a stricter change later?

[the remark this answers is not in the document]
r-n-o: Sounds good to me! Let's keep both function names/implementation around then 👍 

(and we can deprecate if we make a breaking change at some point)

PR #171, opened 2023-11-24
fix: issue where esm did not build with fully specified extensions

jessecollier:
## Summary & Motivation

Previously using a tsconfig.esm.json would compile output for esm, however paths were relative with no extensions.

Typically this is solved by using fully specified extensions in imports in TS files. However, this will not work in a dual output module (cjs and esm).

To resolve this, `rollup` is used to compile and export both cjs and esm as well as types. 

Additionally, by default rollup will bundle everything into one file. To preserve file structure, this is disabled. 

Additionally, without specifiying `type: module` in the package json all ES modules need the `mjs` extension to be interpreted by downstream compilers as ES modules.

## Did you add a changeset?

To add a changeset for your pr run [`pnpm changeset`](https://pnpm.io/using-changesets#adding-new-changesets). `pnpm changest` will generate a file where you should write a human friendly message about the changes. Note how this ([example](https://github.com/tkhq/sdk/blob/b409cd06790f011bf939adcf0755499b8e7497ae/.changeset/extra-http-exports.md?plain=1#L1)) includes the package name (should be auto added by the command) along with the type of [semver change (major.minor.patch)](https://semver.org/) (which you should set).


r-n-o (commented):
A couple of things to resolve before merging:
* The js-build step is failing (looks like relative imports don't work anymore? Seems strange)
* We need a changeset entry for this to make sure we properly release packages that are affected. Should be a minor bump for all I believe? `pnpm run changeset` should guide you!

PR #176, opened 2023-11-28
Update federated-passkeys example to use wallets and wallet accounts

r-n-o:
## Summary & Motivation
I'm using this example to test passkey prompts and there were a few things to tweak here:
* example was using private keys rather than wallets
* added instructions to test with real mobile devices
* reworked the CSS to be simpler and mobile-friendly
* added basic error handling

## Screenshots
image
image


examples/with-federated-passkeys/src/pages/api/createSubOrg.ts
```diff
@@ -81,20 +76,25 @@ export default async function createUser(
     });
 
     const subOrgId = refineNonNull(
-      completedActivity.result.createSubOrganizationResultV3?.subOrganizationId
+      completedActivity.result.createSubOrganizationResultV4?.subOrganizationId
```
andrewkmin, line 79: commenting here but this is really for line 70 -- a tiny comment for the path could be helpful for those unfamiliar ("hey, this is the first/initial wallet -- hardcode to account 0")

andrewkmin (approved):
this is BEAUUUUUTIFULLLLL

examples/with-federated-passkeys/src/pages/api/createSubOrg.ts
r-n-o: Good call! I'll make it a constant as well, for clarity.

PR #179, opened 2023-11-28
Switch sha256 to be synchronous in `@turnkey/webauthn-stamper`

r-n-o:
## Summary & Motivation
Bug discovered by @SimonVillage in https://github.com/tkhq/demo-passkey-wallet/issues/3: on ios/safari, cancelling the passkey prompt results in subsequent prompts automatically failing. The reason why: ios is stricter than other browsers when deciding whether a passkey prompt can be triggered or not. [This blog post](https://fy.blackhats.net.au/blog/2020-08-12-user-gesture-is-not-detected-using-ios-touchid-with-webauthn-rs/) has more information on this, but TL;DR: you need to have a maximum of one async promise between a user gesture (tap, click) and the code triggering a passkey prompt.

In our SDK we use [`crypto.subtle.digest`](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) to compute the SHA256 of the message to sign: https://github.com/tkhq/sdk/blob/bed2beaa87fd4b1024207bb450e0847c744a4d2c/packages/webauthn-stamper/src/index.ts#L72

This means we have an async promise between a button click and the call to `navigator.credentials.get`, causing iOS to fail. The fix? Use a synchonous version of sha256. [`@noble/hashes`](https://www.npmjs.com/package/@noble/hashes) is a great one

## How I Tested These Changes
Used the federated passkey example on my localhost, and reached the demo through an ngrok tunnel. I was able to repro the issue to start with, and observe it fixed after I replaced sha256 with the synchronous variant.

### DIFF
--- a/.changeset/breezy-kangaroos-pay.md
+++ b/.changeset/breezy-kangaroos-pay.md
@@ -1,0 +1,5 @@
+---
+"@turnkey/webauthn-stamper": patch
+---
+
+Make sha256 computation synchronous to resolve ios passkey prompt issues
--- a/examples/with-federated-passkeys/src/pages/index.tsx
+++ b/examples/with-federated-passkeys/src/pages/index.tsx
@@ -184,15 +184,21 @@ export default function Home() {
   );
 
   const login = async () => {
-    // We use the parent org ID, which we know at all times,
-    const res = await turnkeyClient.getWhoami({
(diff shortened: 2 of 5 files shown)


oliviathet (approved):
woah really interesting edge case (kinda seems like it could more common than an edge case). either way, this looks great! 🔥 🔥

PR #183, opened 2023-12-07
pass container element to iframe stamper

oliviathet:
## Summary & Motivation
* Use container element instead of id for iframe stamper constructor. This allows for the iframe stamper to be appended onto an element in the shadow DOM.
* Update auth, recovery, export examples.
* Bump `turnkey/iframe-stamper` package major version.

## How I Tested These Changes
* locally ran auth, recovery, export examples

## Did you add a changeset?

### DIFF
--- a/examples/wallet-export/src/components/Export.tsx
+++ b/examples/wallet-export/src/components/Export.tsx
@@ -3,19 +3,31 @@
 import { IframeStamper } from "@turnkey/iframe-stamper";
 import { Dispatch, SetStateAction, useEffect, useState } from "react";
 
-interface RecoveryProps {
+interface ExportProps {
   iframeUrl: string;
   turnkeyBaseUrl: string;
+  iframeDisplay: string;
   setIframeStamper: Dispatch<SetStateAction<IframeStamper | null>>;
 }
 
 const TurnkeyIframeContainerId = "turnkey-iframe-container-id";
 const TurnkeyIframeElementId = "turnkey-iframe-element-id";
 
-export function Recovery(props: RecoveryProps) {
+export function Export(props: ExportProps) {
   const [iframeStamper, setIframeStamper] = useState<IframeStamper | null>(
     null
   );
+  const [iframeDisplay, setIframeDisplay] = useState<string>("none");
+
+  useEffect(() => {
+    setIframeDisplay(props.iframeDisplay);
+    return () => {
+      if (iframeDisplay === "block") {
+        setIframeDisplay("none");
+      }
+    };
+
+  }, [props.iframeDisplay]);
 
   useEffect(() => {
     try {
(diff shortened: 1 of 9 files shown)


oliviathet:
also, i think the test we have in `packages/iframe-stamper/src/__tests__/iframe-test.ts` is irrelevant now because we're directly passing the html element now. will rewrite this test.

packages/iframe-stamper/src/index.ts
```diff
@@ -67,13 +67,7 @@ export class IframeStamper {
...
     }
 
-    const container = document.getElementById(config.iframeContainerId);
-    if (!container) {
-      throw new Error(
-        `Cannot create iframe stamper: no container with ID ${config.iframeContainerId} exists in the current document`
-      );
-    }
-    this.container = container;
+    this.container = config.iframeContainer;
```
r-n-o, line 70: I think the check above (on line 64) needs to be updated, from `document.getElementById(config.iframeElementId)` to ``this.container.querySelector(`#${config.iframeElement}`)``

r-n-o, line 70: Separately: do we need to check for `config.iframeContainer !== undefined`? Or is it already implicit if the type is HTMLElement?

examples/wallet-export/src/pages/index.tsx
```diff
@@ -5,42 +5,19 @@ import { IframeStamper } from "@turnkey/iframe-stamper";
 import * as React from "react";
 import { useEffect, useState } from "react";
 import axios from "axios";
+import { Export } from "@/components/Export";
```
r-n-o, line 8: Clean!! ❤️

examples/wallet-export/src/components/Export.tsx
```diff
@@ -0,0 +1,75 @@
...
+  }, [props.iframeDisplay]);
+
+  useEffect(() => {
+    try {
+      if (!iframeStamper) {
+        const iframeContainer = document.getElementById(TurnkeyIframeContainerId);
+        if (!iframeContainer) {
+          console.error(`Cannot create iframe stamper: no container with ID ${TurnkeyIframeContainerId} exists`);
+          return;
+        }
```
r-n-o, line 39: Is this still necessary given the component is in control of creating / inserting the container into the DOM?

packages/iframe-stamper/src/index.ts
oliviathet: Good point! Looked into this and the `document.getElementById` and `document.querySelector` both return `HTMLElement | null` when the element's not found. However, they can return `undefined` when they're accessed before the element is created. In short, yep I added this check. Thanks!

examples/wallet-export/src/components/Export.tsx
oliviathet: Removed this check here since I made the change above allowing for `iframeContainer` to be passed as null/undefined and the check is done in the iframe stamper's constructor now.

packages/iframe-stamper/src/index.ts
oliviathet: 👀 👀  Fixed, thanks!

r-n-o (approved):
💯

Issue #120, opened 2023-09-18
Move to Node v18 given v16 is officially past its End-of-Life date


r-n-o:
See https://nodejs.org/en/blog/announcements/nodejs16-eol

When running tests with v18 I get a test failure that we'll need to fix:
```
│     TypeError: mockedFetch.mockReturnValue is not a function
│       22 |   response.json = async () => ({});
│       23 |
│     > 24 |   mockedFetch.mockReturnValue(Promise.resolve(response));
│          |               ^
│       25 |
│       26 |   await TurnkeyApi.getWhoami({
│       27 |     body: {
│       at Object.mockReturnValue (src/__tests__/request-test.ts:24:15)
```

This is probably a straightforward fix, just jotting this down here while it's fresh!

drGrove:
Why not just go to node 20, which hits LTS status next month

r-n-o:
Good point. We can target node 20 provided it doesn't make life difficult for SDK users (we need to be careful not to use node 20 specific features; otherwise we'd force all SDK users to upgrade to node 20+)

r-n-o:
This now done, we're using node v18.

PR #184, opened 2023-12-13
Upgrade to node v18

r-n-o:
## Summary & Motivation
Node 16 has been deprecated since September. Time to upgrade! This PR addresses #120. Because node v18 comes with its own experimental fetch API our code which tried to mock `fetch` was failing. I've opted to always use cross-fetch regardless of whether `fetch` is available in `globalThis` because cross-fetch is meant to be used for exactly this reason! It detects when native implementations are available and uses either browser or node fetch when they're available.

Having `fetch` in a known location is nice because we have a seam to mock it.

## How I Tested These Changes
Unit tests

## Did you add a changeset?

Changeset added and packages bumped!


pnpm-lock.yaml
```diff
@@ -28,11 +28,11 @@ importers:
         specifier: workspace:*
         version: link:internal/jest-config
       '@types/node':
-        specifier: ^16.0.0
-        version: 16.18.12
+        specifier: ^18.0.0
+        version: 18.0.0
```
andrewkmin, line 32: if we want to specify a version, we could use `v18.18.2` here as well

andrewkmin (approved):
lgtm!

andrewkmin:
> I've opted to always use cross-fetch regardless of whether fetch is available in globalThis because cross-fetch is meant to be used for exactly this reason! It detects when native implementations are available and uses either browser or node fetch when they're available.

thanks for adding this clarity

pnpm-lock.yaml
r-n-o: Yep! Good call, I'll update it to match the version in `.nvmrc`.

PR #191, opened 2023-12-26
Re-introduce `VERSION` in @turnkey/http

r-n-o:
## Summary & Motivation
This re-introduces the changes from #186 and #187 after they were reverted in #190.

I'm very, very disappointed in [Changesets](https://github.com/changesets/changesets), a tool we're using for releases. `npm` itself has plenty of lifecycle hooks to handle custom scripts pre and post version (see [this](https://docs.npmjs.com/cli/v9/using-npm/scripts)). But these lifecycle hooks aren't called by `changeset version`! Wut. People are complaining about this...
* https://github.com/pnpm/pnpm/issues/2911
* https://github.com/changesets/changesets/issues/1166
* https://github.com/changesets/changesets/issues/1145

...but no responses from the maintainers.

I tried instrumenting `preversion`, `prepublish`, `version`, etc...nothing seems to work because `changeset` doesn't call any of these. In the end I settled for a custom "version" script. This has the advantage to be explicit but it's one more script to call during releases. The problem: what if we forget to call it when we publish new versions? For this I've got a CI check which should catch any oversight.

## How I Tested These Changes
* Ran version script locally to make sure it works
* Broke things on purpose with [this commit](https://github.com/tkhq/sdk/pull/191/commits/d7ed9b8e7d43aa74ac9a016e16387c186deb8e99) to trigger the CI check
  image

## Did you add a changeset?
Yes! ✅

### DIFF
--- a/.changeset/late-rivers-play.md
+++ b/.changeset/late-rivers-play.md
@@ -1,5 +1,0 @@
----
-"@turnkey/http": patch
----
-
-Export VERSION from turnkey/http
--- a/.changeset/late-rivers-play.md
+++ b/.changeset/late-rivers-play.md
@@ -1,0 +1,5 @@
+---
+"@turnkey/http": patch
+---
+
+Export VERSION from turnkey/http
--- a/.github/workflows/js-build.yml
+++ b/.github/workflows/js-build.yml
@@ -42,6 +42,11 @@ jobs:
       - name: Prettier
         run: pnpm run prettier-all:check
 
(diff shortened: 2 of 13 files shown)


emostov (approved):
Nice - we can include running the `pnpm version` command in release instructions

r-n-o:
> Nice - we can include running the `pnpm version` command in release instructions

Yep! That's my next PR once this one is merged! 🙏

PR #195, opened 2024-01-09
New Turnkey signer: `@turnkey/solana`

r-n-o:
## Summary & Motivation
This is an attempt at wrapping signing logic with Solana. 

The example code for signing with Solana (`examples/with-solana/src/createSolanaTransfer.ts`) is now a lot shorter / cleaner, I think that's a win:
image

Open to feedback on naming and structure of course; unfortunately `@solana/web3.js` doesn't have ready-to-use abstract signer interfaces so this is a wrapper rather than a concrete implementation of a supported abstraction.

I've briefly looked at https://github.com/solana-labs/solana-web3.js/tree/master/packages/signers but this is experimental (only committed 2 months ago) so unlikely to be useful right away. We don't want to force people to upgrade all the way to experimental releases . FWIW `@solana/signers` takes the same approach than what I have here and creates wrappers. If we were to go this route we'd choose to implement the `TransactionPartialSigner` and `MessagePartialSigner` interfaces. Pretty easy to add down the road if we want to!

I've also seen https://github.com/solana-labs/wallet-adapter: seems like a useful library used by many projects. I think we should try to implement our own "wallet" in https://github.com/solana-labs/wallet-adapter/tree/master/packages/wallets? Not sure if that'd be useful?

Oh and: this branch also updates the Solana demo to work with wallets rather than private keys!

## How I Tested These Changes
* New unit tests (created a fresh production org to test against -- I've set the CI variable already)
* Ran the `with-solana` demo manually


packages/solana/src/index.ts
```diff
@@ -0,0 +1,54 @@
...
+    this.client = input.client;
+  }
+
+  /**
+   * This function takes a Solana transaction and adds a signature with Turnkey
+   *
+   * @param tx Transaction object (native @solana/web3.js type)
+   * @param fromAddress Solana address (base58 encoded)
+   */
+  public async addSignature(tx: Transaction, fromAddress: string) {
```
andrewkmin, line 19: supporting just `addSignature` makes sense given this is all a user needs support for, at least for now. (when we implement a signer interface, then things might change as we support `sign` and `partialSign`, etc)

andrewkmin (approved):
looks good to me! functional, and definitely a whole lot cleaner abstraction for users :)

examples/with-solana/src/index.ts
```diff
@@ -92,7 +88,7 @@ async function main() {
     toAddress: destination,
     amount: Number(amount),
     turnkeyOrganizationId: organizationId,
-    turnkeyPrivateKeyId: privateKeyId,
+    turnkeySolAddress: solAddress,
```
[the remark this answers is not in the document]
r-n-o: Oh right! That's a good call. Let's remove `turnkeySolAddress` 👍

packages/solana/src/index.ts
r-n-o: Yeah definitely agree. We can extend the interface as needed in the future. There's also the "sign message" functionality; this can be added fairly easily but I've left it out for now since it's not been requested just yet.

PR #197, opened 2024-01-16
bump viem

andrewkmin:
## Summary & Motivation
bump our upstream dependency, https://github.com/wevm/viem.

note that there aren't many that affect our turnkey viem package directly; most pertain to OP chains, cryptography, etc. however, there are some minor interface changes: see changes to examples/tests.

additionally, you can find all releases here: https://github.com/wevm/viem/releases

## How I Tested These Changes
- unit
- examples

## Did you add a changeset?
yeeeeees


packages/viem/package.json
```diff
@@ -52,7 +52,7 @@
     "compile:contracts": "hardhat compile"
   },
   "peerDependencies": {
-    "viem": "^1.16.6"
+    "viem": "^2.1.1"
   },
```
r-n-o, line 56: Is there a way to be lax here? (basically declaring: "turnkey/viem is compatible with 1.x or 2.x")

I'm afraid that consumers of turnkey/viem who do not want to be on 2.x won't be able to use the new turnkey/viem.

r-n-o: Something like `"viem": "^1.16.6 || ^2.1.1"`?

andrewkmin: good call! https://github.com/tkhq/sdk/pull/197/commits/850634fb27b2c2dfc0b6e1c63bc5d5fe9c7971ba

i bumped the with-viem example's viem dep back down to 1.16.6, and all seems to be well.

r-n-o: You can bump the with-viem example to 2.x I think!

Either way we won't have everything covered: if you leave it at 1.x we ensure compatibility with Viem 1.x, if you bump to 2.x we test for viem 2.x -- either way we lose one side :)

r-n-o (approved):

PR #200, opened 2024-01-21
Taylor/ethers v6

taylorjdawson:
## Summary & Motivation
Updates the @turnkey/ethers package to use ethers v6.

## How I Tested These Changes
Ran all examples that were effected by these changes to ensure they behaved as they did with ether v5. Additionally, ran the tests contained in the @turnkey/ethers package to ensure they are passing.

## Did you add a changeset?


andrewkmin (commented):
lookin niiiiiice! just a few comments thus far~

andrewkmin:
very minor thing to mention: we have a very, very barebones section of our docs that covers troubleshooting for our SDK. once ethers v6 is shipped, we can probably update https://docs.turnkey.com/sdk-troubleshooting#using-sepolia-for-testing as well!

packages/ethers/src/index.ts
```diff
@@ -145,31 +146,30 @@ export class TurnkeySigner extends ethers.Signer implements TypedDataSigner {
...
     return `0x${signedTx}`;
   }
 
   // Returns the signed prefixed-message. Per Ethers spec, this method treats:
   // - Bytes as a binary message
   // - string as a UTF8-message
   // i.e. "0x1234" is a SIX (6) byte string, NOT 2 bytes of data
-  async signMessage(message: string | Bytes): Promise<string> {
-    const hashedMessage = ethers.utils.hashMessage(message);
+  async signMessage(message: string | Uint8Array): Promise<string> {
```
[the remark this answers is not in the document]
taylorjdawson: I was basing it on the `signMessage` method on the [AbstractSigner](https://github.com/ethers-io/ethers.js/blob/a26ff77c210660e641455150ccea6c0530f02f61/src.ts/providers/abstract-signer.ts#L239) class:
```
abstract signMessage(message: string | Uint8Array): Promise<string>;
```


but looks like `ByteLike` is equivalent? The only difference being in the nomenclature:
```
export type DataHexString = string;
/**
 *  A string which is prefixed with ``0x`` and followed by any number
 *  of case-agnostic hexadecimal characters.
 *
 *  It must match the regular expression ``/0x[0-9A-Fa-f]*\/``.
 */
export type HexString = string;
/**
 *  An object that can be used to represent binary data.
 */
export type BytesLike = DataHexString | Uint8Array;
```
([source](https://github.com/ethers-io/ethers.js/blob/a26ff77c210660e641455150ccea6c0530f02f61/lib.esm/utils/data.d.ts#L5-L16))

I am not partial to either, your call!

packages/ethers/src/index.ts
```diff
@@ -260,6 +259,19 @@ export class TurnkeySigner extends ethers.Signer implements TypedDataSigner {
   }
 
   _signTypedData = this.signTypedData.bind(this);
+
+  override async sendTransaction(tx: TransactionRequest): Promise<TransactionResponse> {
```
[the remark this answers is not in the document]
taylorjdawson: In upgrading from v5 to v6, I had to override `sendTransaction` due to a nuanced type issue between `AbstractSigner` and `Transaction`:

The core of the issue is that `sendTransaction` generates a `Transaction` instance via `Transaction.from()`, which is then passed to `signTransaction`. 
Although `signTransaction` is designed to accept a `TransactionRequest`, the structural similarities between `TransactionRequest` and `Transaction` mean TypeScript doesn't identify this as an error, despite a crucial difference: `Transaction`'s properties are private. 
This distinction becomes problematic when `resolveProperties` is invoked on the `TransactionRequest` within `signTransaction`, leading to an empty object because `Object.keys()` cannot access `Transaction`'s private properties. 
Importantly, `TransactionRequest` may contain properties as promises, necessitating the use of `resolveProperties` to ensure proper handling. 
This subtlety in how `Transaction` instances and `TransactionRequest` objects are treated, especially regarding promise properties and the need for `resolveProperties`, indicates a potential area for improvement in ethers library's type system.

examples/deployer/src/index.ts
```diff
@@ -38,18 +38,19 @@ async function main() {
 
   // Connect it with a Provider (https://docs.ethers.org/v5/api/providers/)
```
andrewkmin, line 39: we can probably update/remove this comment now! :)

examples/deployer/src/index.ts
```diff
@@ -38,18 +38,19 @@ async function main() {
...
   const network = "goerli";
```
andrewkmin, line 40: side note: we could probably start moving these examples all over to `sepolia` . totally unrelated to this PR; feel free to leave as is!

examples/deployer/src/index.ts
```diff
@@ -38,18 +38,19 @@ async function main() {
...
-  const provider = new ethers.providers.InfuraProvider(network);
-  const connectedSigner = turnkeySigner.connect(provider);
+  const connectedSigner = turnkeySigner.connect(
+    new ethers.InfuraProvider(network)
+  );
 
-  const chainId = await connectedSigner.getChainId();
+  const chainId = (await connectedSigner.provider?.getNetwork())?.chainId ?? 0;
```
andrewkmin, line 45: i feel like we could make provider a hard requirement (`.provider!`)

packages/ethers/src/index.ts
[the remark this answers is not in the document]
andrewkmin: ah totally makes sense! let's stick with what you have. 

but if we did want to use `BytesLike`, it includes `string`, so it would simply be `signMessage(message: BytesLike)`

packages/ethers/src/index.ts
```diff
@@ -133,19 +142,19 @@ export class TurnkeySigner extends ethers.Signer implements TypedDataSigner {
     return signedTx;
   }
 
-  async signTransaction(
-    transaction: ethers.utils.Deferrable<ethers.providers.TransactionRequest>
-  ): Promise<string> {
-    const unsignedTx = await ethers.utils.resolveProperties(transaction);
-
+  async signTransaction(transaction: TransactionRequest): Promise<string> {
```
andrewkmin, line 145: i think i found another example of someone who tried to use an external signer with ethers v6! https://github.com/ethers-io/ext-signer-ledger/blob/b59a757be7c6b0a4db823ab2903df70f8c52d0af/src.ts/signer-ledger.ts#L84

taking some heavy inspiration from their solution, but we could end up with something like:

```
import { copyRequest, resolveProperties, resolveAddress } from "ethers";

...

async signTransaction(transaction: TransactionRequest): Promise<string> {
    transaction = copyRequest(transaction);
    const { to, from } = await resolveProperties({
      to: transaction.to
        ? resolveAddress(transaction.to, this.provider)
        : undefined,
      from: transaction.from
        ? resolveAddress(transaction.from, this.provider)
        : undefined,
    });

    if (to != null) {
      transaction.to = to;
    }
    if (from != null) {
      transaction.from = from;
    }

    const tx = Transaction.from(<TransactionLike<string>>transaction);
    const unsignedTx = tx.unsignedSerialized.substring(2);

    const signedTx = await this._signTransactionWithErrorWrapping(
      unsignedTx
    );

    return `0x${signedTx}`;
  }
```

the `copyRequest` allows us to maintain all the necessary pieces of the tx! and excitingly, this means that we'd no longer have to override `sendTransaction`!

andrewkmin (commented):
what a diff! 👏😮‍💨 real MVP for cleaning up all the examples as well. this looks great and works! but i think we can end up simplifying `signTransaction` a tiny bit 😄

examples/deployer/src/index.ts
taylorjdawson: Probably not as important for examples but I think best practice is to only use `!` when you know that the value can’t be null or undefined. But I wonder if the types should be adjusted in the signer so that provider isn't undefined? 

```
constructor(config: TConfig, provider?: Provider) {
    super(provider);

    this.client = config.client;
    this.organizationId = config.organizationId;
    this.signWith = config.signWith;
  }

  connect(provider: Provider): TurnkeySigner {
    return new TurnkeySigner(
      {
        client: this.client,
        organizationId: this.organizationId,
        signWith: this.signWith,
      },
      provider
    );
  }
```

It wasn't totally clear to me the reason for having the `connect` method and the constructor. But if we made the provider a hard requirement in the constructor then users wouldn't need to check if the provider is defined on the signer before using it. Is that change possible? Or am I missing context on the usage of the `connect` method?

packages/ethers/src/index.ts
taylorjdawson: ooo I like this a lot, _much_ cleaner!

taylorjdawson: I kept the ens resolveAddress methods as I feel that this could be useful! I can remove if you think otherwise

examples/deployer/.env.local.example
```diff
@@ -2,5 +2,5 @@ API_PUBLIC_KEY="<Turnkey API Public Key (that starts with 02 or 03)>"
 API_PRIVATE_KEY="<Turnkey API Private Key>"
 BASE_URL="https://api.turnkey.com"
 ORGANIZATION_ID="<Turnkey organization ID>"
-SIGN_WITH="<Turnkey Wallet Account Address, Private Key Address, or Private Key ID>" # if blank, we will create a wallet for you
+PRIVATE_KEY_ID="<Turnkey (crypto) private key ID>" # if you leave it blank, we'll create one for you via calling the Turnkey API
```
andrewkmin, line 5: i think we may still need `SIGN_WITH`, since that's what's used in the primary script https://github.com/tkhq/sdk/blob/9041b6a57f515ea8ba2105150ee5bbe251443c1a/examples/deployer/src/index.ts#L36.

generally, i think we can move completely off of `PRIVATE_KEY_ID` in most of these examples, which means we could delete `/createNewEthereumPrivateKey.ts` from most places as well. don't need to comb through every example now, but perhaps for now we can leave this env var as `SIGN_WITH`.

andrewkmin (approved):
looks amazing 👏 overall, green light! 🟢 

i think it would be valuable at some point to individually go through each example. but that doesn't have to be today; i've tested your updated `with-ethers` using v6, and all's well! once released, we can also update https://github.com/tkhq/demo-ethers-passkeys. but this, too, is pretty low priority.

as for the Socket Security check, i'm overall not too concerned. those deps seem to mainly be concerned with the `with-gnosis` example, which we've seen has some wonky stuff going on (requiring us to add

```
...
"tsBuildInfoFile": "./.cache/.tsbuildinfo",
"noUnusedLocals": false,
"noUnusedParameters": false
...
```

to the tsconfig.

one last thing: can you include a changelog (`pnpm changeset`)? we can mention that the overall interface should not change for end-users. we can also link to https://docs.ethers.org/v6/migrating/

andrewkmin:
side observation: looks like ethers v7 is on the horizon 😮 https://github.com/ethers-io/ethers.js/discussions/4312

still not in beta yet or anything like that, and apparently should be pretty painless! but we might have to get back in here within the next ~6 months - 1 year

.changeset/fast-poems-beg.md
```diff
@@ -0,0 +1,66 @@
+---
+"@turnkey/ethers": minor
+---
+
+Updates @turnkey/ethers package and examples to use ethers v6. Refer to https://docs.ethers.org/v6/migrating for full migration instructions.
+
+✨Summary of Changes✨
+
+- `getBalance` iis no longer a method on the signer. It must be obtained via the provider instance.
```
andrewkmin, line 9: tiny nit: `is`

.changeset/fast-poems-beg.md
```diff
@@ -0,0 +1,66 @@
...
+// first get the address
+const address = await connectedSigner.getAddress()
+// then pass it in
+const transactionCount = await connectedSigner.provider?.getTransactionCount(address);
+```
+
+- `getFeeData` is no longer a method on the signer. It must be obtained via the provider instance:
+
+```
+//before
```
andrewkmin, line 50: tiny nit: add a space --> `// before` and for a couple places below

.changeset/fast-poems-beg.md
```diff
@@ -0,0 +1,66 @@
+---
+"@turnkey/ethers": minor
+---
+
+Updates @turnkey/ethers package and examples to use ethers v6. Refer to https://docs.ethers.org/v6/migrating for full migration instructions.
+
+✨Summary of Changes✨
+
+- `getBalance` iis no longer a method on the signer. It must be obtained via the provider instance.
+  Additionally, it requires that an address to be passed in:
```
andrewkmin, line 10: can remove the `that`

.changeset/fast-poems-beg.md
```diff
@@ -0,0 +1,66 @@
...
+
+```
+//before
+const feeData = await connectedSigner.getFeeData();
+
+//after
+const feeData = await connectedSigner.provider?.getFeeData();
+```
+
+- BigNumber -> bigint: numerical values such as, chainId, fee data, balance now use new js primitive `bigint` instead of `BigNumber`.
```
andrewkmin, line 57: nit: "...such as chainId, fee data, and balance now use the new ES6 primitive `bigint`..."

.changeset/fast-poems-beg.md
andrewkmin, line 9: because each of these changes comes with a code snippet, i think it's fine formatting-wise if we don't use bullet points at all! but very subjective here :)

andrewkmin (approved):
thanks for being so thorough with the changeset 🏅 

looks like you may need to regenerate the pnpm lockfile but i'll re-approve it right away!

:shipit:

andrewkmin (approved):

PR #202, opened 2024-01-22
implement signMessage on Solana TurnkeySigner

montycheese:
## Summary & Motivation
- Give sign message functionality to solana turnkey wallets

## How I Tested These Changes
- Manually tested using the unit test and hard coded api keys

## Did you add a changeset?
- yes

### DIFF
--- a/packages/solana/src/__tests__/index-test.ts
+++ b/packages/solana/src/__tests__/index-test.ts
@@ -3,8 +3,14 @@ import { TurnkeySigner } from "../";
 import { TurnkeyClient } from "@turnkey/http";
 import { ApiKeyStamper } from "@turnkey/api-key-stamper";
 import { PublicKey, SystemProgram, Transaction } from "@solana/web3.js";
+import nacl from "tweetnacl";
+import bs58 from "bs58";
 
 describe("TurnkeySigner", () => {
+  const organizationId = "4456e4c2-e8b5-4b93-a0cf-1dfae265c12c";
+  const apiPublicKey =
+    "025d374c674fc389c761462f3c59c0acabdcb3a17c599d9e62e5fe78fe984cfbeb";
+  const turnkeySolAddress = "D8P541wwnertZTgDT14kYJPoFT2eHUFqjTgPxMK5qatM";
   test("can sign a Solana transfer against production", async () => {
     if (!process.env.SOLANA_TEST_ORG_API_PRIVATE_KEY) {
       // This test requires an env var to be set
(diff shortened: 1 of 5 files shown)


packages/solana/src/__tests__/index-test.ts
```diff
@@ -49,4 +50,33 @@ describe("TurnkeySigner", () => {
...
+    const signer = new TurnkeySigner({
+      organizationId,
+      client,
+    });
+
+    const signature = await signer.signMessage(
+      "Hello world",
+      turnkeySolAddress
+    );
+    expect(signature).toBeDefined();
```
andrewkmin, line 80: it would be great to be able to verify the signature on the spot

andrewkmin (approved):
lgtm! thanks for making this PR 🙏 

one tiny thing to call out is that some `signMessages` implementations (e.g. on [solana web3](https://github.com/solana-labs/solana-web3.js/blob/a81a78e196b8946c92fd1532a03a57144acc9508/packages/signers/src/keypair-signer.ts#L42-L43)) are expected to return bytes/`Uint8Array`

montycheese:
> lgtm! thanks for making this PR 🙏 
> 
> 
> 
> one tiny thing to call out is that some `signMessages` implementations (e.g. on [solana web3](https://github.com/solana-labs/solana-web3.js/blob/a81a78e196b8946c92fd1532a03a57144acc9508/packages/signers/src/keypair-signer.ts#L42-L43)) are expected to return bytes/`Uint8Array`

I can return a byte array instead of string to be more generic?

r-n-o:
> > lgtm! thanks for making this PR 🙏
> > one tiny thing to call out is that some `signMessages` implementations (e.g. on [solana web3](https://github.com/solana-labs/solana-web3.js/blob/a81a78e196b8946c92fd1532a03a57144acc9508/packages/signers/src/keypair-signer.ts#L42-L43)) are expected to return bytes/`Uint8Array`
> 
> I can return a byte array instead of string to be more generic?

Yup! That'd be perfect. Also, if we make this change, then I believe the `signMessage` function should accept bytes as input rather than `string`? If you want to keep the convenience of signing utf8-strings then maybe we a separate `signText` function would make sense?

packages/solana/src/__tests__/index-test.ts
```diff
@@ -48,4 +51,42 @@ describe("TurnkeySigner", () => {
...
     expect(transferTransaction.signatures.length).toBe(1);
   });
+
+  test("can sign a message with a Solana account", async () => {
+    if (!process.env.SOLANA_TEST_ORG_API_PRIVATE_KEY) {
+      // This test requires an env var to be set
+      console.warn(
+        "This test is skipped because it cannot run without SOLANA_TEST_ORG_API_PRIVATE_KEY set"
+      );
+      return;
```
r-n-o, line 61: Can you make this a hard failure? I made that change already on lines 11-14 so you can copy/paste:
```js
throw new Error(
  "This test requires SOLANA_TEST_ORG_API_PRIVATE_KEY to be set"
);
```

packages/solana/src/__tests__/index-test.ts
```diff
@@ -48,4 +51,42 @@ describe("TurnkeySigner", () => {
...
+      messageAsUint8Array,
+      turnkeySolAddress
+    );
+
+    const isValidSignature = nacl.sign.detached.verify(
+      messageAsUint8Array,
+      signature,
+      bs58.decode(turnkeySolAddress)
+    );
+    expect(isValidSignature).toBeTruthy();
```
r-n-o, line 90: NICE! 💯

packages/solana/src/index.ts
```diff
@@ -20,14 +21,51 @@ export class TurnkeySigner {
...
+    );
+    return Buffer.from(
+      `${signRawPayloadResult.signRawPayloadResult?.r}${signRawPayloadResult.signRawPayloadResult?.s}`,
+      "hex"
+    );
+  }
+
+  private async signRawPayload(
+    payload: string,
+    encoding: definitions["v1PayloadEncoding"],
```
r-n-o, line 58: You should be able to remove this now that both signMessage and addSignature use the same encoding 🎉

r-n-o (approved):

PR #209, opened 2024-01-30
New package: `@turnkey/react-native-passkey-stamper`

r-n-o:
## Summary & Motivation
This includes a new package `@turnkey/react-native-passkey-stamper` which wraps https://github.com/f-23/react-native-passkey and uses it to produce a Turnkey-compatible `Stamper` (which can be used to make authenticated calls the Turnkey API)

## How I Tested These Changes
- [x] unit tests for the parts that are relevant (it's hard to test that we're calling the right things; we'd have to mock 100% of what we're calling?)
- [x] tested this package inside of a real react native app at https://github.com/r-n-o/passkeyapp


packages/react-native-passkey-stamper/src/index.ts
```diff
@@ -0,0 +1,155 @@
...
+/**
+ * Creates a passkey and returns authenticator params
+ */
+export async function CreatePasskey(
+  config: TPasskeyRegistrationConfig
+): Promise<TurnkeyAuthenticatorParams> {
+  const arr = new Uint8Array(32);
+  crypto.getRandomValues(arr);
+  // TODO: this is probably not right? Need to find the right way to encode random bytes into a string
+  const challenge = new TextDecoder("utf-8").decode(arr);
```
oskarleonard, line 69: - Seems like `crypto` and `TextDecoder` is not available in react-native (testing on android emulator).

packages/react-native-passkey-stamper/src/index.ts
```diff
@@ -0,0 +1,155 @@
...
+
+const defaultTimeout = 5 * 60 * 1000; // five minutes
+const defaultUserVerification = "preferred";
+
+/**
+ * Creates a passkey and returns authenticator params
+ */
+export async function CreatePasskey(
+  config: TPasskeyRegistrationConfig
+): Promise<TurnkeyAuthenticatorParams> {
```
oskarleonard, line 65: `TurnkeyAuthenticatorParams`  contains a lot more fields than is used in this function. For example:

  // Optional challenge. If not provided, a new random challenge will be generated
  challenge?: string;
  
I cannot see that we pass along `challenge` from config to `Passkey.register({`.

Only thing used seems to be `config.user`, `config. rpId `, `config. rpName `?

Maybe we can use similar interface as [this](https://github.com/f-23/react-native-passkey/blob/17184a1b1f6f3ac61e07aa784c9b64efb28b570e/src/__tests__/testData/RegRequest.json), set our defaults and spread the rest? So we would need to change rpName and rpId to be `rp` object. 

Something like:

```
const {challenge, ...rest} = config;

  const registrationResult = await Passkey.register({
    challenge: challenge || getRandomChallenge(),
    ...rest,
    pubKeyCredParams: pubKeyCredParams || [
      {
        type: "public-key",
        alg: -7,
      },
      {
        type: "public-key",
        alg: -257,
      },
    ],
  });
  
```

packages/react-native-passkey-stamper/src/index.ts
oskarleonard: Ah, we just need some polyfill to get this working. Maybe could add to readme

### crypto
1. `yarn add react-native-get-random-values`
2. On react-native versions below 0.6 you need to do this `react-native link react-native-get-random-values`
3. `import 'react-native-get-random-values';`  at the top of your entry file (e.g., index.js):

### TextDecoder
1. yarn add text-encoding
2. Add this to your entry point, e.g., index.js
```
import { TextDecoder, TextEncoder } from 'text-encoding';

global.TextDecoder = global.TextDecoder || TextDecoder;
global.TextEncoder = global.TextEncoder || TextEncoder;
```

r-n-o: Thanks for looking @oskarleonard! I think it makes sense to load polyfills for textEncoder / crypto.getRandomValues

@Balanced02 let's pull this in a different file, you're right! 👍

packages/react-native-passkey-stamper/src/index.ts
r-n-o: That's a good catch, I forgot to make the challenge generation optional!

For pubKeyCredParams I actually think this shouldn't be overridable because these are the only algorithms Turnkey supports right now.

> `TurnkeyAuthenticatorParams` contains a lot more fields than is used in this function

This is the return type of this function. Do you mean `PasskeyRegistrationRequest` instead? (if so, agreed: things like excludeCredential wouldn't be overridable). Also, I should probably take an extra options param so that `withSecurityKey` can be set!

oskarleonard: Oh, i forgot removing that. `pubKeyCredParams` shouldnt be optional (and it is not in the ts type either so dont know why i added it :P).

packages/react-native-passkey-stamper/src/index.ts
oskarleonard: We added these polyfills for our app, so that is no problem for us anymore. Just a FYI, that this package should probably add these instructions to a readme. 

What i mean to say, i dont know if it is best that the polyfills are in this package or if the apps using this package should need to add those polyfills to make this package work.

r-n-o: Addressed in https://github.com/tkhq/sdk/pull/209/commits/0a15f9e757382ab9b7aa058d27481c78672c7c5a

packages/react-native-passkey-stamper/src/index.ts
r-n-o: Addressed in https://github.com/tkhq/sdk/pull/209/commits/b98d1f43641afa9016577c9be555731e539cef70, let me know what you think! I went with the "explicit" approach so that each param can be documented and tweaked if needed.

packages/react-native-passkey-stamper/README.md
```diff
@@ -0,0 +1,62 @@
+# @turnkey/react-native-passkey-stamper
+
+[![npm](https://img.shields.io/npm/v/@turnkey/react-native-passkey-stamper?color=%234C48FF)](https://www.npmjs.com/package/@turnkey/react-native-passkey-stamper)
+
+This package contains a React Native passkey stamper. It uses [`react-native-passkey`](https://github.com/f-23/react-native-passkey) to do the heavy lifting. This stamper is meant to be used with [`@turnkey/http`](https://www.npmjs.com/package/@turnkey/http)
+
+## Installation
+
+- Install both [`react-native-passkey`](https://www.npmjs.com/package/react-native-passkey) and `@turnkey/react-native-passkey-stamper` (this package) in your React Native project.
+- Set up provisioning correctly to make sure you app is signed (needed for passkey functionality)
```
andrewkmin, line 10: tiny nit: `your`

packages/react-native-passkey-stamper/README.md
```diff
@@ -0,0 +1,62 @@
...
+## Usage
+
+### Create a new passkey
+
+```ts
+import { createPasskey } from "@turnkey/react-native-passkey-stamper";
+
+// Returns authenticator params that can be used with sub-org creation, user creation, etc.
+const authenticatorParams = await createPasskey({
+  // Won't be visible to users, this the name of the Turnkey resource
```
andrewkmin, line 23: tiny nit: `...this *is* the name...`

packages/react-native-passkey-stamper/src/__tests__/util-test.ts
```diff
@@ -0,0 +1,26 @@
+import { test, expect } from "@jest/globals";
+import { getChallengeFromPayload, getRandomChallenge } from "../util";
+
+test("generates correct challenges from payloads", async function () {
+  // You can verify that this is the right value with:
```
andrewkmin, line 5: 👏

packages/react-native-passkey-stamper/src/index.ts
```diff
@@ -0,0 +1,207 @@
...
+ */
+export type TurnkeyAuthenticatorParams =
+  TurnkeyApiTypes["v1AuthenticatorParamsV2"];
+
+/**
+ * Header name for a webauthn stamp
+ */
+const stampHeaderName = "X-Stamp-Webauthn";
+
+export type TPasskeyRegistrationConfig = {
```
andrewkmin, line 27: this is some juicy goodness! 🙌 

another tiiiny nit, and one that's ultimately personal: can we add an extra new line in between each "stanza"/param in this struct? just for readability 😄

packages/react-native-passkey-stamper/src/index.ts
```diff
@@ -0,0 +1,207 @@
...
+  attestation?: string;
+  // Optional extensions. Defaults to empty.
+  extensions?: Record<string, unknown>;
+};
+
+export type TPasskeyStamperConfig = {
+  // The RPID ("Relying Party ID") for your app.
+  // See https://github.com/f-23/react-native-passkey?tab=readme-ov-file#configuration to set this up.
+  rpId: string;
+  // Optional timeout value. Defaults to 5 minutes.
```
andrewkmin, line 74: in `ms`?

packages/react-native-passkey-stamper/src/index.ts
```diff
@@ -0,0 +1,207 @@
...
+      stampHeaderName: stampHeaderName,
+      stampHeaderValue: JSON.stringify(stamp),
+    };
+  }
+}
+
+/**
+ * Simple util to convert a base64-encoded string to base64url
+ */
+function base64Tobase64url(s: string): string {
```
andrewkmin, line 205: fine with leaving it since it's only used here, but i suppose this is candidate for moving into `src/util.ts`?

packages/react-native-passkey-stamper/src/util.ts
```diff
@@ -0,0 +1,45 @@
...
+  return hexBuffer.toString("base64");
+}
+
+// Function to return 32 random bytes encoded as hex
+// (e.g "5e4c2c235fc876a9bef433506cf596f2f7db19a959e3e30c5a2d965ec149d40f")
+// This function doesn't return strong cryptographic randomness (Math.random is a PRNG), but this is good enough for registration challenges.
+// If the challenge was not random at all the risk is that someone can replay a previous signature to register an authenticator they don't own.
+// However:
+// - we are creating a brand new authenticator here, which means keygen is happening right as we call this library (which makes the replay attack hard-to-impossible)
+// - even if a replay attack went through, the authenticator wouldn't be usable given Turnkey has anti-replay in place in activity payloads (timestamp field)
```
andrewkmin, line 18: optional: can link to the small blurb we have! 
https://docs.turnkey.com/api-introduction#queries-and-submissions

though on the flip side, there's no guarantee this link will live forever as-is.

andrewkmin (approved):
🙇👏 m.v.p.

packages/react-native-passkey-stamper/src/index.ts
r-n-o: Yep! I'll add 👍

packages/react-native-passkey-stamper/src/index.ts
r-n-o: Oh right! That's a good call. I'll move it.

packages/react-native-passkey-stamper/src/index.ts
r-n-o: Yup! You got it 👍

packages/react-native-passkey-stamper/src/util.ts
r-n-o: Let's do it 👍

andrewkmin (approved):
🚢

PR #212, opened 2024-02-07
React native support for `@turnkey/api-key-stamper` + introduction of `@turnkey/encoding`

r-n-o:
## Summary & Motivation
@turnkey/api-key-stamper doesn't work well on React Native because it defaults to the NodeJS implementation. But `crypto` isn't a thing so it errors out. To solve this I'm introducing a "pureJS" implementation which works in React Native and any environment not falling in the ("crypto browser" | "node") camp. The pureJS implementation should be self-contained, no dependencies on the environment whatsoever; should be a safe fallback!

In the same spirit:
* I got rid of `btoa` because it's not available in React Native either (and easy enough to implement ourselves)
* swapped out `buffer` and `noble/hashes` for the minimal [`sha256-uint8array`](https://github.com/kawanet/sha256-uint8array) which only has sha256 functionality and works great. It's not a very popular library but it's easy enough to audit since it's [a single file](https://github.com/kawanet/sha256-uint8array/blob/main/lib/sha256-uint8array.ts)
* as a result of ^^ we don't have the need for `universal.ts` 🎉 

EDIT: this PR is growing slightly in scope: I've introduced `@turnkey/encoding` to consolidate some utility functions common to multiple packages. This was introduced to share the `btoa` implementation across api-key-stamper and http. I'm making this new util package public because I don't know how to do it otherwise. See https://github.com/lmc-eu/spirit-design-system/issues/1238 for why this is hard.

## How I Tested These Changes
- [x] Unit tests
- [x] Import in a real react native app and remove polyfills
- [x] Import packages from codesandbox in our demo passkey wallet to make sure things are all good

## Did you add a changeset?
Changeset added, and ran `pnpm run changeset version` for release! ✅


andrewkmin (approved):

andrewkmin (approved):
👏👏

PR #221, opened 2024-03-06
Reworked SDK Base HTTP Client

AlwaysBCoding:
## Summary & Motivation
Motivation is to begin aggregating our javascript SDK libraries into a single importable SDK with nicer user abstractions for calling common Turnkey API functionality

## Structure
This PR introduces a new package called `sdk-js` into the `packages/` folder. The package requires two `__inputs__` which are the copied `public_api.swagger.json` file and the `public_api.types.json` from the `mono` repo.

There is a single script `node scripts/codegen.js` which will generate the base sdk client and associated types and save it in the `__generated__` folder.

Finally there is a top-level `sdk-client.ts` file that will contain the exported SDK client and the ability to create synthetic methods for users to interact with that are abstractions on top of just calling the underlying generated API methods.

## What is different than the existing base HTTP Client
1. The syntax for calling methods on the Turnkey client is nicer (you don't need to pass an activity type, timestamp and organizationId into every function call)
2. The TurnkeyClient is initialized with a single config object which can be used to configure anything about the http client functionality (i.e. activity poller polling duration or base api url)
3. Any commands automatically poll for the activity creation before returning a result to the user, you don't need to use an activity poller
4. There is a workflow for extending the existing base client with custom methods


packages/sdk-js/scripts/codegen.js
```diff
@@ -0,0 +1,370 @@
...
+    const methodName = `${
+      operationNameWithoutNamespace.charAt(0).toLowerCase() +
+      operationNameWithoutNamespace.slice(1)
+    }`;
+
+    const methodType = methodTypeFromMethodName(methodName);
+    const signedRequestGeneratorName = `sign${methodName}`;
+    const parameterList = operation["parameters"] ?? [];
+
+    // let responseValue = "void";
```
andrewkmin, line 75: can remove?

packages/sdk-js/scripts/codegen.js
```diff
@@ -0,0 +1,370 @@
...
+    //   responseValue = `operations["${operationId}"]["responses"]["200"]["schema"]["activity"]["result"]["${methodName}Result"]`;
+    // } else if (["noop", "query"].includes(methodType)) {
+    //   responseValue = `operations["${operationId}"]["responses"]["200"]["schema"]`;
+    // } else if (methodType === "activityDecision") {
+    //   responseValue = `operations["${operationId}"]["responses"]["200"]["schema"]["activity"]["result"]`;
+    // }
+
+    let responseValue = "void";
+    if (methodType === "command") {
+      responseValue = `operations["${operationId}"]["responses"]["200"]["schema"]`;
```
andrewkmin, line 86: these all seem to be equivalent. do we mean to keep the commented out version above ^?

packages/sdk-js/scripts/codegen.js
```diff
@@ -0,0 +1,370 @@
...
+        .map((binding) => `type ${binding.name} = ${binding.value};`)
+    );
+
+    // exported type aliases
+    codeBuffer.push(
+      ...[responseTypeBinding, inputTypeBinding, bodyTypeBinding]
+        .filter((binding) => binding.isBound)
+        .map((binding) => `export type ${binding.name} = ${binding.value};`)
+    );
+
```
andrewkmin, line 156: tiny nit: extra newline

packages/sdk-js/scripts/codegen.js
```diff
@@ -0,0 +1,370 @@
...
+          parameters: {...input},
+          organizationId: currentSubOrganization?.organizationId ?? this.config.organizationId,
+          timestampMs: String(Date.now()),
+          type: "ACTIVITY_TYPE_${operationNameWithoutNamespace.replace(/([a-z])([A-Z])/g, '$1_$2').toUpperCase()}"
+        }, ...overrideParams
+      });
+  }`
+      );
+    }
+
```
andrewkmin, line 345: extra new line here as well. can probably use a lint/format

packages/sdk-js/scripts/codegen.js
```diff
@@ -0,0 +1,370 @@
...
+  }
+
+  // End of the TurnkeySDKClient Class Definition
+  codeBuffer.push(`}`);
```
andrewkmin, line 349: might want to add a newline to the very end as well

packages/sdk-js/src/sdk-client.ts
```diff
@@ -0,0 +1,207 @@
...
+  // API
+  createUserAccount = async (email: string): Promise<SdkApiTypes.TCreateSubOrganizationResponse> => {
+    const challenge = generateRandomBuffer();
+    const authenticatorUserId = generateRandomBuffer();
+
+    const attestation = await getWebAuthnAttestation({
+      publicKey: {
+        rp: {
+          id: "localhost",
+          name: "Demo Passkey Wallet"
```
andrewkmin, line 89: we can probably update this to use a generic name, like `Default App Name` or something

andrewkmin (commented):
looking great overall. i think we should be ready to ship this initial version out soon!

packages/sdk-js-server/src/sdk-client.ts
```diff
@@ -0,0 +1,142 @@
...
+  "getUsers",
+  "getWalletAccounts",
+  "getWallets",
+  "getWhoami",
+  "createSubOrganization",
+  "emailAuth",
+  "initUserEmailRecovery",
+];
+
+export class TurnkeyServerSDK {
```
andrewkmin, line 44: Tiny nit: should we rename this? Feel like we could use a single naming convention (`TurnkeyServerSDK...` vs. `TurnkeySDKServer...`)

andrewkmin (approved):

PR #236, opened 2024-04-09
export DEFAULT_ETHEREUM_ACCOUNT helpers from sdk packages

AlwaysBCoding:
Adds default helpers for generating a walletAccount at index and exports the initial configuration as a constant

### DIFF
--- a/packages/sdk-browser/src/index.ts
+++ b/packages/sdk-browser/src/index.ts
@@ -32,6 +32,11 @@ import {
   TurnkeySDKBrowserClient
 } from "./sdk-client";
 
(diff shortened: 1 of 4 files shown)


andrewkmin (approved):
Ah I did something similar in this commit: https://github.com/tkhq/sdk/commit/5ef21b8bf8a5dcd7ed50fa6d9e3d611fee6769d0

But I prefer this implementation so let's go with it! Feel free to remove what I had added

moe-dev (approved):
needs rebase but otherwise lgtm

PR #240, opened 2024-04-11
Apply settings via iframe-stamper

oliviathet:
## Summary & Motivation
ENG-1388. Adds `applySettings` method to the iframe stamper that allows custom settings for surface-level/UI components in the iframe (styles, supported mnemonic lengths, supported private key formats, extra validation, etc.). The first setting we're implementing is `styles`, where CSS styles can be validated, sanitized, and applied to a specific HTML element on import.turnkey.com.

Relevant to https://github.com/tkhq/frames/pull/34

## How I Tested These Changes
using wallet-import-export example against the old and new local instances of import.turnkey.com

## Did you add a changeset?
yes to `iframe-stamper` package

### DIFF
--- a/packages/iframe-stamper/src/index.ts
+++ b/packages/iframe-stamper/src/index.ts
@@ -29,12 +29,18 @@ export enum IframeEventType {
   // Value: none
   // Key Format (optional): the key format to decode the private key in before it's encrypted for import: HEXADECIMAL or SOLANA. Defaults to HEXADECIMAL.
   ExtractKeyEncryptedBundle = "EXTRACT_KEY_ENCRYPTED_BUNDLE",
+  // Event sent by the parent to apply settings on the iframe.
+  // Value: the settings to apply in JSON string format.
+  ApplySettings = "APPLY_SETTINGS",
   // Event sent by the iframe to its parent when `InjectBundle` is successful
   // Value: true (boolean)
   BundleInjected = "BUNDLE_INJECTED",
   // Event sent by the iframe to its parent when `ExtractEncryptedBundle` is successful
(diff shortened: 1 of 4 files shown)


packages/iframe-stamper/src/index.ts
```diff
@@ -66,6 +72,34 @@ export type TIframeStamperConfig = {
...
+  boxShadow?: string;
+  textAlign?: string;
+  overflowWrap?: string;
+  wordWrap?: string;
+  resize?: string;
+};
+
+export type TIframeSettings = {
+  styles?: TIframeStyles;
+};
```
r-n-o, line 101: Interesting. I'm guessing you took this whitelist approach for security reasons? (as opposed to allowing arbitrary styles). If so, might be worth a comment here. I think what you have here covers a lot of use cases btw, good list!

r-n-o (approved):
Build is failing but I don't think it's related to your change? Once you have a green PR feel free to merge/publish this!

packages/iframe-stamper/src/index.ts
oliviathet: so far we've gotten requests for changing margin, padding, borders, and fonts which are mostly covered here. as we get more requests, we can securely add them as a strict regex pattern on frames (for security reasons). and the whitelist here in the sdk is actually more for ux than for security -- ideally a user doesn't need to constantly cross-reference the code in frames to determine which css style properties are supported

oliviathet:
> Build is failing but I don't think it's related to your change? Once you have a green PR feel free to merge/publish this!

Internet says it could be failing because of rate limiting... not sure.

PR #242, opened 2024-04-11
Jordan/updated types

AlwaysBCoding:
This does 3 things:

1. change the naming of `rootOrganizationId` in the SDK config data to `defaultOrganizationId`

2.  merge the `activityId` onto the activity return data for any activity creation (which currently is just returning the nested data at the top-level)

3. Remove the ability to override the activity type of an activity with the parameter `type` when you call the function to create an activity


packages/sdk-js-browser/scripts/codegen.js
```diff
@@ -262,7 +262,13 @@ export class TurnkeySDKClientBase {
     let activityStatus = initialData["activity"]["status"];
```
andrewkmin, line 262: tiny nit: for consistency should this also be `const`? non blocking

andrewkmin (approved):

PR #255, opened 2024-04-17
`@turnkey/iframe-stamper`: add `sandbox` attribute to iframe element

r-n-o:
## Summary & Motivation
This branch adds an `sandbox` attribute to the `<iframe>` element inserted by our iframe stamper. This is a conservative measure taken for security reasons: we do not need most privileges (popups, form submission, top-level page navigation, and so on). Defense in depth!

## How I Tested These Changes
Need to test this locally. Will do so.

## Did you add a changeset?
Yes!


r-n-o:
It turns out `allow-same-origin` is also needed! Great answer explaining why at https://stackoverflow.com/questions/41067069/javascript-errors-localstorage-cookie-loading-sandboxed-iframe-within-chrome

AlwaysBCoding (approved):
awesome!

one note, it looks like the iframe-stamper version just had a major version bump to `2.x.x` -- are there any breaking changes in that where this attribute change might get blocked by supporting that? Feel free to ignore and merge if it's not an issue.

PR #256, opened 2024-04-17
Wrap Iframe Client to prevent leaking stamper encapsulation

AlwaysBCoding:
In-tandem with demo-embedded-wallet PR: https://github.com/tkhq/demo-embedded-wallet/pull/8

This PR wraps the iframe client to have `iframePublicKey` and `injectCredentialBundle` available at the top-level to call. This prevents a developer implementing email auth from having to reach into the stamper to call these functions as can be seen in the demo-embedded-wallet PR


packages/sdk-browser/src/sdk-client.ts
```diff
@@ -346,3 +347,18 @@ export class TurnkeySDKBrowserClient extends TurnkeySDKClientBase {
...
+export class TurnkeySDKIframeClient extends TurnkeySDKBrowserClient {
+  iframePublicKey: string | null;
+
+  constructor(config: TurnkeySDKClientConfig) {
+    super(config);
+    this.iframePublicKey = (config.stamper as IframeStamper).iframePublicKey;
+  }
+
+  injectCredentialBundle = async (credentialBundle: string): Promise<boolean> => {
+    const stamper = this.config.stamper as IframeStamper;
```
andrewkmin, line 360: I think we may want to save the `credentialBundle` in local storage and keep it there to easily access. Otherwise, you'll have to reinject the credential bundle each time the iframe is rendered (since the decrypted credential is stored in memory)

AlwaysBCoding: yeah that could definitely be a good idea. how exactly should that work? does the credential bundle ever expire? does it need to be refreshed if the iframe public key changes? would definitely be a good flow to be able to login with email and sign transactions without needing to reinject

andrewkmin: Yep, they do expire. Here's how we fetch it for example in the existing demo passkey wallet: https://github.com/tkhq/demo-passkey-wallet/blob/2182c36583fbda79f762ce5d0c70db4926feb547/frontend/components/EmailAuth.tsx#L37

And for setting it https://github.com/tkhq/demo-passkey-wallet/blob/2182c36583fbda79f762ce5d0c70db4926feb547/frontend/app/email-auth/page.tsx#L112-L117

andrewkmin: To be clear: the credential bundle does expire, based on the `expiratoinSeconds` specified when someone calls `emailAuth`. The default is 15 minutes. 

And yes, this bundle is encrypted to a specific iframe public key; if the iframe public key changes, then the bundle is no longer valid (in fact the `injectCredentialBundle` call will fail)

andrewkmin: The tricky part might be ensuring `injectCredentialBundle` always happens on each activity call

andrewkmin (approved):

PR #260, opened 2024-04-20
Abstract decoding/encoding logic from frames into a separate js package [WIP]

moe-dev:
## Summary & Motivation

@andrewkmin Drafting this here so we have something to comb through - in theory these should work, logic is mostly 1:1 with the frames auth code 

TODO: 
- Email Auth Example
- SQUASH commits

Discussion Points:
- Naming 
- Structure of Package
- If this should live in a separate package or if we can combine this with our internal package and publish

## Did you add a changeset?


packages/utils/src/index.ts
```diff
@@ -0,0 +1,299 @@
+import * as hpke from '@hpke/core';
+import { subtle, webcrypto} from 'crypto';
+
+
+// Exported Utilities
```
andrewkmin, line 5: i think we can split this file into a couple different ones to separate concerns. maybe based on: 

- key material (generating a target key, importing a credential, compressing/uncompressing, decrypt)
- math
- encoding/decoding

andrewkmin (commented):
looking good so far!

packages/utils/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# @turnkey/utils
```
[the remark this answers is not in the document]
moe-dev: Open to any ideas fs! I think this could be the foundation for more general utils that we'd want to expose to users, but if we want to differentiate igi. 

Some other ideas here:

/crypto-utils
/crypto(/graphy)

packages/utils/src/index.ts
moe-dev: Good call - Totaly agreed!

packages/frame-utils/src/encoding.ts
andrewkmin, line 1: Some repeats across this and https://github.com/tkhq/sdk/blob/main/packages/encoding/src/__tests__/index-test.ts

If we want this to be our canonical utils library, let's deprecate `encoding`

moe-dev: Agreed! Going to make sure everything is working as intended via implementing an auth feature in our react native app and we'll be good to deprecate

PR #261, opened 2024-04-22
changes top-level naming of sdk exports to clients

AlwaysBCoding:
Can see the end-result from the PR to `demo-embedded-wallet` https://github.com/tkhq/demo-embedded-wallet/pull/10

### DIFF
--- a/packages/sdk-react/src/contexts/TurnkeyContext.tsx
+++ b/packages/sdk-react/src/contexts/TurnkeyContext.tsx
@@ -1,21 +1,21 @@
 import { ReactNode, createContext, useState, useEffect, useRef } from "react";
 import {
-  TurnkeyBrowserSDK,
-  TurnkeySDKBrowserClient,
-  TurnkeySDKIframeClient,
+  Turnkey,
+  TurnkeyIframeClient,
+  TurnkeyPasskeyClient,
   TurnkeySDKBrowserConfig,
 } from "@turnkey/sdk-browser";
 
 export interface TurnkeyClientType {
-  turnkeyClient: TurnkeyBrowserSDK | undefined;
-  iframeSigner: TurnkeySDKIframeClient | undefined;
-  passkeySigner: TurnkeySDKBrowserClient | undefined;
+  turnkey: Turnkey | undefined;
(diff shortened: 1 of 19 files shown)


packages/sdk-react/src/contexts/TurnkeyContext.tsx
```diff
@@ -27,14 +27,12 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
   config,
   children,
 }) => {
-  const [turnkeyClient, setTurnkeyClient] = useState<
-    TurnkeyBrowserSDK | undefined
+  const [turnkey, setTurnkey] = useState<Turnkey | undefined>(undefined);
```
andrewkmin, line 30: minor nit: i'm down to initialize these as empty `const [turnkey, setTurnkey] = useState<Turnkey | undefined>();`

packages/sdk-server/src/sdk-client.ts
```diff
@@ -115,10 +116,16 @@ export class TurnkeyServerSDK {
   };
 }
 
-export class TurnkeySDKServerClient extends TurnkeySDKClientBase {
+export class TurnkeyServerClient extends TurnkeySDKClientBase {
```
andrewkmin, line 119: is there anything that explicitly uses `TurnkeyServerClient` now that we also have `TurnkeyApiClient`?

andrewkmin (approved):
tiny nits but lgtm!

packages/sdk-server/src/sdk-client.ts
AlwaysBCoding: no, but this is just modeling the same format as the browser sdk with the intermediary BrowserClient and ServerClient for the individual clients to inherit from

packages/sdk-react/src/contexts/TurnkeyContext.tsx
AlwaysBCoding: what's the difference? I honestly don't know. feels like it would be better to be explicit about the initial value though?

andrewkmin: no difference! down for either. let's ship it 🚢

PR #265, opened 2024-04-25
update with-federated-passkeys example

andrewkmin:
## Did you add a changeset?


packages/sdk-browser/src/sdk-client.ts
```diff
@@ -212,8 +222,8 @@ export class TurnkeyPasskeyClient extends TurnkeyBrowserClient {
         ],
         user: {
           id: config.publicKey?.user?.id ?? authenticatorUserId,
-          name: config.publicKey?.user?.name ?? "",
-          displayName: config.publicKey?.user?.displayName ?? "",
+          name: config.publicKey?.user?.name ?? "Default User",
```
andrewkmin, line 225: Should we rename the `config` variable to `input`? Super minor. But when I see config my first assumption is that it's related to the overall `turnkey.json` config.

Second, is there a particular reason all these fields have to be nested under `publicKey`?

packages/sdk-browser/src/storage.ts
```diff
@@ -40,7 +41,6 @@ export const setStorageValue = async <K extends StorageKeys>(
   const storageLocation: StorageLocation = STORAGE_VALUE_LOCATIONS[storageKey];
   const browserStorageLocation: Storage = STORAGE_LOCATIONS[storageLocation];
   browserStorageLocation.setItem(storageKey, JSON.stringify(storageValue));
-  // eventEmitter.emit('storageChange', { storageKey, storageValue });
```
andrewkmin, line 43: Assuming we don't need this at all at this point

examples/with-federated-passkeys/src/pages/index.tsx
```diff
@@ -117,45 +69,12 @@ export default function Home() {
   };
 
   const createSubOrg = async (data: subOrgFormData) => {
-    const challenge = generateRandomBuffer();
```
andrewkmin, line 120: So hyped about burning this all down

packages/sdk-browser/src/storage.ts
AlwaysBCoding: yeah can be deleted, this is for a flow where the sdk can natively watch for changes to storage values and update the react context but we're not currently doing that so it's ok to delete

packages/sdk-browser/src/sdk-client.ts
AlwaysBCoding: I'd be ok with changing it to `input` -- agree that's more intuitive.

The reason it's nested under `publicKey` is that the input is meant to conform to the native browser `CredentialCreationOptions` type which requires the creation options to be nested underneath `publicKey`. If we know that this can only take configuration related to a `publicKey` we could change the input to conform to the native `PublicKeyCredentialCreationOptions` instead. I think that might be a good idea in this case, but the overall idea is that you could just copy and paste a full native browser Credential configuration into the `createUserPasskey` function and it should work accordingly, as well as a single override parameter if you wanted the defaults for everything else

andrewkmin: That makes sense. Down to leave it as is for now. The reason this came about is that I was fine with all params, but just wanted to customize the name (since the default is ""), but might be too early to tell whether individual params should be overridden, or if they should be holistic as you mention

PR #278, opened 2024-05-06
Crypto package for rn encoding/decoding/decryption

moe-dev:
## Summary & Motivation
This package will serve as the primitives associated with decryption, encryption and keygen. The goal here was originally to make email auth available in a much simpler way for rn clients, but this actually is just a purjs implementation for any runtime for (primarily) hpke decryption. Currently there are some hardcoded values that are tk specific but a fast-follow is generalizing this for anyone to use.

## Did you add a changeset?


packages/crypto/src/constants.ts
```diff
@@ -0,0 +1,30 @@
...
+]);
+
+export const IV_INFO = new Uint8Array([
+  0, 12, 72, 80, 75, 69, 45, 118, 49, 72, 80, 75, 69, 0, 16, 0, 1, 0, 2, 98, 97,
+  115, 101, 95, 110, 111, 110, 99, 101, 0, 143, 195, 174, 184, 50, 73, 10, 75,
+  90, 179, 228, 32, 35, 40, 125, 178, 154, 31, 75, 199, 194, 34, 192, 223, 34,
+  135, 39, 183, 10, 64, 33, 18, 47, 63, 4, 233, 32, 108, 209, 36, 19, 80, 53,
+  41, 180, 122, 198, 166, 48, 185, 46, 196, 207, 125, 35, 69, 8, 208, 175, 151,
+  113, 201, 158, 80,
+]);
```
r-n-o, line 30: it's be nice to have these constants in text / utf8 form and convert them to bytes as needed. For instance:
```js
const LABEL_SECRET = toUint8Array('secret')
const ...etc...
```

Also it'd be good to add comments for what these values are / why they can be hardcoded.

packages/crypto/package.json
```diff
@@ -0,0 +1,53 @@
...
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.js",
+      "default": "./dist/index.mjs"
+    }
+  },
+  "types": "./dist/index.d.ts",
+  "license": "Apache-2.0",
+  "description": "Encoding/Decoding utility functions",
```
r-n-o, line 16: This is `@turnkey/crypto`, let's update the description accordingly! :)

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+  publicKeyUncompressed?: string;
+}
+
+/**
+ * Get PublicKey function
+ * Derives public key from Uint8Array private key
+ *
+ * @param {Uint8Array} privateKey - The Uint8Array representation of a private key
+ * @param {boolean} isCompressed - true by default, specifies whether to return a compressed or uncompressed public key
+ * @returns {Promise<Uint8Array>} - The public key in Uin8Array representation.
```
r-n-o, line 37: this function isn't async anymore! 🎉

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+
+export const getPublicKey = (
+  privateKey: Uint8Array,
+  isCompressed: boolean = true
+): Uint8Array => {
+  return p256.getPublicKey(privateKey, isCompressed);
+};
+/**
+ * HPKE Decrypt Function
+ * Decrypts data using Hybrid Public Key Encryption (HPKE) approach.
```
r-n-o, line 48: this is more than an approach, it's a standard! Let's link https://datatracker.ietf.org/doc/rfc9180/

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+    let info: Uint8Array;
+    const receiverPubBuf = getPublicKey(base64urlDecode(receiverPriv), false);
+    const aad = additionalAssociatedData(encappedKeyBuf, receiverPubBuf);
+    const kemContext = getKemContext(
+      encappedKeyBuf,
+      uint8arrayToHexString(receiverPubBuf)
+    );
+
+    // Step 1: Generate Shared Secret
+    const dh = await deriveDh(encappedKeyBuf, receiverPriv);
```
r-n-o, line 69: let's name this "ss" for "shared secret"?

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+  const publicKeyArray = uint8arrayFromHexString(publicKey);
+
+  const kemContext = new Uint8Array(
+    encappedKeyArray.length + publicKeyArray.length
+  );
+  kemContext.set(encappedKeyArray);
+  kemContext.set(publicKeyArray, encappedKeyArray.length);
+
+  return kemContext;
+};
```
r-n-o, line 498: At the end of the day, this function is doing a simple concatenation:
```js
new Uint8Array([...encappedKeyBuf, ...uint8arrayFromHexString(publicKey)])
```

Let's inline it given it's only used once?

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+ */
+const aesGcmDecrypt = (
+  encryptedData: Uint8Array,
+  key: Uint8Array,
+  iv: Uint8Array,
+  aad?: Uint8Array
+): Uint8Array => {
+  const aes = gcm(key, iv, aad);
+  const data_ = aes.decrypt(encryptedData);
+  return data_;
```
r-n-o, line 478: drop the underscore (unless there's a good reason for it). Or even better, get rid of this variable entirely: `return aes.decrypt(encryptedData)`

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+): Promise<Uint8Array> => {
+  const prk = hkdf.extract(sha256, ikm, sharedSecret);
+  const resp = hkdf.expand(sha256, prk, info, len);
+  return new Uint8Array(resp);
+};
+
+/**
+ * Derive the Diffie-Hellman (DH) shared secret using ECDH.
+ */
+const deriveDh = async (
```
r-n-o, line 456: let's call this `deriveSharedSecret` for consistency with how this secret is referenced in callers

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+  encappedKeyBuf: Uint8Array,
+  receiverPriv: string
+): Promise<Uint8Array> => {
+  const dh = p256.getSharedSecret(
+    base64urlDecode(receiverPriv),
+    encappedKeyBuf
+  );
+  return dh.slice(1);
```
r-n-o, line 464: worth a comment on top of this line to explain why `slice` is needed.

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+
+/**
+ * Perform HKDF extract and expand operations.
+ */
+const extractAndExpand = async (
+  sharedSecret: Uint8Array,
+  ikm: Uint8Array,
+  info: Uint8Array,
+  len: number
+): Promise<Uint8Array> => {
```
r-n-o, line 447: there's no need for this function to be async / return a `Promise`

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+  return new Uint8Array(resp);
+};
+
+/**
+ * Derive the Diffie-Hellman (DH) shared secret using ECDH.
+ */
+const deriveDh = async (
+  encappedKeyBuf: Uint8Array,
+  receiverPriv: string
+): Promise<Uint8Array> => {
```
r-n-o, line 459: let's drop the async / promise interface since everything within this is synchronous (I don't see any `await`)

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+  return array;
+};
+
+/**
+ * Create additional associated data (AAD) for AES-GCM decryption.
+ */
+const additionalAssociatedData = (
+  senderPubBuf: ArrayBuffer,
+  receiverPubBuf: ArrayBuffer
+): Uint8Array => {
```
r-n-o, line 433: Interesting. Why are you using `ArrayBuffer` here (instead of `Uint8Array` like everywhere else?

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+/**
+ * Generate a random Uint8Array of a specific length.
+ */
+const randomBytes = (length: number): Uint8Array => {
+  const array = new Uint8Array(length);
+  for (let i = 0; i < length; i++) {
+    array[i] = Math.floor(Math.random() * 256);
+  }
+  return array;
+};
```
r-n-o, line 425: This isn't secure because these random bytes are from a PRNG, not a CSPRNG. I think we need to find a way to use  the system's CSPRNG using e.g. https://github.com/LinusU/react-native-get-random-values

Because we're using `randomBytes` to create a private key it's not okay to rely on a PRNG unfortunately :(

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+    );
+  }
+  return hexString.padStart(length, "0");
+};
+
+/**
+ * Uncompress a raw public key.
+ */
+
+const uncompressRawPublicKey = (rawPublicKey: Uint8Array): Uint8Array => {
```
r-n-o, line 333: This should probably not be part of our public functions, or we if we want to expose it we need to make sure it reflects the fact that it's P256-specific.

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+  label: Uint8Array,
+  info: Uint8Array,
+  suite_id: Uint8Array,
+  len: number
+): Uint8Array => {
+  const ret = new Uint8Array(
+    9 + suite_id.byteLength + label.byteLength + info.byteLength
+  );
+  ret.set(new Uint8Array([0, len]), 0);
+  ret.set(HPKE_VERSION, 2);
```
r-n-o, line 309: I'm a bit confused: why are we leaving the first two bytes at zero?

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+ * @returns {Promise<Uint8Array | null>} - The decrypted data or null if decryption fails.
+ */
+export const decryptBundle = async (
+  credentialBundle: string,
+  embeddedKey: string
+): Promise<Uint8Array | null> => {
+  try {
+    let bundleBytes: Uint8Array;
+
+    // Decode using either Base58Check or Base64URL
```
r-n-o, line 132: I don't think this is needed, you can drop the base64url branch. All bundles are base58check encoded these days!

r-n-o (commented):
OMG it's happening!! 🎉

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+  }
+  return bytes;
+};
+
+/**
+ * Generate a target key pair using P-256.
+ *
+ * @returns {Promise<KeyPair>} - The generated key pair.
+ */
+export const generateTargetKey = async (): Promise<KeyPair> => {
```
andrewkmin, line 208: Tiny nit: let's add a comment either to this method or `generateP256KeyPair` that will further differentiate the two. We can also consider changing the return types of the two methods based on difference in use cases

packages/crypto/README.md
```diff
@@ -0,0 +1,3 @@
+# @turnkey/crypto
+
+This package consolidates some common cryptographic utilities used across our applications, particularly primitives related to keys, encoding/decoding and encryption.
```
andrewkmin, line 3: `keys, encryption, and decryption` (assuming we relegate all encoding/decoding stuff to `tkhq/encoding`)

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+ * @param {Uint8Array} label - The label to use.
+ * @param {Uint8Array} info - Additional information.
+ * @param {Uint8Array} suite_id - The suite identifier.
+ * @param {number} len - The output length.
+ * @returns {Uint8Array} - The labeled info.
+ */
+const buildLabeledInfo = (
+  label: Uint8Array,
+  info: Uint8Array,
+  suite_id: Uint8Array,
```
andrewkmin, line 302: let's camel case `suite_id`?

packages/crypto/src/crypto.ts
moe-dev: This was a way to get around this ts error:

`Type 'Uint8Array' can only be iterated through when using the '--downlevelIteration' flag or with a '--target' of 'es2015' or higher.`

But I believe I can keep typing more strict and just use Array.from

packages/crypto/src/crypto.ts
moe-dev: Good call!

packages/crypto/src/constants.ts
```diff
@@ -0,0 +1,30 @@
+export const SUITE_ID_1 = new Uint8Array([75, 69, 77, 0, 16]);
+export const SUITE_ID_2 = new Uint8Array([72, 80, 75, 69, 0, 16, 0, 1, 0, 2]);
+export const HPKE_VERSION = new Uint8Array([72, 80, 75, 69, 45, 118, 49]);
+// b"secret"
+export const LABEL_SECRET = new Uint8Array([115, 101, 99, 114, 101, 116]);
+// b"HPKE"
```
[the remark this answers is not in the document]
moe-dev: no need to leave them in - will probably just uint8array the original string for clarity here

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+    const decryptedData = await hpkeDecrypt({
+      ciphertextBuf,
+      encappedKeyBuf,
+      receiverPriv: embeddedKey,
+    });
+
+    return decryptedData;
+  } catch (error) {
+    console.error("Error injecting bundle:", error);
+    return null;
```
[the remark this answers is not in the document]
moe-dev: nope, fine by me!

packages/crypto/src/crypto.ts
moe-dev: Good call!

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,498 @@
...
+
+    // Step 3: Get IV
+    info = IV_INFO;
+    const iv = await extractAndExpand(sharedSecret, ikm, info, 12);
+
+    // Step 4: Decrypt
+    const decryptedData = await aesGcmDecrypt(ciphertextBuf, key, iv, aad);
+    return decryptedData;
+  } catch (error) {
+    console.error("Decryption Error:", error);
```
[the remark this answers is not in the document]
andrewkmin: planning on tackling this? or feel like we don't need to?

packages/crypto/README.md
```diff
@@ -0,0 +1,7 @@
+# @turnkey/crypto
+
+This package consolidates some common cryptographic utilities used across our applications, particularly primitives related to keys, encryption, and decryption. For react-native you will need to polyfill our random byte generation by importing:
+
+```
+import 'react-native-get-random-values'
```
andrewkmin, line 6: could be a good idea to link to https://www.npmjs.com/package/react-native-get-random-values outright

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,405 @@
...
+
+/**
+ * Get PublicKey function
+ * Derives public key from Uint8Array or hexstring private key
+ *
+ * @param {Uint8Array | string} privateKey - The Uint8Array or hexstring representation of a compressed private key
+ * @param {boolean} isCompressed - true by default, specifies whether to return a compressed or uncompressed public key
+ * @returns {<Uint8Array>} - The public key in Uin8Array representation.
+ */
+
```
andrewkmin, line 39: let's remove this empty line for consistency

packages/crypto/src/crypto.ts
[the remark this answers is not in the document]
moe-dev: Sorry - totally missed this, added in my next commit!

packages/crypto/src/crypto.ts
moe-dev: We really need a whitespace linter 😅

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,389 @@
...
+    throw new Error("y is out of range");
+  }
+
+  var uncompressedHexString = "04" + bigIntToHex(x, 64) + bigIntToHex(y, 64);
+  return uint8ArrayFromHexString(uncompressedHexString);
+};
+/**
+ * Compute the modular square root using the Tonelli-Shanks algorithm.
+ */
+
```
andrewkmin, line 261: tiny nit: let's add a newline after line 257, and remove this extra newline here

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,389 @@
...
+      `number cannot fit in a hex string of ${length} characters`
+    );
+  }
+  return hexString.padStart(length, "0");
+};
+
+/**
+ * Uncompress a raw public key.
+ */
+
```
andrewkmin, line 224: can remove this extra newline

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,389 @@
...
+ * @returns {Uint8Array} - The labeled info.
+ */
+const buildLabeledInfo = (
+  label: Uint8Array,
+  info: Uint8Array,
+  suite_id: Uint8Array,
+  len: number
+): Uint8Array => {
+  const ret = new Uint8Array(
+    9 + suite_id.byteLength + label.byteLength + info.byteLength
```
andrewkmin, line 198: should we denote `9` as a const/offset given how we use it multiple times here?

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,389 @@
...
+ * Decrypts data using Hybrid Public Key Encryption (HPKE) standard https://datatracker.ietf.org/doc/rfc9180/.
+ *
+ * @param {HPKEDecyptParams} params - The decryption parameters including ciphertext, encapsulated key, and receiver private key.
+ * @returns {Uint8Array} - The decrypted data.
+ */
+export const hpkeDecrypt = ({
+  ciphertextBuf,
+  encappedKeyBuf,
+  receiverPriv,
+}: HPKEDecyptParams): Uint8Array => {
```
andrewkmin, line 60: tiny typo/nit: `HPKEDecryptParams`

packages/crypto/src/crypto.ts
```diff
@@ -0,0 +1,389 @@
...
+
+interface HPKEDecyptParams {
+  ciphertextBuf: Uint8Array;
+  encappedKeyBuf: Uint8Array;
+  receiverPriv: string;
+}
+
+interface KeyPair {
+  privateKey: string;
+  publicKey: Uint8Array | string;
```
andrewkmin, line 30: now that we only use `KeyPair` in one place, we can probably be more specific with the expected type

packages/crypto/src/__tests__/crypto-test.ts
```diff
@@ -0,0 +1,21 @@
+import { test, expect, describe } from "@jest/globals";
+import { uint8ArrayFromHexString } from "@turnkey/encoding";
+import { getPublicKey, generateP256KeyPair } from "../crypto";
+
+describe("Crypto Utils - Exported Functions", () => {
```
andrewkmin, line 5: would be nice to test other functions that we export as well (a large e2e-like test could do the trick)

andrewkmin (commented):
🔥 just some minor comments. looking ready to ship soon!

packages/crypto/src/crypto.ts
moe-dev: good call - copy pasta from an example haha

packages/crypto/src/crypto.ts
```diff
@@ -184,24 +184,25 @@ const buildLabeledIkm = (
...
  * @returns {Uint8Array} - The labeled info.
  */
 const buildLabeledInfo = (
   label: Uint8Array,
   info: Uint8Array,
-  suite_id: Uint8Array,
+  suiteId: Uint8Array,
   len: number
 ): Uint8Array => {
+  const suiteIdStartIndex = 9; // first two are reserved for length bytes (unused in this case), the next 7 are for the HPKE_VERSION, then the suiteId starts at 9
```
andrewkmin, line 197: very helpful context, thanks for adding

andrewkmin (approved):
lgtm 🚀🚀🚀

PR #282, opened 2024-05-13
enable passkey sessions

andrewkmin:
## Summary & Motivation
See iteration cycles on https://replit.com/@andrew306/RepulsiveRundownObject#index.ts

This flavor of creating passkey sessions assumes that you've created the suborg already and know their suborg ID.

This does not encapsulate logic to, say, create a new suborg and have a fresh passkey session for them out of the gate. However, the abstraction is there: you could call `createEmbeddedAPIKey`, grab the resulting public key and include it as a user's API key within a `CreateSubOrganization` call.

Note that this depends on `hpke-js` which is not available in some environments (such as react native). We'll have to do some followup work there to get native support at a later time.

## Did you add a changeset?

### DIFF
--- a/packages/sdk-browser/src/sdk-client.ts
+++ b/packages/sdk-browser/src/sdk-client.ts
@@ -25,7 +25,7 @@ import { TurnkeyRequestError } from "./__types__/base";
 import { TurnkeySDKClientBase } from "./__generated__/sdk-client-base";
 import type * as SdkApiTypes from "./__generated__/sdk_api_types";
 
-import type { User, SubOrganization } from "./models";
+import type { User, SubOrganization, EmbeddedAPIKey } from "./models";
 import {
   StorageKeys,
   getStorageValue,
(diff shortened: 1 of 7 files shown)


packages/sdk-browser/src/sdk-client.ts
```diff
@@ -246,8 +252,145 @@ export class TurnkeyPasskeyClient extends TurnkeyBrowserClient {
...
+  senderPubBuf: Uint8Array,
+  receiverPubBuf: Uint8Array,
+): Uint8Array => {
+  return new Uint8Array([
+    ...Array.from(senderPubBuf),
+    ...Array.from(receiverPubBuf),
+  ]);
+};
+
+export const compressPubKey = (publicKey: Uint8Array): Uint8Array => {
```
moe-dev, line 373: let's move this to @turnkey/crypto?

packages/sdk-browser/src/sdk-client.ts
```diff
@@ -246,8 +252,145 @@ export class TurnkeyPasskeyClient extends TurnkeyBrowserClient {
...
+
+    return base58encodedBundle;
+  };
 }
 
+// NOTE: the below can be moved to utility packages
+/**
+ * Create additional associated data (AAD) for AES-GCM decryption.
+ */
+const additionalAssociatedData = (
```
moe-dev, line 363: expose this from @turnkey/crypto

packages/sdk-browser/src/sdk-client.ts
```diff
@@ -246,8 +252,145 @@ export class TurnkeyPasskeyClient extends TurnkeyBrowserClient {
       attestation: attestation,
     };
   };
+
+  createPasskeySession = async (
```
moe-dev, line 256: Important to note that this won't be compatible with rn unless https://www.npmjs.com/package/react-native-quick-crypto gets to a place where all this polyfill is supported. Potentially out of this scope, but as some point we prob want to put all of this encoding/decoding logic in @turnkey/crypto

packages/sdk-browser/src/sdk-client.ts
```diff
@@ -246,8 +252,145 @@ export class TurnkeyPasskeyClient extends TurnkeyBrowserClient {
...
+  ) => {
+    const TURNKEY_HPKE_INFO = new TextEncoder().encode("turnkey_hpke");
+
+    // If local storage contains the encrypted key, then use it. OR just create a new session each time.
+
+    const localStorageUser = await getStorageValue(StorageKeys.CurrentUser);
+    userId = userId ?? localStorageUser?.userId;
+
+     // Step 1: create api key (to later encrypt)
+     const p256key = await crypto.subtle.generateKey(
```
moe-dev, line 269: can use [generateP256Key](https://github.com/tkhq/sdk/blob/6cb9035867e19f3dacbe55ebb9fe62d69e3017ce/packages/crypto/src/crypto.ts#L137)

moe-dev (commented):
LG so far! Just calling out some overlap with our new /crypto pkg but potentially out of scope here :)

packages/sdk-browser/src/sdk-client.ts
andrewkmin: yep, precisely the reason for the crypto PR! 😄

packages/sdk-browser/src/sdk-client.ts
andrewkmin: Definitely. I'm currently not expecting this to be used in the RN context. I think it's worth picking up/prioritizing later on if there's an explicit ask , at which point we'd probably need to... essentially implement hpke encrypt 👀

packages/sdk-browser/src/sdk-client.ts
```diff
@@ -183,6 +195,64 @@ export class TurnkeyBrowserClient extends TurnkeySDKClientBase {
     await setStorageValue(StorageKeys.CurrentUser, currentUser);
     return readOnlySessionResult!;
   };
+
+  // createEmbeddedAPIKey creates an embedded API key encrypted to a target key (typically embedded within an iframe).
+  // This returns a bundle that can be decrypted by that target key, as well as the public key of the newly created API key.
+  createEmbeddedAPIKey = async (
```
AlwaysBCoding, line 201: if the `createEmbeddedAPIKey` function is general and doesn't require specific knowledge of the current turnkey instance I think it would be better for it to live in `src/utils.ts` and then have users interact with it via `passkeyClient.createPasskeySession()` or is there a reason why the specific crypto logic needs to be exposed at the top-level of the `turnkey` instance?

andrewkmin: I wanted to abstract it out + still expose it because in theory, a user might want to be able to create an API key within the browser and then include it within a new suborg creation. From the freshly updated PR description:

> This does not encapsulate logic to, say, create a new suborg and have a fresh passkey session for them out of the gate. However, the abstraction is there: you could call createEmbeddedAPIKey, grab the resulting public key and include it as a user's API key within a CreateSubOrganization call.

AlwaysBCoding (approved):

packages/sdk-browser/src/utils.ts
```diff
@@ -1,4 +1,73 @@
...
+
 import { Buffer } from "buffer";
+import bs58check from "bs58check";
+import { AeadId, CipherSuite, KdfId, KemId } from "hpke-js";
+
+import type { EmbeddedAPIKey } from "./models";
+
+// createEmbeddedAPIKey creates an embedded API key encrypted to a target key (typically embedded within an iframe).
+// This returns a bundle that can be decrypted by that target key, as well as the public key of the newly created API key.
+export const createEmbeddedAPIKey = async (
```
moe-dev, line 16: Def not a light lift  but eventually would want to add this (ideally not runtime sensitive) to @turnkey/crypto.

moe-dev (approved):
LGTM!

packages/sdk-browser/src/utils.ts
andrewkmin: totally agree. would be a non trivial amount of work as we're now acutely aware of. but soon, hopefully :)

PR #283, opened 2024-05-13
add utils and rearrange @turnkey/crypto

andrewkmin:
## Summary & Motivation
- move math-related utils to `math.ts`
- rename + rearrange some functions
- add `extractPrivateKeyFromPKCS8Bytes`, specifically for use in https://github.com/tkhq/sdk/pull/282. This is required because exporting private keys using webcrypto is only supported by the `pkcs8` format; you can't export using `raw`. As a result, we have to do some manual manipulation to get the 32-byte "actual" private key out of the pkcs8 envelope

## Did you add a changeset?

### DIFF
--- a/packages/crypto/src/crypto.ts
+++ b/packages/crypto/src/crypto.ts
@@ -8,6 +8,7 @@ import {
 } from "@turnkey/encoding";
 import * as bs58check from "bs58check";
 
+import { modSqrt, testBit, randomBytes } from "./math";
 import {
   AES_KEY_INFO,
   HPKE_VERSION,
@@ -65,7 +66,7 @@ export const hpkeDecrypt = ({
       uint8ArrayFromHexString(receiverPriv),
       false
     );
-    const aad = additionalAssociatedData(encappedKeyBuf, receiverPubBuf);
+    const aad = buildAdditionalAssociatedData(encappedKeyBuf, receiverPubBuf);
     const kemContext = getKemContext(
       encappedKeyBuf,
       uint8ArrayToHexString(receiverPubBuf)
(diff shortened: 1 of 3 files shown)


packages/crypto/src/math.ts
```diff
@@ -0,0 +1,57 @@
+/**
```
moe-dev, line 1: Should we add some tests here since we're exposing these fxns? Also nit: consistent docstrings for exported fxns

packages/crypto/src/crypto.ts
```diff
@@ -148,79 +149,53 @@ export const generateP256KeyPair = (): KeyPair => {
 };
 
 /**
- * Build labeled Initial Key Material (IKM).
- *
- * @param {Uint8Array} label - The label to use.
- * @param {Uint8Array} ikm - The input key material.
- * @param {Uint8Array} suiteId - The suite identifier.
- * @returns {Uint8Array} - The labeled IKM.
+ * Create additional associated data (AAD) for AES-GCM decryption.
```
moe-dev, line 152: Nit: keep docstrings consistent for exported functions

packages/crypto/src/crypto.ts
```diff
@@ -148,79 +149,53 @@ export const generateP256KeyPair = (): KeyPair => {
...
+ * PKCS#8 private key is structured with the key data at a specific position.
+ * The actual key starts at byte 36 and is 32 bytes long
+ * @param {Uint8Array} privateKey
  */
-const buildLabeledInfo = (
-  label: Uint8Array,
-  info: Uint8Array,
-  suiteId: Uint8Array,
-  len: number
+export const extractPrivateKeyFromPKCS8Bytes = (
```
moe-dev, line 170: Can we add a test for this?

moe-dev (approved):
LGTM! Nothing blocking here, minor comments. Perhaps we should also include Encryption somewhere down the line. Also since we're here - can we revisit and remove unused dependencies - i.e react-native & base-64 libs?

packages/crypto/src/crypto.ts
andrewkmin: good call -- added!

packages/crypto/src/math.ts
andrewkmin: We're exposing these functions to use in `crypto.ts`; not quite exporting to the outside world (yet). But you're right, we should probably beef up test coverage in this library overall

moe-dev: Ahh mb but yeah haha wouldn't hurt

andrewkmin: I'm gonna punt on adding tests if that's okay — will leave that as an open issue

packages/crypto/src/math.ts
```diff
@@ -0,0 +1,57 @@
...
+};
+
+/**
+ * Generate a random Uint8Array of a specific length.
+ */
+export const randomBytes = (length: number): Uint8Array => {
+  const array = new Uint8Array(length);
+  // @ts-ignore
+  return crypto.getRandomValues(array);
+};
```
r-n-o, line 39: Not sure I agree this should be in "math.ts". It's squarely a crypto function isn't it?

packages/crypto/src/math.ts
```diff
@@ -0,0 +1,57 @@
...
+  const m = BigInt(1) << BigInt(i);
+  return (n & m) !== BigInt(0);
+};
+
+/**
+ * Generate a random Uint8Array of a specific length.
+ */
+export const randomBytes = (length: number): Uint8Array => {
+  const array = new Uint8Array(length);
+  // @ts-ignore
```
r-n-o, line 37: what is this @ts-ignore for? I'm a bit surprised this is causing typescript to complain, looks pretty straightforward 🤔

r-n-o (approved):
Overall looking solid!

packages/crypto/src/math.ts
andrewkmin: Yeah that's fair!

packages/crypto/src/math.ts
andrewkmin: I suppose it's because the context is ambiguous (node vs. web)... i.e. we could add a `/// <reference lib="dom" />` to reference web or `import crypto from crypto`

r-n-o (approved):

PR #284, opened 2024-05-13
update npm versions

andrewkmin:
## Summary & Motivation
For nextjs, address https://github.com/advisories/GHSA-77r5-gw3j-2mpf and https://github.com/advisories/GHSA-fr5h-rqp8-mj6g.

For @turnkey/crypto, pin versions

## Did you add a changeset?


moe-dev (approved):
Ty!

packages/crypto/package.json
```diff
@@ -41,19 +41,19 @@
...
-    "bs58check": "^3.0.1",
-    "react-native": "^0.74.0",
-    "react-native-get-random-values": "^1.11.0",
-    "react-native-quick-base64": "^2.1.2",
+    "@noble/ciphers": "0.5.3",
+    "@noble/curves": "1.4.0",
+    "@noble/hashes": "1.4.0",
+    "@turnkey/encoding": "workspace:*",
+    "bs58check": "3.0.1",
+    "react-native": "0.74.0",
```
moe-dev, line 49: Now that I'm looking at it - at some point it's worth removing some of these dependencies that are not needed - depcheck didn't seem to do it's job here, but I *think* quick-base64 and react-native are not needed here. Can address this in a separate PR when we do our next minor release.

PR #294, opened 2024-06-03
Moe/add hpke encryption 

moe-dev:
## Summary & Motivation
We got decryption so we may as well have encryption! Afaik, nothing really out there that supports e2e hpke decryption and encryption that can be used with any runtime, so why not :)

## Did you add a changeset?

### DIFF
--- a/packages/crypto/README.md
+++ b/packages/crypto/README.md
@@ -5,3 +5,36 @@ This package consolidates some common cryptographic utilities used across our ap
 ```
 import 'react-native-get-random-values'
 ```
+
+Example usage (Hpke):
+
+```
+const senderKeyPair = generateP256KeyPair();
+const receiverKeyPair = generateP256KeyPair();
+
+
+const receiverPublicKeyUncompressed = uncompressRawPublicKey(
+      uint8ArrayFromHexString(receiverKeyPair.publicKey)
+);
+
+const plaintext = "Hello, this is a secure message!";
+const encryptedData = hpkeEncrypt({
+    plainText: plaintext,
+    encappedKeyBuf: receiverPublicKeyUncompressed, // Assuming the use of uncompressed format
+    senderPriv: senderKeyPair.privateKey,
+    });
+
+    // Extract the encapsulated key buffer and the ciphertext
+    const encappedKeyBuf = encryptedData.slice(0, 33);
+    const ciphertextBuf = encryptedData.slice(33);
+
+    const decryptedData = hpkeDecrypt({
+      ciphertextBuf,
+      encappedKeyBuf: uncompressRawPublicKey(encappedKeyBuf), // Directly pass the buffer without additional processing
+      receiverPriv: receiverKeyPair.privateKey,
+    });
+
+    // Convert decrypted data back to string
+    const decryptedText = new TextDecoder().decode(decryptedData);
+
+```
--- a/packages/crypto/README.md
+++ b/packages/crypto/README.md
@@ -1,40 +1,38 @@
 # @turnkey/crypto
 
-This package consolidates some common cryptographic utilities used across our applications, particularly primitives related to keys, encryption, and decryption. For react-native you will need to polyfill our random byte generation by importing react-native-get-random-values: https://www.npmjs.com/package/react-native-get-random-values
+This package is runtime agnostic and consolidates some common cryptographic utilities used across our applications, particularly primitives related to keys, encryption, and decryption. For react-native you will need to polyfill our random byte generation by importing react-native-get-random-values: https://www.npmjs.com/package/react-native-get-random-values
 
 ```
 import 'react-native-get-random-values'
 ```
 
-Example usage (Hpke):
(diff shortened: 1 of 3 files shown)


packages/crypto/src/crypto.ts
```diff
@@ -97,7 +170,7 @@ export const hpkeDecrypt = ({
 };
 
 /**
- * Decrypt an encrypted credential bundle.
+ * Decrypt an encrypted email auth/recovery credential bundle.
```
andrewkmin, line 173: maybe a slight rewording: `Decrypt an encrypted credential bundle, i.e. for use with Turnkey Email Auth + Recovery.`

moe-dev: ehh, this function specfically works with email bundles, since it's a string. Our export bundles are a bit different, I should probably include a helper function for that as well

packages/crypto/README.md
```diff
@@ -1,7 +1,38 @@
 # @turnkey/crypto
 
-This package consolidates some common cryptographic utilities used across our applications, particularly primitives related to keys, encryption, and decryption. For react-native you will need to polyfill our random byte generation by importing react-native-get-random-values: https://www.npmjs.com/package/react-native-get-random-values
+This package is runtime agnostic and consolidates some common cryptographic utilities used across our applications, particularly primitives related to keys, encryption, and decryption. For react-native you will need to polyfill our random byte generation by importing react-native-get-random-values: https://www.npmjs.com/package/react-native-get-random-values
```
r-n-o, line 3: Nit: another way to say "runtime agnostic" is "pure JS". I've seen this used many times in the JS ecosystem

packages/crypto/README.md
```diff
@@ -1,7 +1,38 @@
...
 
 ```
 import 'react-native-get-random-values'
 ```
+
+Example usage (Hpke E2E):
```
r-n-o, line 9: 🔥

packages/crypto/src/crypto.ts
```diff
@@ -48,6 +54,73 @@ export const getPublicKey = (
...
+  senderPriv,
+}: HpkeEncryptParams): Uint8Array => {
+  try {
+    const senderPubBuf = getPublicKey(
+      uint8ArrayFromHexString(senderPriv),
+      false
+    );
+    const aad = buildAdditionalAssociatedData(senderPubBuf, encappedKeyBuf);
+    // Step 1: Convert the plain text string to a Uint8Array
+    const plaintextBuf = new TextEncoder().encode(plainText);
```
r-n-o, line 77: We shouldn't make the assumption that the encrypted content is utf8 text. It's not true in many cases. Let's take a uin8array as an argument here, it'll be way more flexible (otherwise there's no way to encrypt a raw random secret here since not all bytes will be valid utf8 characters)

packages/crypto/src/crypto.ts
```diff
@@ -48,6 +54,73 @@ export const getPublicKey = (
...
+  plainText,
+  encappedKeyBuf,
+  senderPriv,
+}: HpkeEncryptParams): Uint8Array => {
+  try {
+    const senderPubBuf = getPublicKey(
+      uint8ArrayFromHexString(senderPriv),
+      false
+    );
+    const aad = buildAdditionalAssociatedData(senderPubBuf, encappedKeyBuf);
```
r-n-o, line 75: A comment here here would be nice: are we making assumptions about aad structure? I think we are. It's not generic (but fine IMO). Another way to go about this would be to let callers pass their own aad if we want to maximize flexibility

moe-dev: I'll add a comment here - but for now these functions still contain hardcoded values and functionalities that are specific to Turnkey's way of encrypting and decrypting. I think in a separate version we can start including modularity like passing aad, info, etc.

packages/crypto/src/crypto.ts
moe-dev: On it 🫡

packages/crypto/src/crypto.ts
r-n-o: Yeah makes sense to me!

packages/crypto/README.md
```diff
@@ -1,7 +1,38 @@
...
+const senderKeyPair = generateP256KeyPair();
+const receiverKeyPair = generateP256KeyPair();
+
+const receiverPublicKeyUncompressed = uncompressRawPublicKey(
+  uint8ArrayFromHexString(receiverKeyPair.publicKey),
+);
+
+const plaintext = "Hello, this is a secure message!";
+const encryptedData = hpkeEncrypt({
+  plainText: plaintext,
```
r-n-o, line 21: I think this needs to be updated now that you take a uint8array instead of a string

r-n-o (approved):
No idea why tests are failing, it doesn't seem related to your PR at first glance 👀 

Once you have a green build (and have fixed the minor thing I flagged in the README): ship it! 🚀

PR #300, opened 2024-06-15
signing session api

AlwaysBCoding:
Changes the `authBundle` string storage value for passkey sessions to a `signingSession` object that includes both the `authBundle` and the `expirationTimestamp` to make it easier for application code to check if a signingSession is active vs. expired and to auto-remove the session if it's expired.

open question if we can delete the `authBundle` value and just use `signingSession.authBundle` or if we need to keep it because people are actively using that API?


packages/sdk-react/src/contexts/TurnkeyContext.tsx
```diff
@@ -64,6 +64,15 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
 
           currentClient = authIframeClient;
         }
+      } else {
```
andrewkmin, line 67: A bit confused by this `else`. Wouldn't this mean we weren't able to get the auth bundle? 

Also should we now add a check for a signing session? (in addition to auth bundle)

packages/sdk-browser/src/sdk-client.ts
```diff
@@ -160,13 +160,28 @@ export class TurnkeyBrowserSDK {
...
   };
 
   getAuthBundle = async (): Promise<string | undefined> => {
     return await getStorageValue(StorageKeys.AuthBundle);
   };
+
+  getSigningSession = async (): Promise<SigningSession | undefined> => {
+    const signingSession: SigningSession | undefined = await getStorageValue(StorageKeys.SigningSession);
+    if (signingSession) {
+      if (signingSession.expirationTimestamp > Date.now()) {
```
andrewkmin, line 175: If we're always going to expect a valid `expirationTimestamp` to be provided, then should we just combine these two lines? --> `signingSession?.expirationTimestamp...`

packages/sdk-browser/src/sdk-client.ts
```diff
@@ -160,13 +160,28 @@ export class TurnkeyBrowserSDK {
...
+        return signingSession;
+      } else {
+        await removeStorageValue(StorageKeys.SigningSession);
+        return undefined;
+      }
+    } else {
```
andrewkmin, line 181: Minor stylistic thing: we can remove this else branch and just have a final `return undefined` at the end of this function

AlwaysBCoding: addressed

packages/sdk-react/src/contexts/TurnkeyContext.tsx
AlwaysBCoding: updated the logic for the getCurrentClient function

packages/sdk-browser/src/sdk-client.ts
```diff
@@ -159,28 +159,21 @@ export class TurnkeyBrowserSDK {
 
   logoutUser = async (): Promise<boolean> => {
     await removeStorageValue(StorageKeys.CurrentUser);
-    await removeStorageValue(StorageKeys.AuthBundle);
```
andrewkmin, line 162: I was thinking we could leave these getters/deletions in for now, but remove the _creation path_ for `AuthBundle`s (such that if you have an `AuthBundle`, things will still work, but any time you create a new session, it'll create a `Session`)

But now that I think about it, this is fine too, especially as it'll simplify our logic on the react context side of things. Let's just be sure to make note of this change in the changelog

packages/sdk-browser/src/models.ts
```diff
@@ -22,5 +22,5 @@ export type EmbeddedAPIKey = {
 
 export interface SigningSession {
```
andrewkmin, line 23: Should we rename this to `ReadWriteSession`? Given `SigningSession` could be interpreted as for signing transactions, whereas this "session" is enabling both reading from + writing to Turnkey

packages/sdk-browser/src/sdk-client.ts
```diff
@@ -159,28 +159,21 @@ export class TurnkeyBrowserSDK {
...
 
-  getAuthBundle = async (): Promise<string | undefined> => {
-    return await getStorageValue(StorageKeys.AuthBundle);
-  };
-
   getSigningSession = async (): Promise<SigningSession | undefined> => {
     const signingSession: SigningSession | undefined = await getStorageValue(StorageKeys.SigningSession);
     if (signingSession) {
-      if (signingSession.expirationTimestamp > Date.now()) {
+      if (signingSession.sessionExpiry > Date.now()) {
```
andrewkmin, line 170: `signingSession?.sessionExpiry`?

andrewkmin: And otherwise we can just return `await removeStorageValue(StorageKeys.SigningSession);`

So in all, something like

```
getSigningSession = async (): Promise<SigningSession | undefined> => {
    const signingSession: SigningSession | undefined = await getStorageValue(StorageKeys.SigningSession);
    if (signingSession?.sessionExpiry > Date.now()) {
      return signingSession;
    }
 
    await removeStorageValue(StorageKeys.SigningSession);
    return;
  }
```

andrewkmin (commented):
Looking good just some followups on naming stuff and https://github.com/tkhq/sdk/pull/301 is ready to be rebased off of.

Might be useful to include a changeset with this PR -- `pnpm run changeset` -- that highlights changes, especially for people who were already using `AuthBundle`

packages/sdk-browser/src/sdk-client.ts
AlwaysBCoding: I think this is ok the way it is because you aren't calling `removeStorageValue()` unless there's an active session that you need to expoire

packages/sdk-browser/src/models.ts
AlwaysBCoding: implemented.

packages/sdk-browser/src/sdk-client.ts
AlwaysBCoding: addressed

andrewkmin (approved):
🔥🔥🔥

PR #303, opened 2024-06-19
Add oauth endpoints, Add oauth to allowed proxy methods

moe-dev:
## Did you add a changeset?


examples/with-federated-passkeys/src/pages/api/createSubOrg.ts
```diff
@@ -57,6 +57,7 @@ export default async function createUser(
                 attestation: createSubOrgRequest.attestation,
               },
             ],
+            oauthProviders: [],
```
r-n-o, line 60: interesting. This means is a "breaking" change of sorts, at least at a code level. When customers update the SDK they'll have to update code as well. Not the end of the world but a bit annoying. As long as we're explicit about this in the `CHANGELOG` it shouldn't be a problem, it's a pretty trivial change to make!

AlwaysBCoding:
is there a way to test this flow from the UI? I know you have a branch on the `demo-embedded-wallet` repo is that up to date?

examples/with-federated-passkeys/src/pages/api/createSubOrg.ts
moe-dev: Yessir! CHANGELOG + make this a major release, and we should be good :)

AlwaysBCoding: so if we're following semver this would be `1.0.0` for `sdk-browser` and `sdk-server`. I think that's fine to do and just start developing against the 1.x.x versions going forward, but it's probably something to be conscious of that we might not want to always have to bump major versions if we're constantly changing the activity API. So could possibly have a flow in the future for making backwards compatible changes where the legacy activity syntax still resolves.

AlwaysBCoding (approved):

AlwaysBCoding:
we're going to do a 1.0.0 release of sdk-server, sdk-browser and sdk-react off this PR

PR #310, opened 2024-07-11
New release for Sdk Server, Browser, Http and associated dependents

moe-dev:
## Did you add a changeset?


packages/http/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @turnkey/http
 
+## 2.11.0
+
+### Minor Changes
+
+- Update to use new endpoints. Including CREATE_READ_WRITE_SESSION and CREATE_API_KEYS_V2
```
andrewkmin, line 7: Can we add some more color here? For `CREATE_API_KEYS_V2`, maybe something mentioning the addition of `SECP256K1` keys as a valid authenticator

And for `CREATE_READ_WRITE_SESSION`, a short description of what it enables.

Ideally we can use this changelog as a historical record of the new functionality we add on a release-to-release basis (and one day, it would be super dope to have a full fledged changelog type blog a la https://resources.arc.net/hc/en-us/articles/20498293324823-Arc-for-macOS-2024-Release-Notes)

andrewkmin (approved):
Overall this is great! I think we'll need to update the `codegen.js` files for both `sdk-browser` and `sdk-server`, and then run `node codegen.js` in both of those directories with the added `CREATE_API_KEYS_V2` (if i'm not mistaken)

In any case, giving an approval for now 🚢

moe-dev:
> CREATE_API_KEYS_V2

Codegen already updated ser 🫡

packages/http/CHANGELOG.md
moe-dev: 100p

PR #317, opened 2024-08-02
read+write session

andrewkmin:
## Summary & Motivation
Implements login + read/write sessions.

Also adds documentation/does a lil cleanup

## How I Tested These Changes
Add an example similar to Email Auth

## Did you add a changeset?

### DIFF
--- a/packages/sdk-browser/src/sdk-client.ts
+++ b/packages/sdk-browser/src/sdk-client.ts
@@ -17,7 +17,12 @@ import { TurnkeyRequestError } from "./__types__/base";
 import { TurnkeySDKClientBase } from "./__generated__/sdk-client-base";
 import type * as SdkApiTypes from "./__generated__/sdk_api_types";
 
-import type { User, SubOrganization, ReadWriteSession } from "./models";
+import type {
+  User,
+  SubOrganization,
+  ReadWriteSession,
+  Passkey,
+} from "./models";
 import {
   StorageKeys,
   getStorageValue,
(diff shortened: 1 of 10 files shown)


packages/sdk-browser/src/sdk-client.ts
```diff
@@ -293,8 +382,8 @@ export class TurnkeyPasskeyClient extends TurnkeyBrowserClient {
         {
           apiKeyName: `Session Key ${String(Date.now())}`,
           publicKey,
-          expirationSeconds: expirationSeconds ?? DEFAULT_SESSION_EXPIRATION,
-          curveType: curveType ?? "API_KEY_CURVE_P256",
+          expirationSeconds,
+          curveType: "API_KEY_CURVE_P256",
```
andrewkmin, line 386: By default all embedded API keys will be P256 keys

moe-dev (approved):
Awesome 🙏

packages/sdk-browser/src/sdk-client.ts
```diff
@@ -207,9 +248,39 @@ export class TurnkeyBrowserClient extends TurnkeySDKClientBase {
...
+    });
+
+    // store auth bundle in local storage
+    await setStorageValue(StorageKeys.ReadWriteSession, {
+      authBundle: readWriteSessionResult!.credentialBundle,
+      sessionExpiry: Date.now() + Number(expirationSeconds) * 1000,
+    });
+
+    return readWriteSessionResult;
+  };
```
taylorjdawson, line 283: @andrewkmin I assume I could use this method like this? 
```
authIframeClient.loginWithReadWriteSession(authIframeClient?.iframePublicKey)
```
if the authIframeClient already has this public key I wonder if this method could be on the `TurnkeyIframeClient` so it doesn't have to be passed in?

andrewkmin: > authIframeClient.loginWithReadWriteSession(authIframeClient?.iframePublicKey)

Yeah that's exactly right!

> if the authIframeClient already has this public key I wonder if this method could be on the TurnkeyIframeClient so it doesn't have to be passed in?

Currently, any of the clients can initiate a `login with read write session` activity (e.g. `passkeyClient.loginWithReadWriteSession` is valid as well). You're right in that the TurnkeyIframeClient could override the base implementation (that lives on `TurnkeyBrowserClient`) and use something like `this.iframePublicKey()`

taylorjdawson: tightt that would be so slicckkk!

PR #321, opened 2024-08-06
padding fixes

andrewkmin:
## Summary & Motivation
$title

There are cases where we want to manually pad buffers to get to a certain length, primarily when we're dealing with jwks. 

The nature of this change is similar in spirit to another change we recently rolled out for our iframes: https://github.com/tkhq/frames/pull/44

## How I Tested These Changes
Unit, smoke via https://github.com/tkhq/sdk/pull/322

## Did you add a changeset?

### DIFF
--- a/packages/api-key-stamper/src/tink/elliptic_curves.ts
+++ b/packages/api-key-stamper/src/tink/elliptic_curves.ts
@@ -7,6 +7,10 @@
  */
 
 import * as Bytes from "./bytes";
+import {
+  DEFAULT_JWK_MEMBER_BYTE_LENGTH,
+  uint8ArrayFromHexString,
+} from "@turnkey/encoding";
 
 /**
  * P-256 only
@@ -34,12 +38,20 @@ function byteArrayToInteger(bytes: Uint8Array): bigint {
   return BigInt("0x" + Bytes.toHex(bytes));
 }
 
-/** Converts bigint to byte array. */
-function integerToByteArray(i: bigint): Uint8Array {
+/** Converts bigint to byte array. This implementation has been modified to optionally augment the resulting byte array to a certain length. */
+function integerToByteArray(i: bigint, length?: number): Uint8Array {
   let input = i.toString(16);
   // If necessary, prepend leading zero to ensure that input length is even.
   input = input.length % 2 === 0 ? input : "0" + input;
-  return Bytes.fromHex(input);
+  if (!length) {
+    return Bytes.fromHex(input);
+  }
+  if (input.length / 2 > length) {
+    throw new Error(
+      "hex value cannot fit in a buffer of " + length + " byte(s)"
+    );
+  }
+  return uint8ArrayFromHexString(input, length);
 }
 
 /** Returns true iff the ith bit (in lsb order) of n is set. */
@@ -151,8 +163,14 @@ export function pointDecode(point: Uint8Array): JsonWebKey {
   const result: JsonWebKey = {
     kty: "EC",
     crv: "P-256",
-    x: Bytes.toBase64(integerToByteArray(x), /* websafe */ true),
-    y: Bytes.toBase64(integerToByteArray(y), /* websafe */ true),
+    x: Bytes.toBase64(
+      integerToByteArray(x, DEFAULT_JWK_MEMBER_BYTE_LENGTH),
+      /* websafe */ true
+    ),
+    y: Bytes.toBase64(
+      integerToByteArray(y, DEFAULT_JWK_MEMBER_BYTE_LENGTH),
+      /* websafe */ true
+    ),
     ext: true,
   };
   return result;
--- a/packages/api-key-stamper/src/tink/elliptic_curves.ts
+++ b/packages/api-key-stamper/src/tink/elliptic_curves.ts
@@ -1,6 +1,6 @@
 /**
  * Code modified from https://github.com/google/tink/blob/6f74b99a2bfe6677e3670799116a57268fd067fa/javascript/subtle/elliptic_curves.ts
- * - The implementation of integerToByteArray has been modified to optionally augment the resulting byte array to a certain length.
+ * - The implementation of integerToByteArray has been modified to augment the resulting byte array to a certain length.
  *
  * @license
  * Copyright 2020 Google LLC
--- a/packages/api-key-stamper/src/tink/elliptic_curves.ts
+++ b/packages/api-key-stamper/src/tink/elliptic_curves.ts
@@ -1,5 +1,6 @@
 /**
  * Code modified from https://github.com/google/tink/blob/6f74b99a2bfe6677e3670799116a57268fd067fa/javascript/subtle/elliptic_curves.ts
+ * - The implementation of integerToByteArray has been modified to optionally augment the resulting byte array to a certain length.
  *
  * @license
  * Copyright 2020 Google LLC
@@ -7,10 +8,6 @@
  */
 
 import * as Bytes from "./bytes";
-import {
-  DEFAULT_JWK_MEMBER_BYTE_LENGTH,
-  uint8ArrayFromHexString,
-} from "@turnkey/encoding";
 
 /**
  * P-256 only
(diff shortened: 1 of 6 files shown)


packages/http/src/webauthn-json/base64url.ts
```diff
@@ -17,7 +17,7 @@ export function base64urlToBuffer(
   for (let i = 0; i < str.length; i++) {
     byteView[i] = str.charCodeAt(i);
   }
-  return buffer;
+  return byteView;
```
andrewkmin, line 20: This is unrelated to this specific change, but up until now this function was essentially just returning an empty buffer.

packages/api-key-stamper/src/utils.ts
```diff
@@ -1,5 +1,11 @@
 import { pointDecode } from "./tink/elliptic_curves";
-import { stringToBase64urlString } from "@turnkey/encoding";
+import {
+  stringToBase64urlString,
+  base64urlToBuffer,
+  uint8ArrayToHexString,
+} from "@turnkey/encoding";
+
+const DEFAULT_JWK_MEMBER_BYTE_LENGTH = 32;
```
moe-dev, line 8: super tiny nit, if we plan on setting this in multiple files maybe we can export it from @turnkey/encoding

moe-dev (approved):
LGTM assuming tests pass!

packages/api-key-stamper/src/__tests__/utils-test.ts
```diff
@@ -0,0 +1,33 @@
...
+import { uint8ArrayFromHexString } from "@turnkey/encoding";
+
+Object.defineProperty(globalThis, "crypto", {
+  value: {
+    getRandomValues: (arr) => crypto.randomBytes(arr.length),
+  },
+});
+
+test("correctly converts turnkey API key to JWK", function () {
+  for (let i = 0; i < 100; i++) {
```
andrewkmin, line 16: Jest doesn't like indefinite (while: true) loops, so just picked an arbitrary fairly large value here

packages/api-key-stamper/src/utils.ts
```diff
@@ -21,21 +21,21 @@ export function convertTurnkeyApiKeyToJwk(input: {
   // Ensure that each of the constituent parts are sufficiently padded
   jwkCopy.d = hexStringToBase64urlString(uncompressedPrivateKeyHex);
 
-  // Manipulate x and y
-  const decodedX = base64urlToBuffer(jwkCopy.x!);
-  const paddedX = hexStringToBase64urlString(
-    uint8ArrayToHexString(new Uint8Array(decodedX)),
-    DEFAULT_JWK_MEMBER_BYTE_LENGTH
-  );
+  // // Manipulate x and y
```
andrewkmin, line 24: temporarily commenting this logic out in order to be able to reproduce the issue when converting turnkey API key to jwk

packages/api-key-stamper/src/utils.ts
```diff
@@ -23,17 +56,29 @@ function hexStringToUint8Array(input: string): Uint8Array {
...
   );
+
+  if (!length) {
+    return buffer;
+  }
+
+  // If a length is specified, ensure we sufficiently pad
+  let paddedBuffer = new Uint8Array(length);
+  paddedBuffer.set(buffer, length - buffer.length);
+  return paddedBuffer;
```
r-n-o, line 74: I think it's important to handle cases where `length` is _less_ than `buffer.length`. In this case, maybe an error?

packages/api-key-stamper/src/utils.ts
```diff
@@ -9,12 +15,39 @@ export function convertTurnkeyApiKeyToJwk(input: {
...
-  return jwk;
+  const decodedY = base64urlToBuffer(jwkCopy.y!);
+  const paddedY = hexStringToBase64urlString(
+    uint8ArrayToHexString(new Uint8Array(decodedY)),
+    DEFAULT_JWK_MEMBER_BYTE_LENGTH
+  );
+
+  jwkCopy.d = paddedD;
+  jwkCopy.x = paddedX;
+  jwkCopy.y = paddedY;
```
r-n-o, line 42: What led to padding (x, y, d) as opposed to just `d`? Given the jwk is already created I think we're good with the existing components within it (x, y)

packages/api-key-stamper/src/utils.ts
```diff
@@ -9,12 +15,39 @@ export function convertTurnkeyApiKeyToJwk(input: {
 
   const jwk = pointDecode(hexStringToUint8Array(compressedPublicKeyHex));
```
r-n-o, line 16: If x and y aren't padded for whatever reason, let's fix this inside of `pointDecode` so we don't have to extract, pad, and then put things "back" in the JWK:
https://github.com/tkhq/sdk/blob/08e2b8c30848498ed0850aa1e1ede67775b001e5/packages/api-key-stamper/src/tink/elliptic_curves.ts#L154-L155

packages/api-key-stamper/src/tink/elliptic_curves.ts
```diff
@@ -7,6 +7,10 @@
  */
 
 import * as Bytes from "./bytes";
+import {
+  DEFAULT_JWK_MEMBER_BYTE_LENGTH,
+  uint8ArrayFromHexString,
+} from "@turnkey/encoding";
```
r-n-o, line 13: Mmm, this seems a bit off because we have a vendor'd library import our encoding lib :/

packages/api-key-stamper/src/tink/elliptic_curves.ts
```diff
@@ -34,12 +38,20 @@ function byteArrayToInteger(bytes: Uint8Array): bigint {
   return BigInt("0x" + Bytes.toHex(bytes));
 }
 
-/** Converts bigint to byte array. */
-function integerToByteArray(i: bigint): Uint8Array {
+/** Converts bigint to byte array. This implementation has been modified to optionally augment the resulting byte array to a certain length. */
```
r-n-o, line 41: Let's insert this comment at the top of the file so it's obvious (in case someone is trying to update this file down the road)

packages/api-key-stamper/src/tink/elliptic_curves.ts
```diff
@@ -34,12 +38,20 @@ function byteArrayToInteger(bytes: Uint8Array): bigint {
...
-  return Bytes.fromHex(input);
+  if (!length) {
+    return Bytes.fromHex(input);
+  }
+  if (input.length / 2 > length) {
+    throw new Error(
+      "hex value cannot fit in a buffer of " + length + " byte(s)"
+    );
+  }
+  return uint8ArrayFromHexString(input, length);
```
r-n-o, line 54: I don't think this is necessary. Let's **require** the `length` argument (`integerToByteArray` isn't called anywhere with an undefined length), and pad with `0` without relying on our library function. We can also remove the special case of an odd-length `input` and just rely on generic padding:
```typescript
function integerToByteArray(i: bigint, length: number): Uint8Array {
  let input = i.toString(16);
  let numHexChars = length * 2
  if numHexChars < input.length {
    throw new Error(`cannot pack integer with ${input.length} hex chars into$ {length} bytes`);
  } else {
    let padding = "0".repeat(numHexChars - input.length);
  }
  return Bytes.fromHex(padding + input);
}
```

packages/encoding/src/index.ts
```diff
@@ -1,12 +1,63 @@
...
  */
+export const DEFAULT_JWK_MEMBER_BYTE_LENGTH = 32;
+
 export function stringToBase64urlString(input: string): string {
   // string to base64 -- we do not rely on the browser's btoa since it's not present in React Native environments
   const base64String = btoa(input);
   return base64StringToBase64UrlEncodedString(base64String);
 }
 
+export function base64urlToBuffer(baseurl64String: string): ArrayBuffer {
```
r-n-o, line 12: Aside from tests this isn't used anywhere

packages/encoding/src/index.ts
```diff
@@ -1,12 +1,63 @@
...
+  // Base64url to Base64
+  const padding = "==".slice(0, (4 - (baseurl64String.length % 4)) % 4);
+  const base64String =
+    baseurl64String.replace(/-/g, "+").replace(/_/g, "/") + padding;
+
+  // Base64 to binary string
+  const str = atob(base64String);
```
r-n-o, line 19: Generally we should avoid relying on things that are "magically" available. This may work in Node but might not work in React Native, for example (I'd like our core "encoding" functions to be pure JS so we don't have to deal with polyfills or worry about different implementations across JS runtimes)

packages/encoding/src/index.ts
```diff
@@ -1,12 +1,63 @@
...
+  // Binary string to buffer
+  const buffer = new ArrayBuffer(str.length);
+  const byteView = new Uint8Array(buffer);
+  for (let i = 0; i < str.length; i++) {
+    byteView[i] = str.charCodeAt(i);
+  }
+  return byteView;
+}
+
+export function bufferToBase64url(buffer: ArrayBuffer): string {
```
r-n-o, line 30: Same here. This doesn't seem to be used anywhere outside of the new tests.

andrewkmin: I was planning on consolidating various utils in `packages/encoding` instead of having the same util functions in `http` (specifically `packages/http/src/webauthn-json/base64url.ts`)

packages/encoding/src/index.ts
andrewkmin: That makes sense! This also was cut/pasted from `packages/http/src/webauthn-json/base64url.ts` / `packages/webauthn-stamper/src/webauthn-json/base64url.ts` . But if this is "over-consolidation", happy to take this out. That and the fact that this isn't pure JS

packages/encoding/src/index.ts
andrewkmin: (same as https://github.com/tkhq/sdk/pull/321#discussion_r1717251137)

packages/encoding/src/index.ts
andrewkmin: ended up removing for now! maybe we can move it back, consolidate utilities in `encoding` down the line, along with a pureJS implementation of `atob` (similar to the `btoa` impl we have vendored in)

r-n-o (approved):
Looks good!

PR #323, opened 2024-08-08
Pad SDK browser Import Keys

omkarshanbhag:
## Summary & Motivation

Some JWK buffers that we construct need to be manually padded because they may get truncated during conversion between different formats in the case that the first byte(s) are 0's. This protects against error cases where the JWK components are found to be less than 32 bytes that some clients seem to run into. 

Some similar changes: 
- https://github.com/tkhq/frames/pull/44
- https://github.com/tkhq/sdk/pull/321

## How I Tested These Changes

- unittests added

## Did you add a changeset?

Yes -- changeset has been added


packages/sdk-browser/src/utils.ts
```diff
@@ -28,11 +28,30 @@ export const createEmbeddedAPIKey = async (
...
+  const { x, y } = getXYComponentsFromHexString(targetPublicKey);
+
   const targetKey = await crypto.subtle.importKey(
-    "raw",
-    targetKeyBytes,
+    "jwk",
+    {
+      kty: "EC",
+      crv: "P-256",
+      x: base64UrlEncode(uint8ArrayFromHexStringPadded(x, 32)),
```
andrewkmin, line 51: Down to call this `uint8ArrayFromHexStringWithPadding`, or add the length argument as optional to the existing `uint8ArrayFromHexString` implementation

packages/sdk-browser/src/utils.ts
```diff
@@ -28,11 +28,30 @@ export const createEmbeddedAPIKey = async (
...
+      y: base64UrlEncode(uint8ArrayFromHexStringPadded(y, 32)),
+      ext: true
+    }, // this is what needs to be changed
```
andrewkmin, line 54: Can probably remove?

andrewkmin (approved):
Some minor comments but looks good overall! 🥳 just needs some smoke testing

packages/encoding/src/__tests__/index-test.ts
```diff
@@ -69,3 +69,27 @@ test("uint8ArrayFromHexString", async function () {
...
+
+
+  // TOO LONG - test a hex string with less bytes than the "length" parameter provided 
+  const hexString2 =
+    "5234d08dfa2c815f3097b8ba848a28172e85bec78886e8e201afccb166fcfafbfcfd"; // length is 34 bytes, so no additional padding will be added
+  const expectedUint8Array2 = new Uint8Array([
+    0, 0, 82, 52, 208, 141, 250, 44, 129, 95, 48, 151, 184, 186, 132, 138, 40, 23, 46,
+    133, 190, 199, 136, 134, 232, 226, 1, 175, 204, 177, 102, 252, 250, 251, 252, 253,
+  ]);
+  expect(uint8ArrayFromHexString(hexString, 32)).toEqual(expectedUint8Array); // Hex string => Uint8Array
```
r-n-o, line 94: doesn't seem right, you're passing `hexString` instead of `hexString2`? `expectedUint8Array2` seems unused too

packages/sdk-browser/src/utils.ts
```diff
@@ -28,11 +28,30 @@ export const createEmbeddedAPIKey = async (
...
+  // Function to extract x and y components from a hex string public key
+  function getXYComponentsFromHexString(publicKeyHex: string) {
+    publicKeyHex = publicKeyHex.replace(/^0x/, '');
+    if (!publicKeyHex.startsWith('04')) {
+      throw new Error("Public key is not in uncompressed format");
+    }
+    const x = publicKeyHex.slice(2, 66);
+    const y = publicKeyHex.slice(66, 130);
+    return { x, y };
+  }
```
r-n-o, line 40: This is reminiscent of `pointDecode`: https://github.com/tkhq/sdk/blob/08e2b8c30848498ed0850aa1e1ede67775b001e5/packages/api-key-stamper/src/tink/elliptic_curves.ts#L135-L159

Except that here we're decoding from an **uncompressed** public key (which has both x and y components). Might make sense to unify and support both uncompressed and compressed hex strings in `pointDecode`?`pointDecode` ensures the x and y are valid P256 point coordinates, which is nice. Not sure if we could provide the same guarantee with uncompressed pub keys?

omkarshanbhag: Thanks for the callout here -- I was just looking into uncompressed vs compressed. I'll look into these and consider adding uncompressed support to PointDecode

packages/encoding/src/__tests__/index-test.ts
omkarshanbhag: Yeah -- oversight on my end! Good catch

packages/sdk-browser/src/utils.ts
omkarshanbhag: Added uncompressed checking to point decode

packages/encoding/src/__tests__/index-test.ts
```diff
@@ -110,3 +110,23 @@ test("hexStringToBase64url", async function () {
     hexStringToBase64url("0100", 1);
   }).toThrow("hex value cannot fit in a buffer of 1 byte(s)");
 });
+
+// Test array padding for uint8ArrayFromHexString
+// Convert hex string to uint8 array and pads it with zeroes to a particular length if its less
+test("uint8ArrayFromHexString Test padding", async function () {
```
andrewkmin, line 116: Let's combine these with the `uint8ArrayFromHexString` tests above in line 64

packages/api-key-stamper/src/tink/elliptic_curves.ts
```diff
@@ -136,35 +136,86 @@ function getY(x: bigint, lsb: boolean): bigint {
...
- * Decodes a public key in _compressed_ format.
+ *
+ * Given x and y coordinates of a JWK, checks whether these are valid points on
+ * the P-256 elliptic curve.
+ *
+ * P-256 only
+ *
+ * @param x x-coordinate
+ * @param y y-coordinate
+ * @return boolean validitiy
```
andrewkmin, line 147: tiny nit: `validity`

packages/api-key-stamper/src/tink/elliptic_curves.ts
```diff
@@ -136,35 +136,86 @@ function getY(x: bigint, lsb: boolean): bigint {
...
+      "Invalid length: point is not in compressed or uncompressed format"
+    );
   }
-  const lsb = point[0] === 3; // point[0] must be 2 (false) or 3 (true).
-  const x = byteArrayToInteger(point.subarray(1, point.length));
-  const p = getModulus();
-  if (x < BigInt(0) || x >= p) {
-    throw new Error("x is out of range");
+  // Decodes point if its length and first bit match the compressed format
+  if ((point[0] === 2 || point[0] === 3) && point.length == compressedLength) {
```
andrewkmin, line 177: We can draw inspiration from the original implementation, which we're getting closer to given our support for multiple point types (compressed + uncompressed): https://github.com/tink-crypto/tink/blob/6f74b99a2bfe6677e3670799116a57268fd067fa/javascript/subtle/elliptic_curves.ts#L586-L652

We can borrow some of their implementation for checks, like
```
if (point.length !== 1 + 2 * fieldSize || point[0] !== 4) {
        throw new InvalidArgumentsException('invalid point');
      }
```

omkarshanbhag: hmmm it looks like they have an argument called `pointFormatType` which can be something like `UNCOMPRESSED` or `COMPRESSED` and then within the function body they check whether given a point format type, the input matches the format. 

Do we want to mimic this entirely? 

Right now I'm just doing a check to match compressed format on 177, in which I decode compressed. 

```
// Decodes point if its length and first bit match the compressed format
  if ((point[0] === 2 || point[0] === 3) && point.length == compressedLength) {
```

And if it doesn't match 177 then on 194 I check for matching uncompressed (effectively the inverse of what they check for). 

```
    // Decodes point if its length and first bit match the uncompressed format
  } else if (point[0] === 4 && point.length == uncompressedLength) {
```

Just to clarify -- are you saying we should move towards passing in a parameter for point type???

packages/api-key-stamper/src/tink/elliptic_curves.ts
```diff
@@ -136,35 +136,86 @@ function getY(x: bigint, lsb: boolean): bigint {
...
  */
 export function pointDecode(point: Uint8Array): JsonWebKey {
   const fieldSize = fieldSizeInBytes();
-
-  if (point.length !== 1 + fieldSize) {
-    throw new Error("compressed point has wrong length");
-  }
-  if (point[0] !== 2 && point[0] !== 3) {
-    throw new Error("invalid format");
+  const compressedLength = fieldSize + 1;
```
r-n-o, line 166: This looks like a lot of new code, and we're in a **vendored** file! At this point it'd be really hard for someone to understand how to upgrade tink if we ever had to.

Still, can you add a summary of what was done in this PR at the top of the file?

packages/api-key-stamper/src/tink/elliptic_curves.ts
```diff
@@ -136,35 +136,86 @@ function getY(x: bigint, lsb: boolean): bigint {
...
+ * @param y y-coordinate
+ * @return boolean validity
+ */
+function validateUncompressedXY(x: bigint, y: bigint): boolean {
+  const p = getModulus();
+  const a = p - BigInt(3);
+  const b = getB();
+  const rhs = ((x * x + a) * x + b) % p;
+  const lhs = y ** BigInt(2) % p;
+  return lhs === rhs;
```
r-n-o, line 155: aka `isP256CurvePoint` :)

packages/api-key-stamper/src/__tests__/elliptic-curves-test.ts
```diff
@@ -0,0 +1,55 @@
+import { test, expect } from "@jest/globals";
+import { pointDecode } from "../tink/elliptic_curves";
+import {
+  uint8ArrayFromHexString,
+  uint8ArrayToHexString,
+} from "@turnkey/encoding";
+
+test("pointDecode -> uncompressed invalid", async function () {
```
r-n-o, line 8: I don't see a test for compressed keys and invalid lengths cases. Let's cover these?

packages/api-key-stamper/src/tink/elliptic_curves.ts
```diff
@@ -136,35 +136,86 @@ function getY(x: bigint, lsb: boolean): bigint {
...
+ * Given x and y coordinates of a JWK, checks whether these are valid points on
+ * the P-256 elliptic curve.
+ *
+ * P-256 only
+ *
+ * @param x x-coordinate
+ * @param y y-coordinate
+ * @return boolean validity
+ */
+function validateUncompressedXY(x: bigint, y: bigint): boolean {
```
r-n-o, line 149: Something to note here: it's great that you've thought about validating that the point is on the curve, but this is going a bit "above and beyond": when this JWK structure is _actually_ imported with `crypto.importKey`, this validation will occur anyway (`crypto.importKey` will throw an exception)

packages/api-key-stamper/src/tink/elliptic_curves.ts
omkarshanbhag: Ah yeah wanted to follow the earlier suggestion of moving the uncompressed Pub key point into pointDecode -- didn't realize this is a vendored file until later will definitely add some context on top.

Also -- it seems like we've changed the file significantly from what the original implementation is for our own purposes. The original implementation also does have the capacity to check uncompressed points!

packages/api-key-stamper/src/__tests__/elliptic-curves-test.ts
omkarshanbhag: I'll add that in!

packages/api-key-stamper/src/tink/elliptic_curves.ts
omkarshanbhag: Right yeah makes sense -- wanted to provide similar guarantees within our code for compressed as we did for uncompressed and catch errors within our own code first but I see how it's an error case either way.

packages/api-key-stamper/src/tink/elliptic_curves.ts
omkarshanbhag: (all this is to say that our implementation is already quite different from what's in the original, and this actually kinda brings it closer to the original)

packages/sdk-browser/src/__tests__/utils-test.ts
```diff
@@ -0,0 +1,14 @@
+import { test, expect } from "@jest/globals";
+import { createEmbeddedAPIKey } from "../utils";
+
+test("createEmbeddedAPIKey", async function () {
+  // Test valid uncompressed public key
+  const uncompPubKey =
+    "0400d2eb47be2006c29db5fe9941dd686d19ddeea85a0328894f08091f6b5be9b366c4872345594c12a7f7c47c62dd8074542934820fce5ee0ddc55d6d1d8dd312";
+  await expect(createEmbeddedAPIKey(uncompPubKey)).resolves.not.toThrow();
```
andrewkmin, line 8: We could do a basic check of the return value: verify that the resulting `authBundle` and `publicKey` are valid, non empty strings

andrewkmin (commented):
looking good!

packages/api-key-stamper/src/index.ts
```diff
@@ -78,3 +78,7 @@ export class ApiKeyStamper {
     };
   }
 }
+
+import { pointDecode } from "./tink/elliptic_curves";
```
andrewkmin, line 82: I believe you can consolidate this into one line: `export { pointDecode } from "./tink/elliptic_curves";`

packages/api-key-stamper/src/tink/elliptic_curves.ts
```diff
@@ -1,6 +1,8 @@
 /**
  * Code modified from https://github.com/google/tink/blob/6f74b99a2bfe6677e3670799116a57268fd067fa/javascript/subtle/elliptic_curves.ts
  * - The implementation of integerToByteArray has been modified to augment the resulting byte array to a certain length.
+ * - The implementation of PointDecode has been modified to decode both compressed and uncompressed points by checking for correct format
+ * - Methoed isP256CurvePoint added to check whether an uncompressed point is valid
```
andrewkmin, line 5: tiiiny nit: `Method`

packages/api-key-stamper/src/__tests__/elliptic-curves-test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+test("pointDecode -> uncompressed valid", async function () {
+  // Valid uncompressed public key with 00 as the first bit (test against 'x' field of JWK getting truncated)
+  const uncompPubKey =
+    "0400d2eb47be2006c29db5fe9941dd686d19ddeea85a0328894f08091f6b5be9b366c4872345594c12a7f7c47c62dd8074542934820fce5ee0ddc55d6d1d8dd312";
+  const jwk = pointDecode(uint8ArrayFromHexString(uncompPubKey));
+  expect(jwk.x).toBeDefined();
+  expect(jwk.y).toBeDefined();
+
+  // Convert x value to make sure it matches first half of uncompressed key WITHOUT truncating the first 0 bit
```
andrewkmin, line 25: This is great! If you'd like, you could also try importing the resulting JWK as a sanity check as well (e.g. the `crypto.subtle.importKey...` codepath that happens in `packages/sdk-browser/src/utils.ts`).

Note: if you do go down this path, you might need to import `crypto` to have it defined in the test setting:

```
Object.defineProperty(globalThis, "crypto", {
  value: {
    getRandomValues: (arr) => crypto.randomBytes(arr.length),
    subtle: crypto.subtle,
  },
});
```

Can use `packages/api-key-stamper/src/__tests__/utils-test.ts` in [this diff](https://github.com/tkhq/sdk/pull/322/files#diff-b3f5d0933ac8568c7a900ac4c1b3a23afa74499ec22d8b6d0d7e44d200192b58) as a reference

andrewkmin: Totally fine if you think this is overkill given some of the manual validation you're doing here

packages/api-key-stamper/src/index.ts
omkarshanbhag: Done

packages/api-key-stamper/src/tink/elliptic_curves.ts
omkarshanbhag: done

packages/api-key-stamper/src/__tests__/elliptic-curves-test.ts
omkarshanbhag: Thanks for the suggestion here!! I'll leave the manual validation as is for right now so we can get this over the line in a timely manner

r-n-o (approved):
🏅

PR #335, opened 2024-08-21
Simple `@turnkey/sdk-server` example

r-n-o:
## Summary & Motivation
Simple example to show how one can use `@turnkey/sdk-browser` to make API calls. Nothing fancy, but could be a good base for node scripts using the Turnkey API.

Bonus: spotted and fixed a mistake in the sdk-browser package README :)

## How I Tested These Changes
Ran the new example.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,9 +71,11 @@ typings/
 # dotenv environment variables file
 .env
 .env.*
+config.json
 !.env.local.example
 !.env.test.example
 !.env.example
+!config.json.example
 
 # parcel-bundler cache (https://parceljs.org/)
 .cache
--- a/packages/sdk-server/README.md
+++ b/packages/sdk-server/README.md
@@ -16,7 +16,7 @@ $ npm install @turnkey/sdk-server
 const { Turnkey } = require("@turnkey/sdk-server");
 
 // This config contains parameters including base URLs, API credentials, and org ID
-const turnkeyConfig = JSON.parse(fs.readFileSync("./turnkey.json"), "utf8");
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -71,9 +71,11 @@ typings/
 # dotenv environment variables file
 .env
 .env.*
+config.json
```
andrewkmin, line 74: @taylorjdawson and I chatted a bit about this — we thought we should move on from storing env vars in a `config.json` and instead inject these variables from `.env` directly

andrewkmin: But that's a later discussion, probably introduced with Taylor's upcoming Nextjs 14 example

packages/sdk-server/README.md
```diff
@@ -16,7 +16,7 @@ $ npm install @turnkey/sdk-server
 const { Turnkey } = require("@turnkey/sdk-server");
 
 // This config contains parameters including base URLs, API credentials, and org ID
-const turnkeyConfig = JSON.parse(fs.readFileSync("./turnkey.json"), "utf8");
+const turnkeyConfig = JSON.parse(fs.readFileSync("./turnkey.json", "utf8"));
```
andrewkmin, line 19: 🏅

andrewkmin (approved):
👏

.gitignore
r-n-o: Nice! I agree, `.env` is more standard. When we support this we'll have a place to update (this example!). I imagine we'll have a convenience function "turnkey.fromEnv" or similar?

PR #342, opened 2024-08-26
Release V2024.8.15 - Changelog

omkarshanbhag:
## Did you add a changeset?

### DIFF
--- a/packages/http/CHANGELOG.md
+++ b/packages/http/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Patch Changes
 
-- e2f2e0b: updates to sdk-browser, sdk-server and http
+- e2f2e0b: Added two new endpoints for deleting private keys and deleting wallets
 
 ## 2.12.2
 
--- a/packages/http/CHANGELOG.md
+++ b/packages/http/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @turnkey/http
 
+## 2.12.3
+
+### Patch Changes
+
+- e2f2e0b: updates to sdk-browser, sdk-server and http
+
 ## 2.12.2
 
 ### Patch Changes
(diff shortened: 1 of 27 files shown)


packages/http/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @turnkey/http
 
+## 2.12.3
+
+### Patch Changes
+
+- e2f2e0b: updates to sdk-browser, sdk-server and http
```
andrewkmin, line 7: We can mention just the changes to `http` here (pulled from https://github.com/tkhq/sdk/pull/341 -- primarily two new endpoints, for deleting private keys + wallets)

We don't have to mention `sdk-browser` and `sdk-server` here given `http` does not depend on those libraries, and in fact the opposite is the case (those libraries depend on `http`)

packages/sdk-browser/CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # @turnkey/sdk-browser
 
+## 1.2.4
+
+### Patch Changes
+
+- e2f2e0b: updates to sdk-browser, sdk-server and http
```
andrewkmin, line 7: We can remove this line if the only real change to `sdk-browser` is the `http` bump

packages/sdk-server/CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # @turnkey/sdk-server
 
+## 1.2.4
+
+### Patch Changes
+
+- e2f2e0b: updates to sdk-browser, sdk-server and http
```
andrewkmin, line 7: Same here

andrewkmin (approved):
looks great overall, just some minor comments! 🚀

PR #348, opened 2024-09-04
Synchronous createAccountWithAddress

mshrieve:
## Summary & Motivation

The viem `createAccount` function is asynchronous, which can make a turnkey viem client tricky to use as a plug-and-play alternative to a standard viem client.
`createAccount` only needs to be asynchronous if the `signWith` parameter is a private key ID _and_ the corresponding ethereum address is not provided.

I've added a synchronous `createAccountWithAddress` which works if either the `signWith` parameter is an address, or, if not, if the `ethereumAddress` parameter is provided. 

This is a minimal change that won't affect current usage of the package.  I would actually suggest making the `createAccount` itself asynchronous and moving the code to fetch the ethereum address corresponding to a private key ID to a separate asynchronous function.

## How I Tested These Changes

I've added tests to cover the changes, as well as a small update to `.env.example`.
My deny policy tests fail... I'm not exactly which policy would give the required error.  I simply added a `EFFECT_DENY` policy to my test user I've set up for these tests.

```
"cause": [TurnkeyRequestError: Turnkey error 7: policy engine denied request explicitly (Details: [])],
    -   "message": "Failed to sign: Turnkey error 7: policy engine denied request explicitly (Details: [])",
    +   "cause": [TurnkeyRequestError: Turnkey error 7: You don't have sufficient permissions to take this action. Please add a policy granting this user permissions. Learn more at https://docs.turnkey.com/concepts/policies/quickstart....
```

My policy looks like this:

```
{
  "effect": "EFFECT_DENY",
  "consensus": "approvers.any(user, user.id == '<USER_ID>')",
  "condition": "eth.tx.to == '0x0000000000000000000000000000000000000000'"
}
```

## Did you add a changeset?

I'll add a changeset if you're ok with the changes.


mshrieve:
Also, I noticed that the deny policy tests will pass if the call to `signTransaction` does _not_ throw an error.

andrewkmin (approved):
Overall, looking really good. Thank you so much once again @mshrieve !

One minor request, think you could run `pnpm run prettier-all:write` from the root of `/sdk`? Thank you!

mshrieve:
done @andrewkmin

andrewkmin:
merged via https://github.com/tkhq/sdk/pull/354 ! thank you

Issue #121, opened 2023-09-22
`err.walk` is not a function


davidlee1435:
Hi all, I'm using Turnkey with Viem and running into an issue when trying to call `writeContract`. I suspect it's an error with Turnkey and not Viem because I am able to use Viem's `privateKeyToAccount` to successfully call `writeContract`.

Here's the part of the TypeScript script that I have that's throwing the error:

```
  const turnkeyClient = new TurnkeyClient(
    {
      baseUrl: TurnkeyConfig.baseUrl,
    },
    new ApiKeyStamper({
      apiPublicKey: TurnkeyConfig.apiPublicKey,
      apiPrivateKey: process.env.TURNKEY_API_PRIVATE_KEY,
    }),
  );

  const account = await createAccount({
    client: turnkeyClient,
    organizationId: TurnkeyConfig.organizationId,
    privateKeyId: TurnkeyConfig.privateKeyIds.minter,
  });

  const publicClient = createPublicClient({
    chain,
    transport: http(),
  });

  const walletClient = createWalletClient({
    account,
    chain,
    transport: http(),
  });
  
  // This returns a correct response
  const { request } =
    await publicClient.simulateContract({
      account,
      <my address>,
      abi: myErc20ABI,
      functionName: "mint",
      args: [recipient, amount],
    });

  const hash = await walletClient.writeContract(request);
```

**Stack trace:**
```
[Uncaught Error]: err.walk is not a function

TypeError: err.walk is not a function
    at getNodeError (/Users/davidlee/dev/peregrine/stable/packages/scripts/node_modules/viem/utils/errors/getNodeError.ts:36:38)
    at getTransactionError (/Users/davidlee/dev/peregrine/stable/packages/scripts/node_modules/viem/utils/errors/getTransactionError.ts:23:27)
    at sendTransaction (/Users/davidlee/dev/peregrine/stable/packages/scripts/node_modules/viem/actions/wallet/sendTransaction.ts:185:30)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async writeContract (/Users/davidlee/dev/peregrine/stable/packages/scripts/node_modules/viem/actions/wallet/writeContract.ts:133:16)
    at async processCommand (/Users/davidlee/dev/peregrine/stable/packages/scripts/src/mint-usdlr.ts:83:16)
```

**Dependency versions:**
```
"@turnkey/api-key-stamper": "^0.1.1",
"@turnkey/http": "^1.2.0",
[...]

PR #350, opened 2024-09-04
handle consensus

andrewkmin:
## Summary & Motivation
$title

TLDR: Previously, the `ACTIVITY_STATUS_CONSENSUS_NEEDED` case was blatantly unhandled, resulting in misleading error messages/failure modes.

This change does the following:
- makes changes to the base sdk-browser and sdk-server clients to handle non-terminal activity statuses
- update various client libraries that depend on sdk-browser and sdk-server to allow for awaiting consensus
- adds helpers to retrieve signed transactions + signed messages once consensus has been met on the Turnkey side (i.e. in case of a timeout)
- `@turnkey/viem` now returns viem `BaseError`s that wrap around `TurnkeyActivityError`s. This also implicitly addresses https://github.com/tkhq/sdk/issues/121
- includes new example scripts for the ethers, viem, and solana packages (e.g. `retry.ts` and `consensus.ts`) which exercise this new functionality

The end user experience is such that if there is a policy enforcing consensus, you can configure a poller to await consensus to be met. See screen recordings in the testing section below.

TODO:
- add unit tests (ideally — this requires policies)

## How I Tested These Changes
local examples

Live polling for consensus during Ethers signing scripts:

https://github.com/user-attachments/assets/d788d721-4016-4d1e-a0f9-0e3ae0d5233f

Consensus is not met at first within the polling retry limit, then later met, and the signature is retroactively retrieved + added to a Solana transaction
https://github.com/user-attachments/assets/af7fb6e3-46df-4785-a8ec-e2e7d4d1ff7e

## Did you add a changeset?


packages/ethers/src/index.ts
```diff
@@ -128,13 +128,21 @@ export class TurnkeySigner extends AbstractSigner implements ethers.Signer {
...
-      const activity = await this.client.signTransaction({
+      const { activity, signedTransaction } = await this.client.signTransaction({
         signWith: this.signWith,
         type: "TRANSACTION_TYPE_ETHEREUM",
-        unsignedTransaction: message,
+        unsignedTransaction,
       });
 
-      return assertNonNull(activity?.signedTransaction);
+      if (activity.status !== "ACTIVITY_STATUS_COMPLETED") {
```
zkharit, line 137: What exactly does this look like in the case where an activity requires consensus? _Should_ an error be thrown?

andrewkmin: Great question — ended up having the default behavior be to poll for 3 seconds, then return an error. At that point, it's up to the end-user to figure out how they want to handle the error. I've included some examples of handling in the various `retry.ts` files

examples/with-ethers/src/consensus.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  const turnkeyClient = new TurnkeyServerSDK({
+    apiBaseUrl: process.env.BASE_URL!,
+    apiPrivateKey: process.env.API_PRIVATE_KEY!,
+    apiPublicKey: process.env.API_PUBLIC_KEY!,
+    defaultOrganizationId: process.env.ORGANIZATION_ID!,
+    // The following config is useful in contexts where an activity requires consensus.
+    // By default, if the activity is not initially successful, it will poll a maximum
+    // of 3 times with an interval of 1000 milliseconds.
```
r-n-o, line 27: Nit: 10k milliseconds according to line 32?

examples/with-ethers/src/consensus.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  print("Network:", `${network} (chain ID ${chainId})`);
+  print("Address:", address);
+  print("Balance:", `${ethers.formatEther(balance)} Ether`);
+  print("Transaction count:", `${transactionCount}`);
+
+  // 1. Sign a raw payload (`eth_sign` style)
+  const message = "Hello Turnkey";
+  const signature = await connectedSigner.signMessage(message);
+  const recoveredAddress = ethers.verifyMessage(message, signature);
```
r-n-o, line 66: What makes this example "consensus.ts"? I think consensus vs. not will be driven by the organization setting? (1 vs N root quorum users?)

(basically, it doesn't seem necessary to have separate consensus examples unless I'm missing something 🤔 )

packages/ethers/src/index.ts
```diff
@@ -128,13 +134,23 @@ export class TurnkeySigner extends AbstractSigner implements ethers.Signer {
...
+      );
 
-      return assertNonNull(activity?.signedTransaction);
+      if (activity.status !== "ACTIVITY_STATUS_COMPLETED") {
+        throw new TurnkeyActivityError({
+          message: `Unexpected activity status: ${activity.status}`,
+          activityId: activity.id,
+          activityStatus: activity.status as TActivityStatus,
+        });
+      }
```
r-n-o, line 151: I think the key is to throw a _different_ error when the activity is in need of consensus. Something like:
```ts
if (activity.status == "ACTIVITY_STATUS_NEEDS_CONSENSUS") {
    throw new TurnkeyActivityNeedsConsensus({
          message: `Activity requires consensus`,
          activityId: activity.id,
          activityStatus: activity.status as TActivityStatus,
        });
}
```

Then I agree with @andrewkmin, the retry or error behavior can be up to the caller.

examples/with-ethers/src/consensus.ts
andrewkmin: good call! not that useful. `retry.ts` should be enough to serve the purpose of demonstrating the scenario where one might be waiting for consensus

packages/ethers/src/index.ts
andrewkmin: ended up adding some bespoke errors to viem. the thing is it needs `.walk()` so I ended up extending the `BaseError` class. let me know what you think — if it adds unnecessary complexity, happy to cut it or reduce it down!

examples/with-ethers/src/index.ts
```diff
@@ -25,6 +33,16 @@ async function main() {
...
+    // The following config is useful in contexts where an activity requires consensus.
+    // By default, if the activity is not initially successful, it will poll a maximum
+    // of 3 times with an interval of 10000 milliseconds.
+    //
+    // -----
+    //
+    // activityPoller: {
+    //   intervalMs: 10_000,
+    //   numRetries: 5,
+    // },
```
r-n-o, line 45: Is this still relevant?

packages/ethers/src/index.ts
```diff
@@ -240,35 +254,34 @@ export class TurnkeySigner extends AbstractSigner implements ethers.Signer {
...
-      }
+      checkActivityStatus({
+        id: activity.id,
+        status: activity.status,
+      });
 
       result = assertNonNull(activity?.result?.signRawPayloadResult);
     } else {
-      result = await this.client.signRawPayload({
+      // const trying = this.client as TurnkeyClient;
```
r-n-o, line 264: this comment was probably not meant to be a permanent comment? :)

packages/http/src/shared.ts
```diff
@@ -5,31 +5,187 @@ export type TActivityResponse = definitions["v1ActivityResponse"];
...
+
+    this.name = "TurnkeyActivityConsensusNeededError";
+    this.activityId = activityId ?? undefined;
+    this.activityStatus = activityStatus ?? undefined;
+    this.activityType = activityType ?? undefined;
+    this.cause = cause ?? undefined;
+  }
+}
+
+export function checkActivityStatus(input: {
```
r-n-o, line 60: `checkActivityStatus` implies it's a stateless innocuous function but I think this is more heavy than it looks: it will throw errors! So maybe `assertActivityCompleted` since that's kind of what it boils down to?

packages/http/src/shared.ts
```diff
@@ -5,31 +5,187 @@ export type TActivityResponse = definitions["v1ActivityResponse"];
...
+export type TSignature = definitions["v1SignRawPayloadResult"];
+export type TSignedTransaction =
+  definitions["v1SignTransactionResult"]["signedTransaction"];
 
 export class TurnkeyActivityError extends Error {
-  activityId: TActivityId | null;
-  activityStatus: TActivityStatus | null;
-  activityType: TActivityType | null;
-  cause: Error | null;
+  activityId: TActivityId | undefined;
```
r-n-o, line 13: Why the change from `null` to `undefined`?

packages/http/src/shared.ts
```diff
@@ -5,31 +5,187 @@ export type TActivityResponse = definitions["v1ActivityResponse"];
...
+ */
+export function getSignatureFromActivity(activity: TActivity): TSignature {
+  if (
+    ![
+      "ACTIVITY_TYPE_SIGN_RAW_PAYLOAD",
+      "ACTIVITY_TYPE_SIGN_RAW_PAYLOAD_V2",
+    ].includes(activity.type)
+  ) {
+    throw new TurnkeyActivityError({
+      message: `Unexpected activity type: ${activity.type}`,
```
r-n-o, line 118: the message could be "Cannot get signature from activity type ...etc...". Otherwise it's a bit unclear why it's "unexpected".

packages/http/src/shared.ts
```diff
@@ -5,31 +5,187 @@ export type TActivityResponse = definitions["v1ActivityResponse"];
...
+ * @return signature {r, s, v}
+ */
+export function getSignatureFromActivity(activity: TActivity): TSignature {
+  if (
+    ![
+      "ACTIVITY_TYPE_SIGN_RAW_PAYLOAD",
+      "ACTIVITY_TYPE_SIGN_RAW_PAYLOAD_V2",
+    ].includes(activity.type)
+  ) {
+    throw new TurnkeyActivityError({
```
r-n-o, line 117: I'm also thinking this should not be called "TurnkeyActivityError". It's likely an SDK-level / user-level issue if this function is called on the wrong activity type. We might need a plain `BadArgumentError`? Otherwise users will be led to believe the issue is within the turnkey activity processing when they get this error.

packages/http/src/shared.ts
```diff
@@ -5,31 +5,187 @@ export type TActivityResponse = definitions["v1ActivityResponse"];
...
+/**
+ * This function is a helper method to easily extract signature strings from a completed signing activity.
+ * Particularly useful for scenarios where a signature requires consensus
+ *
+ * @param activity the signing activity
+ * @return signatures {r, s, v}[]
+ */
+export function getSignaturesFromActivity(activity: TActivity): TSignature[] {
+  if (!["ACTIVITY_TYPE_SIGN_RAW_PAYLOADS"].includes(activity.type)) {
+    throw new TurnkeyActivityError({
```
r-n-o, line 143: likewise here `UnexpectedInput` or `BadArgument` would capture more closely the type of error here.

packages/sdk-browser/scripts/codegen.js
```diff
@@ -307,43 +309,52 @@ export class TurnkeySDKClientBase {
...
+
+    const pollStatus = async (activityId: string): Promise<TResponseType> => {
+      const pollBody = { activityId };
+      const pollData = await this.getActivity(pollBody) as ActivityResponse;
+
+      if (attempts > maxRetries) {
+        return handleResponse(pollData);
+      }
+
+      attempts += 1;
```
r-n-o, line 341: Nice. Finally fixing the "retry forever" behavior! 🏆

packages/sdk-browser/src/__types__/base.ts
```diff
@@ -44,26 +53,23 @@ export class TurnkeyRequestError extends Error {
...
 }
 
 interface BaseSDKClientConfig {
   apiBaseUrl: string;
   organizationId: string;
-  activityPoller?: {
-    duration: number;
-    timeout: number;
-  };
+  activityPoller?: TActivityPollerConfig | undefined;
```
r-n-o, line 72: Should this be considered a breaking change or nah?

packages/viem/src/index.ts
```diff
@@ -10,11 +10,69 @@ import type {
...
+  TurnkeyClient,
+} from "@turnkey/http";
 import { ApiKeyStamper } from "@turnkey/api-key-stamper";
 import type { TurnkeyBrowserClient } from "@turnkey/sdk-browser";
 import type { TurnkeyServerClient } from "@turnkey/sdk-server";
 
+export type TTurnkeyConsensusNeededErrorType = TurnkeyConsensusNeededError & {
+  name: "TurnkeyConsensusNeededError";
+};
+export class TurnkeyConsensusNeededError extends BaseError {
```
r-n-o, line 30: Instead of defining these errors here, can we pull that into `http`? Is there a way to comply with viem's error interface (implement `walk` is one thing; maybe others?) without importing viem's BaseError directly?

We'd need to implement this interface in our http/shared errors: https://github.com/wevm/viem/blob/408948bb74685e9d562cd39f50768f9b282463fe/src/errors/base.ts#L37-L83

I don't think it'd hurt anything to have our errors implement more methods than they need to, but maybe not something we need right now; feels a bit awkward to be pushed by a particular framework to implement these methods for all Turnkey errors.

packages/http/src/shared.ts
andrewkmin: def more accurate1

examples/with-ethers/src/index.ts
andrewkmin: Yep! a bit ambiguous with the numbers though; will change them such that it might be a bit less confusing

packages/http/src/shared.ts
andrewkmin: I went into this because of type compatibility, and ended up going with undefined for consistency. (functionally, `null == undefined = true`). Specifically, I feel like `param: <Type X> | undefined` better conveys the idea that the absence of the param is okay

packages/http/src/shared.ts
andrewkmin: good call

andrewkmin: ended up going with `InvalidArgumentError`

packages/sdk-browser/src/__types__/base.ts
andrewkmin: It could be, but at the same time don't feel like it quite justifies a major version bump. I've explicitly called it out in the changeset!

packages/viem/src/index.ts
andrewkmin: +1 on the sentiment. I'm down to keep these encapsulated to just Viem for now

packages/sdk-browser/src/__types__/base.ts
r-n-o: Cool! I'm good with that 👍

packages/http/src/shared.ts
r-n-o: Is `activityId: TActivityId | undefined;` equivalent to `activityId?: TActivityId;` then?

r-n-o: If that's the case I'm a definite +1 on the change; simpler to reason about!

packages/http/src/shared.ts
```diff
@@ -5,31 +5,211 @@ export type TActivityResponse = definitions["v1ActivityResponse"];
...
+
+    this.name = "InvalidArgumentError";
+    this.activityId = activityId ?? undefined;
+    this.activityStatus = activityStatus ?? undefined;
+    this.activityType = activityType ?? undefined;
+    this.cause = cause ?? undefined;
+  }
+}
+
+export function assertActivityCompleted(input: {
```
r-n-o, line 84: Now that I'm seeing this with fresh eyes: the activity ID isn't used anywhere really, we're simply comparing the passed in `status` to `CONSENSUS_NEEDED` and `COMPLETED`. Don't think it's necessarily worth changing this but...a bit of an odd interface given nothing guarantees the passed in status come from the passed in activity ID :/

Maybe it'd be clearer if a `TActivity` was expected and passed in instead?

r-n-o (approved):
🚀

packages/http/src/shared.ts
andrewkmin: Attempting this here: https://github.com/tkhq/sdk/pull/350/commits/aa16d4d5c4e130a8ede3946eca3ec13868ed437d

The drawback of this approach is that it's a bit duplicative -- e.g. we return something like (paraphrasing)

{
  ...
  result: activityData["result"]["activityResult"],
  activity: activityData
  ...
}

So you see the overlap there, but having the `result` up front and center just makes access easier. The reason we needed this in the first place was originally for type-compatibility, but I realized it generally could be helpful to return all activity details

r-n-o: yeah it's definitely not super clear-cut, but I like this new version: `assertActivityCompleted(activity);` reads suuuper nice 😍

r-n-o (approved):

PR #351, opened 2024-09-05
clean up --> kitchen sink

andrewkmin:
## Summary & Motivation
$title

Create a folder called `kitchen-sink` to serve as a catch-all for various ways you can perform operations using Turnkey's various libraries. Ideally, instead of fumbling around for tkcli commands to use, you can come and use the kitchen-sink folder directly for one-off requests you'd like to make

This also performs a lot of cleanup in updating existing examples to move from using a combination of `@turnkey/http + @turnkey/api-key-stamper` to using the newer `@turnkey/sdk-server`

Where relevant, we also swap out usage of private keys in favor of wallets

## How I Tested These Changes
- build + typecheck

## Did you add a changeset?


examples/deployer/src/createNewWallet.ts
```diff
@@ -1,36 +1,21 @@
-import {
-  TurnkeyClient,
-  createActivityPoller,
-  TurnkeyActivityError,
-} from "@turnkey/http";
-import { ApiKeyStamper } from "@turnkey/api-key-stamper";
+import { Turnkey as TurnkeySDKServer } from "@turnkey/sdk-server";
```
moe-dev, line 1: Much cleaner 😍 - I think we should continue leaving the lower level http functions for our authentication examples (email auth, otp auth, oauth) for now though (which is what was done here 🤝)

examples/deployer/src/createNewWallet.ts
```diff
@@ -55,18 +37,10 @@ export async function createNewWallet() {
         `- Wallet ID: ${walletId}`,
         `- Address: ${address}`,
         ``,
-        "Now you can take the address, put it in `.env.local` (`SIGN_WITH=<address>`), then re-run the script.",
+        "Now you can take the address, put it in `.env.local`, then re-run the script.",
```
moe-dev, line 40: nit: was this intentional?

examples/kitchen-sink/README.md
```diff
@@ -0,0 +1,7 @@
+# Kitchen Sink
```
moe-dev, line 1: Loving this so much. Have we discussed alternative names for this example? Not sure if everyone will automatically associated kitchen-sink with an sdk playground.

examples/rebalancer/package.json
```diff
@@ -9,9 +9,7 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
-    "@turnkey/api-key-stamper": "workspace:^",
```
moe-dev, line 12: 🧹

examples/sweeper/src/provider.ts
```diff
@@ -1,17 +1,18 @@
 import * as path from "path";
 import * as dotenv from "dotenv";
 import { ethers } from "ethers";
+
+// Load environment variables from `.env.local`
+dotenv.config({ path: path.resolve(process.cwd(), ".env.local") });
```
moe-dev, line 6: nit: any reason this was moved up here instead of after all imports or after * imports?

examples/with-nonce-manager/src/createNewWallet.ts
```diff
@@ -55,7 +40,7 @@ export async function createNewWallet() {
         `- Wallet ID: ${walletId}`,
         `- Address: ${address}`,
         ``,
-        "Now you can take the address, put it in `.env.local` (`SIGN_WITH=<address>`), then re-run the script.",
+        "Now you can take the address, put it in `.env.local`, then re-run the script.",
```
moe-dev, line 43: Same as the first file I commented on - this seems intentional and consistent - all good then 👍

moe-dev (approved):
nits: nothing blocking - haven't tested locally but under the assumption everything is working as expected, code LGTM! 🥇 

PS: also needs prettier

examples/sweeper/src/provider.ts
andrewkmin: nope no particular reason... I guess the mentality was: external imports first, then env var stuff, then turnkey libraries, and then local imports (e.g. util files). just a matter of preference, nothing else

examples/deployer/src/createNewWallet.ts
andrewkmin: gut feeling is we should start to migrate everything to the new libraries, but maybe it'd be useful to leave one or two using the "legacy" suite of libraries

examples/deployer/src/createNewWallet.ts
andrewkmin: nice catch!! 🔎  will revert

examples/with-nonce-manager/src/createNewWallet.ts
andrewkmin: nice catch here as well, will put it back

PR #362, opened 2024-09-11
Exposes the storage APIs for the sdk-browser

taylorjdawson:
## Summary & Motivation
There are times where a developers may want to manage the local storage of the users, sessions and auth bundles themselves. When implementing the demo-embeded-wallet I found that I needed to store the current session after creating a new sub organization. Without this, you need to call one of the `login` functions such as `passkeyClient.login()` in order to save the session to local storage. However, this would require an additional passkey tap.

This would allow for creating a session for a new sub org using the parent org and then saving the session in local storage.

## Did you add a changeset?

### DIFF
--- a/.changeset/popular-rats-dance.md
+++ b/.changeset/popular-rats-dance.md
@@ -1,0 +1,5 @@
+---
+"@turnkey/sdk-browser": patch
+---
+
+Exposes storage APIs used by the sdk for managing users & sessions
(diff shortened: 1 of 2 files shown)


andrewkmin (approved):

andrewkmin (approved):
👏

.changeset/popular-rats-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+"@turnkey/sdk-browser": patch
```
r-n-o, line 2: nit: in semver this should be a minor change (new feature), not a patch (which would be a "bug fix")

r-n-o (approved):

PR #372, opened 2024-09-18
Moe: Importing in node

moe-dev:
## Summary & Motivation
Will require @turnkey/crypto major release (hpkeEncrypt is a breaking change)
Allows clients to programmatically import private keys and wallets in a node environment 

New requirements:
"@peculiar/webcrypto": "^1.5.0", 

For node crypto polyfill

## Did you add a changeset?


packages/crypto/src/crypto.ts
```diff
@@ -466,3 +493,183 @@ const bigIntToHex = (num: bigint, length: number): string => {
...
   return hexString.padStart(length, "0");
 };
+
+/**
+ * Verifies a signature from an enclave using ECDSA and SHA-256.
+ *
+ * @param {string} enclaveQuorumPublic - The public key of the enclave signer.
+ * @param {string} publicSignature - The ECDSA signature in DER format.
+ * @param {string} signedData - The data that was signed.
+ * @param {Environemnt} environemnt - An enum PROD or PREPROD to verify against the correct signer enclave key
```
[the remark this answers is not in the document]
andrewkmin: let's put these super Turnkey-specific functions in a separate file? pretty much all of these new additions

examples/import-in-node/src/index.ts
```diff
@@ -0,0 +1,119 @@
...
+  const targetKeyBuf = uint8ArrayFromHexString(signedData.targetPublic);
+  if (importType == "wallet") {
+    const mnemonic = await input({
+      message: "Enter mneomnic seed phrase for wallet to import",
+    });
+    const plainTextBuf = new TextEncoder().encode(mnemonic);
+    const walletBundle = hpkeEncrypt({ plainTextBuf, targetKeyBuf });
+    const walletImportResult = await turnkeyClient.apiClient().importWallet({
+      userId: userId,
+      walletName: `example-wallet-node-${Date.now()}`,
```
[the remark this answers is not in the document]
andrewkmin: these are just some tiny nits; feel free to disregard if you disagree

packages/crypto/src/crypto.ts
```diff
@@ -466,3 +493,183 @@ const bigIntToHex = (num: bigint, length: number): string => {
...
+
+/**
+ * Decodes a private key based on the specified format.
+ *
+ * @param {string} privateKey - The private key to decode.
+ * @param {string} keyFormat - The format of the private key (e.g., "SOLANA", "HEXADECIMAL").
+ * @returns {Uint8Array} - The decoded private key.
+ */
+
+export const decodeKey = (privateKey: string, keyFormat: any) => {
```
andrewkmin, line 651: let's add the return type to the function signature
```suggestion
export const decodeKey = (privateKey: string, keyFormat: any) : Uint8Array => {
```

packages/crypto/src/crypto.ts
```diff
@@ -466,3 +493,183 @@ const bigIntToHex = (num: bigint, length: number): string => {
...
+ * @returns {Promise<boolean>} - Returns true if the signature is valid, otherwise throws an error.
+ */
+
+export const verifyEnclaveSignature = async (
+  enclaveQuorumPublic: string,
+  publicSignature: string,
+  signedData: string,
+  environment?: Environment
+) => {
+  let signer_key =
```
andrewkmin, line 513: in that separate file, let's declare these as consts?

packages/crypto/src/crypto.ts
```diff
@@ -20,7 +20,12 @@ import {
   SUITE_ID_1,
   SUITE_ID_2,
 } from "./constants";
+import bs58 from "bs58";
 
+export enum Environment {
```
andrewkmin, line 25: can move these enums to a separate turnkey-specific file as well

examples/import-in-node/src/index.ts
```diff
@@ -0,0 +1,119 @@
...
+  const organizationId = process.env.ORGANIZATION_ID!;
+  const userId = process.env.USER_ID!;
+  const turnkeyClient = new Turnkey({
+    apiBaseUrl: process.env.BASE_URL!,
+    apiPublicKey: process.env.API_PUBLIC_KEY!,
+    apiPrivateKey: process.env.API_PRIVATE_KEY!,
+    defaultOrganizationId: organizationId,
+  });
+  const importType = await input({
+    message: "Enter Import Type, either wallet or key",
```
andrewkmin, line 26:
```suggestion
    message: `Enter Import Type, either "wallet" or "key"`,
```

examples/import-in-node/src/index.ts
```diff
@@ -0,0 +1,119 @@
...
+  if (importType == "wallet") {
+    initResult = await turnkeyClient.apiClient().initImportWallet({
+      userId,
+    });
+  } else if (importType == "key") {
+    initResult = await turnkeyClient.apiClient().initImportPrivateKey({
+      userId,
+    });
+  } else {
+    throw new Error(`Invalid import type. Please enter wallet or key`);
```
andrewkmin, line 38:
```suggestion
    throw new Error(`Invalid import type. Please enter "wallet" or "key"`);
```

packages/crypto/src/__tests__/crypto-test.ts
```diff
@@ -23,7 +23,7 @@ const mockCredentialBundle =
   "w99a5xV6A75TfoAUkZn869fVyDYvgVsKrawMALZXmrauZd8hEv66EkPU1Z42CUaHESQjcA5bqd8dynTGBMLWB9ewtXWPEVbZvocB4Tw2K1vQVp7uwjf";
 
 describe("HPKE Encryption and Decryption", () => {
-  test("hpkeEncrypt and hpkeDecrypt - end-to-end encryption and decryption", () => {
+  test("hpkeEncrypt and hpkeDecrypt - end-to-end encryption and decryption", async () => {
```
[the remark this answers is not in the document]
moe-dev: Ahh good call, thought I had to make this async so changed this but changed it back without reverting the tess changes

examples/wallet-import-export/src/components/ImportWallet.tsx
```diff
@@ -71,7 +71,6 @@ export function ImportWallet(props: ImportWalletProps) {
       alert("Encrypted bundle is empty.");
       return;
     }
-
```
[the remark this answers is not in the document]
moe-dev: hmm may be prettier...

examples/import-in-node/src/index.ts
```diff
@@ -0,0 +1,116 @@
+import * as dotenv from "dotenv";
+import * as path from "path";
+import { input } from "@inquirer/prompts";
+import { Turnkey } from "@turnkey/sdk-server";
+import { Crypto } from "@peculiar/webcrypto";
```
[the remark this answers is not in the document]
moe-dev: need the polyfill, we do the same thing with our RN example

packages/crypto/src/crypto.ts
```diff
@@ -466,3 +492,142 @@ const bigIntToHex = (num: bigint, length: number): string => {
...
+  signedData: string,
+  environment?: Environment
+) => {
+  const signer_key = SIGNER_PUBLIC_KEY[environment ?? Environment.PROD];
+
+  if (enclaveQuorumPublic != signer_key) {
+    throw new Error(
+      "expected signer key does not match signer key from bundle"
+    );
+  }
```
r-n-o, line 518: I don't think this is checking the bundle key? It's only matching what's arbitrarily passed in (`enclaveQuorumPublic`). Nothing guarantees it comes from the bundle.

I think a better interface would be "verifyBundle"?

packages/crypto/src/crypto.ts
```diff
@@ -466,3 +492,142 @@ const bigIntToHex = (num: bigint, length: number): string => {
...
+      return decodedKeyBytes.subarray(0, 32);
+    case "HEXADECIMAL":
+      if (privateKey.startsWith("0x")) {
+        return uint8ArrayFromHexString(privateKey.slice(2));
+      }
+      return uint8ArrayFromHexString(privateKey);
+    default:
+      console.warn(
+        `invalid key format: ${keyFormat}. Defaulting to HEXADECIMAL.`
+      );
```
r-n-o, line 627: nice!

examples/import-in-node/src/index.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  // Load target public key generated from enclave and set in local storage
+  const targetKeyBuf = uint8ArrayFromHexString(signedData.targetPublic);
+  if (importType == "wallet") {
+    const mnemonic = await input({
+      message: "Enter mnemonic seed phrase for wallet to import",
+    });
+    const plainTextBuf = new TextEncoder().encode(mnemonic);
+    const walletBundle = hpkeEncrypt({ plainTextBuf, targetKeyBuf });
```
r-n-o, line 80: This **whole** section is very very security sensitive so I don't think we should leave that up to users to copy/paste. IMO the interface should be reduced to:
```ts
// This parses the bundles, verifies the signature and organization ID, then calls hpkeEncrypt
const walletBundle = const crypto.ImportWalletWithBundle(plainTextBuf, initResult.importBundle, userId, organizationId)
```

And in the case of private keys:
```ts
// does that same for private keys, including taking care of key format shenanigans
const keyBundle = const crypto.ImportPrivateKeyWithBundle(plainTextBuf, format, initResult.importBundle, userId, organizationId)
```

Basically let's make this as foolproof as possible.

packages/crypto/src/constants.ts
```diff
@@ -21,3 +21,13 @@ export const IV_INFO = new Uint8Array([
...
+export enum Environment {
+  PROD = "PROD",
+  PREPROD = "PREPROD",
+}
+export const SIGNER_PUBLIC_KEY = {
+  [Environment.PROD]:
+    "04cf288fe433cc4e1aa0ce1632feac4ea26bf2f5a09dcfe5a42c398e06898710330f0572882f4dbdf0f5304b8fc8703acd69adca9a4bbf7f5d00d20a5e364b2569",
+  [Environment.PREPROD]:
+    "04f3422b8afbe425d6ece77b8d2469954715a2ff273ab7ac89f1ed70e0a9325eaa1698b4351fd1b23734e65c0b6a86b62dd49d70b37c94606aac402cbd84353212",
+};
```
r-n-o, line 33: I'm a bit nervous about this. I'd rather check in only one public key: our production key. Then let's allow an override through a node env var?

```ts
export const TURNKEY_SIGNER_PUBLIC_KEY = "04cf288fe433cc4e1aa0ce1632feac4ea26bf2f5a09dcfe5a42c398e06898710330f0572882f4dbdf0f5304b8fc8703acd69adca9a4bbf7f5d00d20a5e364b2569;
```

And when it's used we can look at an env var (say, "OVERRIDE_TURNKEY_SIGNER_PUBLIC_KEY"). That way it's very explicit that the public key should always be our production key, and if we are overriding this key it's up to us individuals, for testing purposes.

packages/crypto/src/crypto.ts
```diff
@@ -466,3 +492,142 @@ const bigIntToHex = (num: bigint, length: number): string => {
...
+  return await crypto.subtle.verify(
+    { name: "ECDSA", hash: { name: "SHA-256" } },
+    quorumKey,
+    publicSignatureBuf,
+    signedDataBuf
+  );
+};
+
+/**
+ * Converts an ASN.1 DER-encoded ECDSA signature to the raw format that WebCrypto uses.
```
r-n-o, line 540: wow that's wild that webcrypto doesn't deal with DER-encoded signatures! dang.

packages/crypto/src/crypto.ts
```diff
@@ -66,16 +70,28 @@ export const hpkeEncrypt = ({
...
   targetKeyBuf,
   senderPriv,
-}: HpkeEncryptParams): Uint8Array => {
+}: HpkeEncryptParams): string => {
   try {
-    const senderPubBuf = getPublicKey(
-      uint8ArrayFromHexString(senderPriv),
-      false
-    );
-    const aad = buildAdditionalAssociatedData(senderPubBuf, targetKeyBuf); // Eventually we want users to be able to pass in aad as optional
```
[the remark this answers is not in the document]
moe-dev: Yep! we can say bye bye to this. Not relevant within the context of turnkey

examples/import-in-node/src/index.ts
```diff
@@ -0,0 +1,92 @@
+import * as dotenv from "dotenv";
+import * as path from "path";
+import { input } from "@inquirer/prompts";
+import { Turnkey } from "@turnkey/sdk-server";
+import { Crypto } from "@peculiar/webcrypto";
+import { encryptPrivateKeyBundle, encryptWalletBundle } from "@turnkey/crypto";
```
r-n-o, line 6: `encryptPrivateKeyBundle` and `encryptWalletBundle` suggest that we have a private key or wallet bundle, and that we are encrypting them. This is not what's happening. We are actually using a bundle to encrypt a piece of data. So I think the "with" or "via" or "to" in the name is needed (`encryptPrivateKeyWithBundle` or `encryptPrivateKeyViaBundle` or `encryptPrivateKeyToBundle`)

packages/crypto/src/crypto.ts
```diff
@@ -466,3 +568,251 @@ const bigIntToHex = (num: bigint, length: number): string => {
...
+      return uint8ArrayFromHexString(privateKey);
+  }
+};
+
+/**
+ * Encrypts a private key bundle using HPKE and verifies the enclave signature.
+ *
+ * @param {string} privateKey - The private key in hexadecimal format to be encrypted.
+ * @param {string} keyFormat - The format of the private key (e.g., "SOLANA", "HEXADECIMAL").
+ * @param {any} importBundle - The bundle containing the enclave's public key and data signature for verification.
```
r-n-o, line 713: this should be a string. This function should JSON.parse the bundle and error out if it's an invalid bundle after asserting that the right properties are defined on it. That will let you avoid the "any" in here, and it'll also let you be more confident about what's in the bundle (you can early throw if there are non-empty keys for data, signature, or quorum key)

packages/crypto/src/crypto.ts
```diff
@@ -466,3 +568,251 @@ const bigIntToHex = (num: bigint, length: number): string => {
...
+ * @param {string} userId - The user ID that must match the signed data in the bundle.
+ * @param {string} organizationId - The organization ID that must match the signed data in the bundle.
+ * @returns {Promise<string>} - A promise that resolves to a JSON string representing the encrypted wallet bundle.
+ * @throws {Error} - If enclave signature verification or any other validation fails.
+ */
+
+export const encryptWalletBundle = async (
+  mnemonic: string,
+  importBundle: any,
```
r-n-o, line 778: same here. Let's have a string and document that it should be a JSON encoded string received from the activity result.

Users should just pass `initResult.bundle` straight to `encryptWalletWithBundle`; max foolproofness.

packages/crypto/src/crypto.ts
```diff
@@ -466,3 +568,251 @@ const bigIntToHex = (num: bigint, length: number): string => {
...
+
+/**
+ * Decodes a private key based on the specified format.
+ *
+ * @param {string} privateKey - The private key to decode.
+ * @param {string} keyFormat - The format of the private key (e.g., "SOLANA", "HEXADECIMAL").
+ * @returns {Uint8Array} - The decoded private key.
+ */
+
+export const decodeKey = (privateKey: string, keyFormat: any): Uint8Array => {
```
r-n-o, line 682: should we export these functions if they're not gonna be used externally? I think it's safer to export the minimal surface first. If we get feedback that these internal functions would be helpful we can certainly export them. But I'd rather err on the side of being conservative with what we export / promise as far as package interface goes.

r-n-o (commented):
Example reads so much better now 😍

packages/crypto/src/crypto.ts
moe-dev: Agreed!

packages/crypto/src/constants.ts
```diff
@@ -21,3 +21,5 @@ export const IV_INFO = new Uint8Array([
   41, 180, 122, 198, 166, 48, 185, 46, 196, 207, 125, 35, 69, 8, 208, 175, 151,
   113, 201, 158, 80,
 ]); //base_nonce
+export const SIGNER_PUBLIC_KEY =
```
r-n-o, line 24: Nit: let's be explicit about the fact that we're talking about Turnkey's signer, and in prod? `TURNKEY_PRODUCTION_SIGNER_ENCLAVE_PUBLIC_KEY` would be the explicit const name, but I'd settle for `PRODUCTION_SIGNER_PUBLIC_KEY` given it's in the Turnkey SDK already, and used in "verify enclave signature" functionality.

packages/crypto/src/crypto.ts
```diff
@@ -119,6 +131,96 @@ export const hpkeEncrypt = ({
...
+
+/**
+ * Format HPKE Buffer Function
+ * Returns a JSON string of an encrypted bundle, separating out the cipher text and the sender public key
+ *
+ * @param {Uint8Array} encryptedBuf - The result of hpkeAuthEncrypt or hpkeEncrypt
+ * @returns {string} - A JSON string with "encappedPublic" and "ciphertext"
+ */
+
+export const formatHpkeBuf = (encryptedBuf: Uint8Array): string => {
```
r-n-o, line 209: does this need to be exported or can we get away without it?

packages/crypto/src/crypto.ts
```diff
@@ -466,3 +568,253 @@ const bigIntToHex = (num: bigint, length: number): string => {
...
+ * @param {Environemnt} environment - An enum PROD or PREPROD to verify against the correct signer enclave key
+ * @returns {Promise<boolean>} - Returns true if the signature is valid, otherwise throws an error.
+ */
+
+const verifyEnclaveSignature = async (
+  enclaveQuorumPublic: string,
+  publicSignature: string,
+  signedData: string
+) => {
+  if (enclaveQuorumPublic != SIGNER_PUBLIC_KEY) {
```
r-n-o, line 587: This would make it impossible to import into preprod. We might need a way to override this? extra `signerEnclaveOverride?:` param?

packages/crypto/src/crypto.ts
```diff
@@ -466,3 +568,253 @@ const bigIntToHex = (num: bigint, length: number): string => {
...
+    throw new Error(
+      "expected signer key does not match signer key from bundle"
```
r-n-o, line 589: Nit: could be nice to template both values ("expected signer key (VALUE) does not match bundle public key (VALUE)", something like that)

packages/crypto/src/crypto.ts
```diff
@@ -65,17 +89,22 @@ export const getPublicKey = (
...
   plainTextBuf,
   targetKeyBuf,
-  senderPriv,
 }: HpkeEncryptParams): Uint8Array => {
   try {
-    const senderPubBuf = getPublicKey(
-      uint8ArrayFromHexString(senderPriv),
-      false
+    let senderPrivBuf: Uint8Array | null = null;
+    let senderPubBuf: Uint8Array;
```
r-n-o, line 95: Are these types needed/correct? They seem a bit strange.

packages/crypto/src/crypto.ts
```diff
@@ -466,3 +585,255 @@ const bigIntToHex = (num: bigint, length: number): string => {
...
+  if (
+    enclaveQuorumPublic != PRODUCTION_SIGNER_PUBLIC_KEY &&
+    enclaveQuorumPublic != dangerouslyOverrideSignerPublicKey
+  ) {
+    throw new Error(
+      `expected signer key ${
+        dangerouslyOverrideSignerPublicKey ?? PRODUCTION_SIGNER_PUBLIC_KEY
+      } does not match signer key from bundle: ${enclaveQuorumPublic}`
+    );
+  }
```
r-n-o, line 614: I think you can consolidate:
```js
const expectedSignerPublicKey = dangerouslyOverrideSignerPublicKey || PRODUCTION_SIGNER_PUBLIC_KEY;
if (enclaveQuorumPublic != expectedSignerPublicKey) {
    throw new Error(....)
}
```

r-n-o (approved):
2 final, minor things. Otherwise we're good to go! Ship it! 🚀

PR #383, opened 2024-10-07
add example using biconomy paymasters (account abstraction)

andrewkmin:
## Summary & Motivation
$title

## How I Tested These Changes
local testing

## Did you add a changeset?


examples/with-biconomy-aa/src/viem.ts
```diff
@@ -0,0 +1,176 @@
...
+  // Initialize a Turnkey-powered Viem Account
+  const turnkeyAccount = await createAccount({
+    client: turnkeyClient.apiClient(),
+    organizationId: process.env.ORGANIZATION_ID!,
+    signWith: process.env.SIGN_WITH!,
+  });
+
+  const network = "sepolia";
+
+  // Bring your own provider (such as Alchemy or Infura: https://docs.ethers.org/v6/api/providers/)
```
moe-dev, line 50: Well we kinda force them to use infura from our env, so perhaps we should just allow them to pass in a provider URL? But this is consistent with other examples - just noting. No action needed

moe-dev (approved):
Beauty

examples/with-biconomy-aa/src/viem.ts
andrewkmin: true! I'll reduce this comment for now

moe-dev (approved):

PR #389, opened 2024-10-11
Solana with passkeys - Fix ENV Var Name

omkarshanbhag:
This makes it so the rest of the code within the example uses `NEXT_PUBLIC_TURNKEY_API_BASE_URL` instead of `NEXT_PUBLIC_BASE_URL`  --> the readme and example local env use the former (which is more descriptive) and the code uses the latter causing some confusion

## Did you add a changeset?

### DIFF
--- a/examples/with-solana-passkeys/src/pages/_app.tsx
+++ b/examples/with-solana-passkeys/src/pages/_app.tsx
@@ -4,7 +4,7 @@ import Head from "next/head";
 import { TurnkeyProvider } from "@turnkey/sdk-react";
 
 const turnkeyConfig = {
-  apiBaseUrl: process.env.NEXT_PUBLIC_BASE_URL!,
+  apiBaseUrl: process.env.NEXT_PUBLIC_TURNKEY_API_BASE_URL!,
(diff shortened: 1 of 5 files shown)


r-n-o (approved):

examples/with-solana-passkeys/src/pages/_app.tsx
```diff
@@ -4,7 +4,7 @@ import Head from "next/head";
 import { TurnkeyProvider } from "@turnkey/sdk-react";
 
 const turnkeyConfig = {
-  apiBaseUrl: process.env.NEXT_PUBLIC_BASE_URL!,
+  apiBaseUrl: process.env.NEXT_PUBLIC_TURNKEY_API_BASE_URL!,
```
andrewkmin, line 7: Great catch! I think in most env var files (e.g. `.env.local.example`) we call this `NEXT_PUBLIC_BASE_URL`, so it might be easier to swap the name in `.env.local.example` for consistency across the repo?

andrewkmin (approved):
left a note for a minor revision, but giving a stamp for now so you can merge afterwards 👍

PR #391, opened 2024-10-13
fix for type issues viem

moe-dev:
## Did you add a changeset?

### DIFF
--- a/packages/http/src/__generated__/services/coordinator/public/v1/public_api.client.ts
+++ b/packages/http/src/__generated__/services/coordinator/public/v1/public_api.client.ts
@@ -286,12 +286,13 @@ import type {
 export class TurnkeyClient {
   config: THttpConfig;
   stamper: TStamper;
+  type: string;
 
   constructor(config: THttpConfig, stamper: TStamper) {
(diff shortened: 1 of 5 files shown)


packages/http/src/__generated__/services/coordinator/public/v1/public_api.client.ts
```diff
@@ -286,12 +286,13 @@ import type {
...
   config: THttpConfig;
   stamper: TStamper;
+  type: string;
 
   constructor(config: THttpConfig, stamper: TStamper) {
     if (!config.baseUrl) {
       throw new Error(`Missing base URL. Please verify env vars.`);
     }
-
+    this.type = "http";
```
zkharit, line 295: If this field is going to be a hard-coded string, can we create some const's or an enum for the values?

moe-dev: this enum would live in http package and would only containt one const - In this case I'm not sure it makes sense unless we store all these constants in a separate package (seems a bit overkill for a hacky fix).

zkharit (approved):

PR #394, opened 2024-10-15
Add full coverage of default accounts in our sdk for all address formats

moe-dev:
## Did you add a changeset?

### DIFF
--- a/packages/sdk-browser/src/turnkey-helpers.ts
+++ b/packages/sdk-browser/src/turnkey-helpers.ts
@@ -4,16 +4,59 @@ interface WalletAccount {
   path: string;
   addressFormat:
     | "ADDRESS_FORMAT_ETHEREUM"
-    | "ADDRESS_FORMAT_UNCOMPRESSED"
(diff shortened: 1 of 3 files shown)


packages/sdk-browser/src/turnkey-helpers.ts
```diff
@@ -3,14 +3,63 @@ interface WalletAccount {
...
+    path: `m/44'/118'/${pathIndex}'/0/0`,
+    addressFormat: "ADDRESS_FORMAT_COSMOS",
+  };
+};
+
+export const DEFAULT_COSMOS_ACCOUNTS: WalletAccount[] = [
+  defaultCosmosAccountAtIndex(0),
+];
+
+// Ethereum
```
andrewkmin, line 62: Tiny nit -- can we have ETH at the top here?

andrewkmin (approved):
🔥 thanks for adding these. Can you add these into `sdk-server` as well? 

bonus points down the line if we use these exported values in our examples :)

r-n-o (approved):
💯

PR #399, opened 2024-10-17
Add initial Telegram Stamper package

zkharit:
## Summary & Motivation
Add compatibility for storing Turnkey api keys in Telegram Cloud Storage
## Did you add a changeset?


packages/telegram-stamper/src/index.ts
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// config when telegram stamper is passed in an api key
+export type TTelegramStamperConfig = {
+  apiPublicKey: string;
+  apiPrivateKey: string;
+};
+
+const PUBLIC_KEY_CLOUD_STORAGE_KEY = "TURNKEY_API_PUBLIC_KEY";
+const PRIVATE_KEY_CLOUD_STORAGE_KEY = "TURNKEY_API_PRIVATE_KEY";
```
robin-turnkey, line 18: Pull from env vars?

zkharit: I wanted to keep these constant from application to application was the reason I didn't make these configurable, having said that, im not really sure what the best approach here is

packages/telegram-stamper/src/util/telegram-environment.ts
```diff
@@ -0,0 +1,17 @@
+import JSDOMEnvironment from "jest-environment-jsdom";
```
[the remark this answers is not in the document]
zkharit: Originally I had it there, but jest thought it was a test file and yelled at me for it not having any tests inside of it.

packages/telegram-stamper/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2023 Turnkey
```
r-n-o, line 189: 2024

packages/telegram-stamper/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# @turnkey/telegram-stamper
```
r-n-o, line 1: to be future proof I wonder if we should name this "telegram-cloud-storage-stamper"? I'm thinking we might have a "telegram-biometrics-stamper" down the road.

packages/telegram-stamper/README.md
```diff
@@ -0,0 +1,69 @@
...
+```
+
+Use an exisiting key that has been previously stored in Telegram Cloud Storage
+
+```ts
+import { TelegramStamper } from "@turnkey/telegram-stamper";
+import { TurnkeyClient } from "@turnkey/http";
+
+// create a new telegram stamper
+const stamper = new TelegramStamper();
```
r-n-o, line 50: I think the same can be said here, let's combine init and constructor?

```ts
// creates an API key, or gets the API key from cloudstorage default entry
const stamper = new TelegramStamper(); 

// creates an API key, or gets the API key from "MY_CUSTOM_ENTRY"
const stamper = new TelegramStamper({
    cloudStorageKey: "MY_CUSTOM_ENTRY"
});
```

packages/telegram-stamper/src/__tests__/stamp-test.ts
```diff
@@ -0,0 +1,41 @@
...
+
+test("uses provided signature to make stamp", async function () {
+  const { privateKey, publicKey, pemPublicKey } = await readFixture();
+
+  const stamper = new TelegramStamper({
+    apiPublicKey: publicKey,
+    apiPrivateKey: privateKey,
+  });
+  // ToDo: init fails since there isn't an actual Telegram environment here
+  // await stamper.init()
```
r-n-o, line 14: You can provide a fake TG storage by aliasing `window.Telegram.WebApp.CloudStorage.setItem` to `window.localStorage.setItem` (and likewise for get). That will let your tests pass and actually test the init code you wrote up until the point where the TG API is invoked.

packages/telegram-stamper/src/index.ts
```diff
@@ -0,0 +1,186 @@
...
+// config when telegram stamper is passed in an api key
+export type TTelegramStamperConfig = {
+  apiPublicKey: string;
+  apiPrivateKey: string;
+  cloudStorageKeySuffix?: string;
+};
+
+// Constants for default key names, note these should NOT be used directly, use the getPublicKeyCloudStorageKeyName() functions as they include the user passed key suffix
+const PUBLIC_KEY_CLOUD_STORAGE_KEY = "TURNKEY_API_PUBLIC_KEY";
+const PRIVATE_KEY_CLOUD_STORAGE_KEY = "TURNKEY_API_PRIVATE_KEY";
```
r-n-o, line 20: Might be worth considering a single key that is a JSON value? That way there's no doubt about the public key and the private key matching. `setItem` can either succeed or fail which is good. If we have this split in two entries we need to worry about 2 setItem calls, and rollbacks and inconsistencies.

packages/telegram-stamper/README.md
```diff
@@ -0,0 +1,69 @@
...
+  apiPublicKey: "...",
+  apiPrivateKey: "...",
+});
+
+// initialize the telegram stamper, this stores the api key credentials in Telegram Cloud Storage for the first time
+try {
+  stamper.init();
+} catch (err) {
+  throw new Error(`Failed initializing Telegram Stamper: ${err}`)
+}
```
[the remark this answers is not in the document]
zkharit: Having said that, maybe its not super likely thats how the tgram stamper will be used? Maybe this should just be removed overall? (from the readme)

packages/telegram-stamper/README.md
zkharit: The init is actually an async method, so it needs to be separate from the constructor (i know there is no await in the readme so that was unclear, ive just added it locally)

packages/telegram-stamper/README.md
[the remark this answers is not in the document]
r-n-o: yep yep, I get that. But it could create an API key if nothing's in cloudStorage right? From a dev's POV the construction becomes the same regardless of whether a user has a key or not.

Currently devs will be forced to branch:
```ts
if (hasApiKeyInCloudStorage) {
   const stamper = new TelegramStamper({});
} else {
   const stamper = new TelegramStamper({
     apiPublicKey: "...", // also need to worry about how to generate it
     apiPrivateKey: "...",
   }
}
```

Whereas if we had this behavior (generate key when storage is empty), the init simplifies to:
```ts
const stamper = new TelegramStamper({});
```

I guess this is only important if creating and persisting a **new** key is a common pattern. If the stamper is almost always initiated with an existing key from somewhere else, not worth worrying about.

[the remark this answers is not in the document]
zkharit: To be clear, the tgram stamper does not create any api keys for users, they need to be passed to the stamper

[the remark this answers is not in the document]
zkharit: OOp I wrote that comment before I reloaded and saw your response, but yeah Im not exactly sure it makes sense for the stamper to actually create the api keys, because those api keys will have to be added to the users account somehow meaning that the tgram stamper already needs a valid api key, or something needs to pass the tgram stamper a valid key

packages/telegram-stamper/src/index.ts
zkharit: These are now semi-configurable through the cloudStorageKeySuffix optional parameter in the constructor

packages/telegram-stamper/src/index.ts
zkharit: I changed this so that the priv and public key are stored as a json object in 1 key in cloud storage. Theres a good bit of parsing, error, and type checking logic added as a result of this

packages/telegram-stamper/README.md
r-n-o: ah, I see. So that's a typescript limitation. Why does `const stamper = await new TelegramStamper(...)` not work 😭 

...did some quick Googling: https://stackoverflow.com/questions/35743426/async-constructor-functions-in-typescript...

And yeah, makes sense. The return type must be the type, not `Promise<Type>`. I like the alternative that's suggested though. We could have static method: `const stamper = await TelegramStamper.create(...);`

r-n-o: This isn't really blocking btw; just food for thought for interfaces!

packages/telegram-stamper/README.md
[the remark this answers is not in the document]
r-n-o: Cool! I'm okay with that. Let's not bundle API key creation in there and always expect public/private key. As you build out the TG mini-app I think patterns will naturally emerge anyway. I suspect you'll need to create a stamper from a bundle (coming from oauth or email auth). That might be a convenience function we want to provide (`TelegramStamper.fromBundle(...)` would decrypt the bundle and store it in cloudstorage in one go)

packages/telegram-stamper/src/index.ts
```diff
@@ -0,0 +1,185 @@
+/// <reference lib="dom" />
+import { ApiKeyStamper } from "@turnkey/api-key-stamper";
+import { TelegramStamperError } from "./errors";
+
+declare global {
+  interface Window {
+    Telegram: any;
```
r-n-o, line 7: can we be more specific than "any" here? Let's bring this package: https://www.npmjs.com/package/telegram-webapps?

It has all the types we need for CloudStorage:
https://github.com/DavisDmitry/telegram-webapps/blob/d935b6aec22d9361982ba54667eb56faf76b474a/src/index.d.ts#L970

packages/telegram-stamper/src/index.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+}
+
+// config when telegram stamper is passed in an api key
+export type TTelegramStamperConfig = {
+  apiPublicKey: string;
+  apiPrivateKey: string;
+  cloudStorageKeySuffix?: string;
```
r-n-o, line 15: let's remove this option and replace it with "cloudStorageKey" to give customers full control over where keys are stored?

packages/telegram-stamper/src/index.ts
```diff
@@ -0,0 +1,185 @@
...
+};
+
+type CloudStorageAPIKey = {
+  apiPublicKey: string;
+  apiPrivateKey: string;
+};
+
+// Constant for default key name, note this should NOT be used directly, use the getCloudStorageKeyName() function as it includes the user passed key suffix
+const TURNKEY_CLOUD_STORAGE_KEY = "TURNKEY_API_KEY";
```
r-n-o, line 24: I don't think you need this comment because this constant isn't exported, and so not usable by default. If we go with the custom key route (and forgo the prefix), this whole prefix logic goes away so the comment can go as well!

packages/telegram-stamper/src/index.ts
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * Stamper to use with `@turnkey/http`'s `TurnkeyClient`
+ */
+export default class TelegramStamper {
+  stamper: ApiKeyStamper | undefined;
+  cloudStorageKeySuffix: string = "";
```
r-n-o, line 31: The fact that a stamper contains another stamper might be a bit surprising for some readers. I wonder if that's worth a comment in the top level docstring?

We could call this `TelegramCloudStorageStamperContainer` if we were writing enterprise Java...but we're not. So I'm down to keep `TelegramStamper` or `CloudStorageStamper` (slight preference because it's more specific and describes the implementation better?)

packages/telegram-stamper/src/index.ts
```diff
@@ -0,0 +1,185 @@
...
+    );
+  }
+
+  checkTelegramContext() {
+    if (window?.Telegram?.WebApp == null) {
+      throw new TelegramStamperError(
+        "Cannot use telegram stamper in non telegram mini-app environment"
+      );
+    }
+  }
```
r-n-o, line 144: Instead of checking for `WebApp` let's check for what we need? `Telegram.WebApp.CloudStorage`. I don't know if there's ever a case where `WebApp` would be defined while `WebApp.CloudStorage` isn't, but if there is we should error out right away (could be the case if users have the option to disable cloudstorage someday?)

packages/telegram-stamper/src/index.ts
```diff
@@ -0,0 +1,185 @@
...
+          );
+        }
+      }
+    );
+  }
+
+  checkTelegramContext() {
+    if (window?.Telegram?.WebApp == null) {
+      throw new TelegramStamperError(
+        "Cannot use telegram stamper in non telegram mini-app environment"
```
r-n-o, line 141: This message should be more specific than this and state that `Telegram.WebApp` is not defined in the current environment. Otherwise users will have to wonder how we determined that the code was run "in a non telegram mini-app environment"

packages/telegram-stamper/src/index.ts
```diff
@@ -0,0 +1,185 @@
...
+        };
+      }
+
+      return {
+        parsedPublicKey: parsedApiKey.apiPublicKey,
+        parsedPrivateKey: parsedApiKey.apiPrivateKey,
+      };
+    } catch (err) {
+      throw new TelegramStamperError(
+        "Failed parsing api key from Telegram Cloud Storage"
```
r-n-o, line 174: nit: API (capitalize)

packages/telegram-stamper/src/index.ts
```diff
@@ -0,0 +1,185 @@
...
+  async stamp(payload: string) {
+    // check to see if were in a telegram mini app context
+    this.checkTelegramContext();
+
+    // check to see that the stamper was initialized
+    if (!this.stamper) {
+      throw new TelegramStamperError(
+        "Cannot stamp with unintialized telegram stamper, call TelegramStamper.init()"
+      );
+    }
```
r-n-o, line 116: this could be axed if we had the `.create()` pattern, potentially! It's pretty easy to imagine devs forgetting to call `.init()`. Not the worst thing but kind of annoying (it's one more thing to remember)

packages/telegram-stamper/README.md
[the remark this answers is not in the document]
zkharit: Yeah, I think thatll be a common pattern for sure, that sounds like a nice convenience for sure!

packages/telegram-stamper/README.md
zkharit: Yeah, I used init as we have a similar `.init()` for the [iframestamper](https://github.com/tkhq/sdk/blob/main/packages/iframe-stamper/README.md) , so I wanted to have a similar flow between all of our sdk abstractions :)

packages/telegram-stamper/src/index.ts
zkharit: The comment was more for developers who would be writing code for the Telegram stamper, but tbh youre right having a prefix just feels clunky, giving dev's full control is much better

packages/telegram-stamper/src/index.ts
zkharit: So this isnt an "official" Telegram package, and I've been going back and forth about including it. Telegram unfortunately only officially offers `<script src=telegram-web-app.js />`. Its really behind, it terms of how modern apps are developed, and other developers have created much better abstractions like the one you linked. However now were depending on random devs to keep it up to date rather than using the official Telegram source. Given that do you have a strong opinion here?

packages/telegram-stamper/src/index.ts
zkharit: The reason I choose to check for WebApp rather than CloudStorage is there docs: [Once the script is connected, a window.Telegram.WebApp object will become available with the following fields:...](https://core.telegram.org/bots/webapps#initializing-mini-apps), but youre probably right, lets just check for what we need

packages/telegram-stamper/src/index.ts
zkharit: I see, that makes sense. Ill look into that,

packages/telegram-stamper/README.md
r-n-o: Yeah I hear you. Consistency vs. improvement of the interface. That's a tough one because I don't want to force us to rework all of the other stampers. But at the same time, there will never be a better opportunity to improve this interface. Once it's released it's much harder.

I'll leave this up to you and others to decide on this one; no strong feelings either way!

packages/telegram-stamper/README.md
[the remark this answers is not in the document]
r-n-o: by the way this brings up a tangential point: it's probably best not to merge this PR until the TG mini-app is built: that way you can get a "feel" for the interface you're building with this stamper, hammer out the bug, etc. The iteration loop I'd recommend is:
1. TG stamper PR opened (what you have here)
1. TG mini app repo or branch somewhere which imports the codesandbox version of this package
1. Try your mini app in TG, look for bugs; look at the code of the mini-app, look for awkwardness. Go back to step 1 and edit to solve for bugs and awkwardness! 

and when we're convinced this stamper is a fit for the mini-app: merge PR, publish to NPM, and import the final package in the mini app.

r-n-o:
Don't forget to add this new package to https://github.com/tkhq/sdk/blob/main/.codesandbox/ci.json, this will let you build an importable version of it right from this branch so you can iterate and get feedback quickly for integration into a TG mini-app.

packages/telegram-stamper/README.md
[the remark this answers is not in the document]
zkharit: 🫡

zkharit:
didnt know about that ^ awesome!!

packages/telegram-stamper/src/index.ts
zkharit: Changed to cloudStorageKey

packages/telegram-stamper/src/index.ts
zkharit: changed to check for window.Telegram.WebApp.CloudStorage

packages/telegram-stamper/src/index.ts
zkharit: updated this comment

packages/telegram-stamper/src/index.ts
zkharit: fixed a bunch of these in error messages

packages/telegram-stamper/README.md
zkharit: I think the present is the best time to have change to move to best practices, I changed to a .create() method and made the constructor private. This was a decent sized refactor

packages/telegram-stamper/src/index.ts
zkharit: changed to this pattern :)

packages/telegram-stamper/CHANGELOG.md
zkharit: changed to telegram-cloud-storage-stamper

packages/telegram-stamper/src/index.ts
zkharit: changed to TelegramCloudStorageStamper

packages/telegram-cloud-storage-stamper/src/util/telegram-environment.ts
```diff
@@ -0,0 +1,17 @@
+import JSDOMEnvironment from "jest-environment-jsdom";
+
+class TelegramEnvironment extends JSDOMEnvironment {
+  override async setup() {
+    await super.setup();
+
+    this.global.window.Telegram = {
+      WebApp: {},
+    };
```
r-n-o, line 9: I think you might need "CloudStorage" in there as well now that the feature detection code changed.

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,224 @@
...
+    TelegramCloudStorageStamper.checkTelegramContext();
+
+    let cloudStorageKey = DEFAULT_TURNKEY_CLOUD_STORAGE_KEY;
+    let apiPublicKey = "";
+    let apiPrivateKey = "";
+
+    // set the api pub and priv key if config is passed in
+    if (config) {
+      apiPublicKey = config.apiPublicKey;
+      apiPrivateKey = config.apiPrivateKey;
```
r-n-o, line 62: if the API public / private key is null/undefined/empty we should probably throw here? Otherwise we're gonna fail with "TelegramCloudStorageStamperError(`Failed storing api key in Telegram Cloud Storage`)"

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,224 @@
...
+
+      try {
+        await TelegramCloudStorageStamper.setCloudStorageItem(
+          cloudStorageKey,
+          TelegramCloudStorageStamper.stringifyAPIKey(
+            config.apiPublicKey,
+            config.apiPrivateKey
+          )
+        );
+      } catch (e) {
```
r-n-o, line 76: worth templating `e.toString()` in the error message maybe? There could be multiple types of failures underneath, might be useful to know what happened exactly.

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+```ts
+import { TelegramCloudStorageStamper } from "@turnkey/telegram-cloud-storage-stamper";
+import { TurnkeyBrowserClient, TurnkeySDKClientConfig } from "@turnkey/sdk-browser";
+
+// create a new telegram cloud storage stamper
+const stamper = new TelegramCloudStorageStamper({
+  apiPublicKey: "...",
+  apiPrivateKey: "...",
+});
```
r-n-o, line 22: This needs to be updated to use the new `.create()` pattern

r-n-o (commented):
👏 

Looking real real good!

packages/telegram-cloud-storage-stamper/src/index.ts
zkharit: The reason I didnt display the error message here specifically is that the getter and setter error messages that are thrown display the key and value that were failed to be input. This would display the pub/priv api key which i think in this case is a no no. However maybe I just shouldnt have the key/value printed in the underlying error message, what do you think?

packages/telegram-cloud-storage-stamper/src/index.ts
zkharit: Its not possible for them to be null/undefined/etc.. because the function takes an optional argument of type `TTelegramCloudStorageStamperConfig` which requires that those 2 fields are populated. If the config is not passed in, this code path is skipped

r-n-o: So, maybe not `null` or `undefined`, but empty is still possible? (`""`)

That might be enough of an edge case to be "okay" though.

zkharit: Yeah, theoretically they can pass any value into there "abc" which wouldnt be a valid apikey. Was kinda relying on api-key-stamper's error handling there

.changeset/itchy-balloons-call.md
```diff
@@ -0,0 +1,5 @@
+---
+"@turnkey/telegram-cloud-storage-stamper": minor
+---
+
+Initial release
```
andrewkmin, line 5: let's add a bit more context, or you can point it to the README if you'd like to consolidate the commentary there

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,77 @@
+# @turnkey/telegram-cloud-storage-stamper
+
+[![npm](https://img.shields.io/npm/v/@turnkey/telegram-cloud-storage-stamper?color=%234C48FF)](https://www.npmjs.com/package/@turnkey/telegram-cloud-storage-stamper)
+
+This package contains functions to store a turnkey api public/private key within [Telegram Cloud Storage](https://core.telegram.org/bots/webapps#cloudstorage). This package also handles stamping a Turnkey request with that api key. It is meant to be used with [`@turnkey/http`](https://www.npmjs.com/package/@turnkey/http)
```
andrewkmin, line 5: tiny nit for capitalization -- `Turnkey` and `API`

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,77 @@
...
+// initialize the telegram cloud storage stamper, this stores the api key credentials in Telegram Cloud Storage for the first time
+try {
+  await stamper.init();
+} catch (err) {
+  throw new Error(`Failed initializing Telegram Cloud Storage Stamper: ${err}`)
+}
+
+// use the stamper in the client config
+const browserConfig: TurnkeySDKClientConfig = {
+  stamper: TelegramCloudStorageStamper!,
```
andrewkmin, line 33: from what I can tell, `TelegramCloudStorageStamper` is the name of the class, and `stamper` is the name of the variable. so maybe this needs a minor update?

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,77 @@
...
+// create a TurnkeyClient with the initialized Telegram Cloud Storage Stamper
+const client = new TurnkeyBrowserClient(browserConfig);
+
+// make a request with the client
+const whoamiResponse = await client.getWhoami({
+  organizationId: <ORGANIZATION_ID>,
+});
+```
+
+Use an exisiting key that has been previously stored in Telegram Cloud Storage
```
andrewkmin, line 47: tiny nit: `existing`

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+```ts
+import { TelegramCloudStorageStamper } from "@turnkey/telegram-cloud-storage-stamper";
+import { TurnkeyBrowserClient, TurnkeySDKClientConfig } from "@turnkey/sdk-browser";
+
+// create a new telegram cloud storage stamper
+const stamper = new TelegramCloudStorageStamper();
+
+// initialize the telegram cloud storage stamper, this gets the api key credentials from Telegram Cloud Storage
```
andrewkmin, line 56: tiny nit: `telegram cloud storage stamper` --> `TelegramCloudStorageStamper`, `api key` --> `API key`

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,77 @@
...
+try {
+  await stamper.init();
+} catch (err) {
+  throw new Error(`Failed initializing Telegram Cloud Storage Stamper: ${err}`)
+}
+
+// use the stamper in the client config
+const browserConfig: TurnkeySDKClientConfig = {
+  stamper: TelegramCloudStorageStamper!,
```
andrewkmin, line 65: similar thing here: think we can just use the `stamper` variable

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  async stamp(payload: string) {
+    // check to see if were in a telegram mini app context
+    TelegramCloudStorageStamper.checkTelegramContext();
+
+    // check to see that the stamper was initialized
+    if (!this.stamper) {
+      throw new TelegramCloudStorageStamperError(
+        "Cannot stamp with unintialized telegram stamper"
```
andrewkmin, line 114: tiny nit: `uninitialized`

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,224 @@
...
+        throw new TelegramCloudStorageStamperError(
+          `Failed storing api key in Telegram Cloud Storage`
+        );
+      }
+    } else {
+      try {
+        const apiKey = await TelegramCloudStorageStamper.getCloudStorageItem(
+          cloudStorageKey
+        );
+        ({ apiPublicKey, apiPrivateKey } = TelegramCloudStorageStamper.parseAPIKey(apiKey));
```
andrewkmin, line 86: any particular reason for this method of destructuring? how about

```
const { apiPublicKey, apiPrivateKey } = TelegramCloudStorageStamper.parseAPIKey(apiKey);
```

packages/telegram-cloud-storage-stamper/src/util/telegram-environment.ts
zkharit: changed

packages/telegram-cloud-storage-stamper/src/index.ts
zkharit: Updated this

packages/telegram-cloud-storage-stamper/src/index.ts
zkharit: changed

packages/telegram-cloud-storage-stamper/README.md
zkharit: changed, was a typo!

packages/telegram-cloud-storage-stamper/README.md
zkharit: changed 😅

packages/telegram-cloud-storage-stamper/README.md
zkharit: changed

packages/telegram-cloud-storage-stamper/README.md
zkharit: updated

packages/telegram-cloud-storage-stamper/README.md
zkharit: this is outdated at this point :)

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+// create a new Telegram Cloud Storage Stamper
+const stamper = await TelegramCloudStorageStamper.create({
+  apiPublicKey: "...",
+  apiPrivateKey: "...",
+});
+
+if (!stamper) {
+  // There was a problem initializing the stamper
+}
```
r-n-o, line 45: Is this actually true? I'd expect an error thrown if the stamper couldn't be initialized (instead of a null/undefined/falsy stamper)

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+const apiPublicKey2 = "...";
+const apiPrivateKey2 = "...";
+
+// insert the API key in Telegram Cloud Storage
+const stored = await stamper.setSigningKey(apiPublicKey2, apiPrivateKey2);
```
r-n-o, line 189: what is the value of `stored` here?

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+if(!item) {
```
r-n-o, line 191: `item` is undefined in this code snippet :)

packages/telegram-cloud-storage-stamper/src/__tests__/stamp-test.ts
```diff
@@ -0,0 +1,40 @@
+import { test, expect } from "@jest/globals";
+import TelegramCloudStorageStamper from "../index";
+import { readFixture } from "../__fixtures__/shared";
+import { assertValidSignature } from "./shared";
+
+test("uses provided signature to make stamp", async function () {
+  const { privateKey, publicKey, pemPublicKey } = await readFixture();
+
+  // ToDo: init fails since there isn't an actual Telegram environment here
```
r-n-o, line 9: Is this still relevant? Looks like `create` is succeeding just fine if this test is passing?

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,296 @@
...
+ */
+export default class TelegramCloudStorageStamper {
+  // This stamper uses a typical apikey stamper under the hood and abstracts away the storage of the actual api keys
+  stamper?: ApiKeyStamper | undefined;
+
+  // the constructor checks if were in a telegram mini app context, and then gives the developer an interface into looking into CloudStorage even without having an API key attached
+  constructor() {
+    // check to see if were in a telegram mini app context
+    this.checkTelegramContext();
+  }
```
r-n-o, line 36: is this useful in practice? If this feature detection is useful I'd rather export it as a standalone function ("supportsCloudStorage"). Using this interface means writing:
```ts
try {
    // this actually just checks for local storage, trust me
    // (or go read Turnkey's docs)
    new TelegramCloudStorageStamper()
} catch (e) {
    // abort case
}
// now we're good, cloud storage is supported (look at the try/catch above)
```

Compare this with:
```ts
if (supportsCloudStorage()) {
    // abort
} else {
    // we're good
}
```
No need for extra code comments, the intent becomes very crisp.

r-n-o: Actually, I now see what you mean. The goal of this isn't to do feature detection, the goal is to act as a storage wrapper. So the code would be:
```ts
stamper = new TelegramCloudStorageStamper()
await stamper.getItem("MY_KEY")
await stamper.getItem("OTHER_KEY", "VALUE")
```

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  async insertItem(key: string, value: string) {
+    // check to see if were in a telegram mini app context
+    this.checkTelegramContext();
+
+    try {
+      return await this.setCloudStorageItem(key, value)
+    } catch {
+      return false
+    }
```
r-n-o, line 127: I'm not sure this is a good pattern in general. I'd rather let the exception bubble up to the caller so they can handle it the way they see fit. With "false" returned it's too easy to forget to check this value.

More generally this "check return value" pattern is very C-like or Go-like; most JS devs expect exceptions to be thrown when something goes wrong rather than boolean values indicating success.

packages/telegram-cloud-storage-stamper/README.md
r-n-o: now I have my answer actually: is a success boolean!

packages/telegram-cloud-storage-stamper/src/index.ts
zkharit: There are really 2 sets of functions available for the developer to use. The lower level `setCloudStorageItem`, or `setItem`. I tried to simplify them with the `setItem` variants of the functions and still have setCloudStorageItem available to throw exceptions if the developers want to use them

packages/telegram-cloud-storage-stamper/src/__tests__/stamp-test.ts
zkharit: This test isnt passing just yet, this still needs some work :)

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,195 @@
+# @turnkey/telegram-cloud-storage-stamper
+
+[![npm](https://img.shields.io/npm/v/@turnkey/telegram-cloud-storage-stamper?color=%234C48FF)](https://www.npmjs.com/package/@turnkey/telegram-cloud-storage-stamper)
+
+This package contains functions to store a Turnkey API public/private key within [Telegram Cloud Storage](https://core.telegram.org/bots/webapps#cloudstorage). This package also handles stamping a Turnkey request with that API key. It is meant to be used with [`@turnkey/sdk-browser`](https://www.npmjs.com/package/@turnkey/sdk-browser).
```
andrewkmin, line 5: If we're going to mention `@turnkey/sdk-browser` specifically here, it would be optimal to have some integration guide type docs to flesh that out. 

Technically, if this implements the general stamper interface, it could be used with the lowest level `@turnkey/http` as well — we could mention that but generally recommend usage with `@turnkey/sdk-browser`. This also might make sense to... add to `@turnkey/sdk-react` as well?

In any case, a docs update to https://docs.turnkey.com/api-overview/stamps#stampers at the very least seems relevant

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,195 @@
...
+Telegram Bot API >= 6.9
+
+The Telegram Cloud Storage Stamper has a few different modes of operation, namely a classic [stamper](https://docs.turnkey.com/api-overview/stamps) for stamping requests made to Turnkey's API, and an interface for a Telegram Mini App built with Turnkey to interact with Telegram Cloud Storage. This provides the developer of the application utilities such as creating stamps on requests made by users, storing user API keys, storing temporary keys that are needed for decrypting credential bundles for activites like [email auth](https://docs.turnkey.com/features/email-auth) or [oauth](https://docs.turnkey.com/features/oauth), or storing arbitrary values that would be helpful to have saved for a user from session to session on device to device.
+
+The Telegram Cloud Storage Stamper will, by default, store the API key used for signing in Telegram Cloud Storage under the key `TURNKEY_API_KEY`. A Cloud Storage "key" is the index under which a value is stored in Telegram Cloud Storage. This can be changed when using the `.create()` or `.setSigningKey()` functions. An API key is stored within Cloud Storage as a JSON string of the following object:
+```
```
andrewkmin, line 13: If we want to go above and beyond here we can include how you might generate this key (using https://github.com/tkhq/sdk/blob/ed92c90b6d4259f010933dfb9e16b1b62a7dcba2/packages/crypto/src/crypto.ts#L268-L284)

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,195 @@
...
+{
+  apiPublicKey: "compressedApiPublicKeyHex",
+  apiPrivateKey: "apiPrivateKeyHex",
+}
+```
+
+#### Argument Usage
+
+The `.create()` and `.setSigningKey()` functions take one of the following 4 sets of arguments:
+- No arguments: Use an API key at the default location within Telegram Cloud Storage `TURNKEY_API_KEY` and set that as the signing key
```
andrewkmin, line 23: What does the default location mean here?

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,195 @@
...
+- Just an API key: Store the passed in API key and at the default Telegram Cloud Storage location and set that as the signing key
+- Just a Cloud Storage key: Use an API key stored at the specified Telegram Cloud Storage key location and set that as the signing key
+- Both an API key and a Cloud Storage key: Store the passed API key at the specified Telegram Cloud Storage key location and set that as the signing key
+  
+The `.getAPIKey()` and `.setAPIKey()` functions operate in a similar manner taking an optional `key` parameter that will be used to `get` or `set` the API key at that location if it is passed, or at the default location if it is not passed.
+
+### Usage
+Insert a new api key into Telegram Cloud Storage at the default API key location
```
andrewkmin, line 31: tiny nit: `api -> API`

packages/telegram-cloud-storage-stamper/README.md
andrewkmin: On the flip side, I can see the argument that we shouldn't because telegram is a pretty separate path and thus shouldn't be included with the other more vanilla browser use cases

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,195 @@
...
+// make a request with the client
+const whoamiResponse = await client.getWhoami({
+  organizationId: <ORGANIZATION_ID>,
+});
+```
+
+Use an existing key that has been previously stored in Telegram Cloud Storage at the default API key location key location
+
+```ts
+import { TelegramCloudStorageStamper } from "@turnkey/telegram-cloud-storage-stamper";
```
andrewkmin, line 66: thoughts on calling this just `@turnkey/telegram-stamper`? Would that be pigeonholing too much? i.e. are there other types of telegram stampers that we could foresee having some day in the future?

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,195 @@
...
+// create a TurnkeyClient with the initialized Telegram Cloud Storage Stamper
+const client = new TurnkeyBrowserClient(browserConfig);
+
+// make a request with the client
+const whoamiResponse = await client.getWhoami({
+  organizationId: <ORGANIZATION_ID>,
+});
+```
+
+View an entry in Telegram Cloud Storage without inserting an API key, note the usage difference between the `new` and `.create()` here. `.create()` will do the work of getting/setting an API key in Cloud Storage where as  `new` will not
```
andrewkmin, line 92: tiny nit: extra space between `as` and `new`?

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,195 @@
...
+
+// create a new Telegram Cloud Storage Stamper
+const stamper = new TelegramCloudStorageStamper();
+
+if (!stamper) {
+  // There was a problem initializing the stamper
+}
+
+// the key used to index Telegram Cloud Storage
+const telegramCloudStorageKey = "telegramCloudStorageKey";
```
andrewkmin, line 105: For consistency, we could type this out a tiny bit differently using a `@turnkey/` prefix, similar to what's been defined here: https://github.com/tkhq/sdk/blob/main/packages/sdk-browser/src/storage.ts

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,195 @@
...
+  // There was a problem initializing the stamper
+}
+
+const apiPublicKey = "...";
+const apiPrivateKey = "...";
+
+// the key used to index Telegram Cloud Storage
+const telegramCloudStorageKey = "telegramCloudStorageKey";
+
+// insert the API key in Telegram Cloud Storage
```
andrewkmin, line 134: for the email auth type use case I imagine?

packages/telegram-cloud-storage-stamper/README.md
andrewkmin, line 1: definitely appreciate the thoroughness of this readme!

packages/telegram-cloud-storage-stamper/README.md
zkharit: changed this to throw an error instead of returning undefined

packages/telegram-cloud-storage-stamper/README.md
zkharit: updated the readme in a few places for this

packages/telegram-cloud-storage-stamper/README.md
zkharit: Had this discussion with @r-n-o earlier. It was originally telegram-stamper. There are talks of including Login with Telegram outside of a telegram mini app context. If we want to create a specialized stamper for that, this one will not work because there will not be access to TG cloud storage in a typical web app context so it is possible there is a different telegram stamper that could be developed. But its possible wed just advise users to utilize one our existing stampers if they are not in a telegram context

packages/telegram-cloud-storage-stamper/README.md
zkharit: Can be anything the user desires here really :) Just tried to be flexible for whatever the user might need. Certainly email auth is a good example

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,296 @@
...
+    this.checkTelegramContext();
+
+    try {
+      return await this.getCloudStorageItem(key)
+    } catch {
+      return ""
+    }
+  }
+
+  async getAPIKey(key?: string) {
```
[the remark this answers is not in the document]
zkharit: changed

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,296 @@
...
+const DEFAULT_TURNKEY_CLOUD_STORAGE_KEY = "TURNKEY_API_KEY";
+
+/**
+ * Stamper to use within a `TurnkeyClient`
+ */
+export default class TelegramCloudStorageStamper {
+  // This stamper uses a typical apikey stamper under the hood and abstracts away the storage of the actual api keys
+  stamper?: ApiKeyStamper | undefined;
+
+  // the constructor checks if were in a telegram mini app context, and then gives the developer an interface into looking into CloudStorage even without having an API key attached
```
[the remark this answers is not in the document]
zkharit: updated in a bunch of places

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,296 @@
...
+};
+
+// Constant for default key name
+const DEFAULT_TURNKEY_CLOUD_STORAGE_KEY = "TURNKEY_API_KEY";
+
+/**
+ * Stamper to use within a `TurnkeyClient`
+ */
+export default class TelegramCloudStorageStamper {
+  // This stamper uses a typical apikey stamper under the hood and abstracts away the storage of the actual api keys
```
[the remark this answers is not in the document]
zkharit: updated

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,296 @@
...
+  cloudStorageKey?: string;
+};
+
+export type CloudStorageAPIKey = {
+  apiPublicKey: string;
+  apiPrivateKey: string;
+};
+
+// Constant for default key name
+const DEFAULT_TURNKEY_CLOUD_STORAGE_KEY = "TURNKEY_API_KEY";
```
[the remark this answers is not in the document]
zkharit: changed to `@turnkey/TURNKEY_API_KEY`

packages/telegram-cloud-storage-stamper/README.md
zkharit: changed

packages/telegram-cloud-storage-stamper/README.md
zkharit: fixed

packages/telegram-cloud-storage-stamper/README.md
zkharit: fixed in lots of places

packages/telegram-cloud-storage-stamper/README.md
zkharit: Fleshed out the first example to include using generateP256KeyPair

packages/telegram-cloud-storage-stamper/src/index.ts
zkharit: They will always be `TelegramCloudStorageStamperError`'s You can see [here](https://github.com/tkhq/sdk/blob/a375c266e55596bea09a2f493dcc3ee2af2e34bc/packages/telegram-cloud-storage-stamper/src/index.ts#L250-L256). There are some lower level functions that will throw errors and then some higher level functions which I thought would be easier for developers to use. Ill make a distinction about this in the README so that is clear.

packages/telegram-cloud-storage-stamper/README.md
zkharit: I specifically removed a reference to @turnkey/http here after our convo during standup yesterday 😅 to push users toward our higher level libraries. I do agree that extra doc will be helpful here, and I'm planning to add some :)

.changeset/itchy-balloons-call.md
zkharit: added!

packages/telegram-stamper/src/__tests__/stamp-test.ts
zkharit: Added a mock local storage here to test this functinoality!

packages/telegram-cloud-storage-stamper/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# @turnkey/telegram-cloud-storage-stamper
+
+## 0.1.0
+
+Initial release
```
r-n-o, line 5: if you have a changeset I think you don't need a changelog? It'll automatically be generated when someone calls `pnpm run changeset version`

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,210 @@
...
+// create a TurnkeyClient with the initialized Telegram Cloud Storage Stamper
+const client = new TurnkeyBrowserClient(browserConfig);
+
+// make a request with the client
+const whoamiResponse = await client.getWhoami({
+  organizationId: <ORGANIZATION_ID>,
+});
+```
+
+View an entry in Telegram Cloud Storage without inserting an API key, note the usage difference between the `new` and `.create()` here. `.create()` will do the work of getting/setting an API key in Cloud Storage where as `new` will not
```
r-n-o, line 99: "whereas"

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+```ts
+import TelegramCloudStorageStamper, {
+  CloudStorageAPIKey,
+} from "@turnkey/telegram-cloud-storage-stamper";
+
+// create a new Telegram Cloud Storage Stamper, "new" is used when you don't want to store or retreieve any API keys, and just need an interface into Cloud Storage
```
r-n-o, line 106: retrieve

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,210 @@
...
+// insert the API key in Telegram Cloud Storage
+const stored = await stamper.insertAPIKey(
+  apiPublicKey,
+  apiPrivateKey,
+  telegramCloudStorageKey
+);
+
+if (!stored) {
+  // failed storing API key
+}
```
r-n-o, line 143: I think this can be omitted now that you switched to the exception model?

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,210 @@
...
+
+// insert the API key in Telegram Cloud Storage
+const stored = await stamper.setSigningKey({
+  cloudStorageAPIKey: apiKey,
+  cloudStorageKey: telegramCloudStorageKey,
+});
+
+if (!stored) {
+  // failed setting signing key
+}
```
r-n-o, line 176: likewise here

packages/telegram-cloud-storage-stamper/README.md
```diff
@@ -0,0 +1,210 @@
...
+};
+
+// insert the API key in Telegram Cloud Storage
+const stored = await stamper.setSigningKey({
+  cloudStorageAPIKey: apiKey2,
+});
+
+if (!stored) {
+  // failed setting signing key
+}
```
r-n-o, line 209: And here

packages/telegram-cloud-storage-stamper/src/__tests__/stamp-test.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    // We assert against these keys one-by-one because P256 signatures aren't deterministic.
+    // Can't snapshot!
+    expect(decodedStamp["publicKey"]).toBe(publicKey);
+    expect(decodedStamp["scheme"]).toBe("SIGNATURE_SCHEME_TK_API_P256");
+    assertValidSignature({
+      content: messageToSign,
+      pemPublicKey: pemPublicKey,
+      signature: decodedStamp["signature"],
+    });
```
r-n-o, line 70: Beyond the stamp, let's also check that the right values are persisted in fake-cloud-storage aka fake-local-storage?

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,300 @@
...
+      let { apiPublicKey, apiPrivateKey } = await this.getAPIKey(
+        DEFAULT_TURNKEY_CLOUD_STORAGE_KEY
+      );
+
+      if (apiPublicKey && apiPrivateKey) {
+        this.stamper = new ApiKeyStamper({
+          apiPublicKey,
+          apiPrivateKey,
+        });
+        return true;
```
r-n-o, line 79: I don't think it's necessary to return a boolean, the function doesn't document a return type. I think `return;` (implicitly: `return undefined`) work better? Otherwise we're back to having devs wondering what this boolean means and what they should do with it.

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,300 @@
...
+
+      if (stored === true) {
+        this.stamper = new ApiKeyStamper({
+          apiPublicKey: config.cloudStorageAPIKey.apiPublicKey,
+          apiPrivateKey: config.cloudStorageAPIKey.apiPrivateKey,
+        });
+        return true;
+      }
+    }
+    return false;
```
r-n-o, line 126: let's throw an exception here, guessing this is not supposed to happen!

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,300 @@
...
+  ) {
+    try {
+      return await this.insertItem(
+        key,
+        this.stringifyAPIKey(apiPublicKey, apiPrivateKey)
+      );
+    } catch {
+      return false;
+    }
+  }
```
r-n-o, line 153: same here. I think the interface should just be an undefined return value rather than an undocumented boolean. Let's throw if something goes wrong.

I actually think this function can potentially stop existing because it'd just be:
```ts
async insertAPIKey(
    apiPublicKey: string,
    apiPrivateKey: string,
    key: string = DEFAULT_TURNKEY_CLOUD_STORAGE_KEY
  ) {
    return await this.insertItem(
        key,
        this.stringifyAPIKey(apiPublicKey, apiPrivateKey)
      );
}
```

I think the caller can call `this.InsertItem` themselves? I don't feel super strongly either way.

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,300 @@
...
+        (err: any, value: string) => {
+          if (err != null || !value) {
+            reject(
+              new TelegramCloudStorageStamperError(
+                `Failed getting value: ${key} from Telegram Cloud Storage${
+                  err && `: ${err}`
+                }`
+              )
+            );
+          }
```
r-n-o, line 248: We need to be a bit careful here, if value is "", this will throw an error. Being able to distinguish between "error" and "empty" cases is important and right now it'd be really hard (devs would have to try/catch and inspect the message).

Instead, let's separate the two?
* If there's an error, `reject` (what you have)
* if there's no error, return `value` (doesn't matter that the value is `""` or undefined IMO, passing what you get out of the TG API will be less confusing than if you tried to transform it)

packages/telegram-cloud-storage-stamper/src/index.ts
r-n-o: Basically, if this function didn't exist, we'd go from:
```ts
let stored = await this.insertAPIKey(
    config.cloudStorageAPIKey.apiPublicKey,
    config.cloudStorageAPIKey.apiPrivateKey,
    DEFAULT_TURNKEY_CLOUD_STORAGE_KEY
);
```

to
```ts
this.insertItem(
    DEFAULT_TURNKEY_CLOUD_STORAGE_KEY,
    this.stringifyAPIKey(config.cloudStorageAPIKey.apiPublicKey, config.cloudStorageAPIKey.apiPrivateKey)
);
```

I actually like the version with `this.insertAPIKey` so maybe let's leave this function in! Cheap enough to keep :)

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,300 @@
...
+        (err: any, removed: boolean) => {
+          if (err || !removed) {
+            reject(
+              new TelegramCloudStorageStamperError(
+                `Failed removing key: ${key}${err && `: ${err}`}`
+              )
+            );
+          }
+
+          resolve(removed);
```
r-n-o, line 295: you can simply `resolve()` without a boolean since you already check the boolean above. I think it's a cleaner interface?

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,300 @@
...
+            reject(
+              new TelegramCloudStorageStamperError(
+                `Failed inserting value: ${value} into Telegram Cloud Storage at key: ${key}${
+                  err && `: ${err}`
+                }`
+              )
+            );
+          }
+
+          resolve(stored);
```
r-n-o, line 272: likewise here. You don't return a typed promise (e.g. `Promise<boolean>`) so I think this can be `resolve()`. There's no value in returning this boolean because it will be `true`, always. If it was `false`, you'd error out on line 264 :)

r-n-o (commented):
Sorry, one more round of comments on this! We're getting closer and closer and closer I swear!

packages/telegram-cloud-storage-stamper/src/index.ts
zkharit: removed

packages/telegram-cloud-storage-stamper/src/index.ts
zkharit: throwing

packages/telegram-cloud-storage-stamper/src/index.ts
zkharit: changed

packages/telegram-cloud-storage-stamper/src/index.ts
zkharit: changed

packages/telegram-cloud-storage-stamper/src/index.ts
zkharit: This logic has been overhauled and is no longer relevant

packages/telegram-cloud-storage-stamper/src/__tests__/stamp-test.ts
zkharit: added a test in here to confirm that the api key stored in local/cloud storage equals the one that was generated at the beginning of the test

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,274 @@
...
+      );
+    }
+  }
+
+  isApiKey(apiKey: CloudStorageAPIKey) {
+    return (
+      typeof apiKey.apiPublicKey === "string" &&
+      typeof apiKey.apiPrivateKey === "string"
+    );
+  }
```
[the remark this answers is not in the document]
zkharit: The usage of this function comes from the need to determine if a value obtained and parsed from tg cloud storage is of type CloudStorageAPIKey, not necessarily that it is a "valid" api key. I changed the name to be isCloudStorageAPIKey and added a comment to be more clear

packages/telegram-cloud-storage-stamper/src/index.ts
```diff
@@ -0,0 +1,274 @@
...
+
+  async getItem(key: string) {
```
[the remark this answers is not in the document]
zkharit: added the type annotations, typescript could not infer it previously

PR #400, opened 2024-10-17
Refactoring auth examples to use sdk-server/sdk-react

moe-dev:
## Did you add a changeset?


robin-turnkey (approved):
Valor.

examples/email-auth/src/pages/api/auth.ts
```diff
@@ -29,53 +28,34 @@ export default async function auth(
...
-    if (!apiKeyId) {
-      throw new Error("Expected a non-null API key ID!");
+    if (!userId || !apiKeyId) {
+      throw new Error("Expected non-null values for userId and apiKeyId.");
     }
 
     res.status(200).json({
       userId,
       apiKeyId,
-      // This is simple in the case of a single organization
```
andrewkmin, line 72: is this no longer valid?

examples/email-auth/src/pages/index.tsx
```diff
@@ -43,75 +39,57 @@ export default function AuthPage() {
...
-          {
-            curve: "CURVE_SECP256K1",
-            pathFormat: "PATH_FORMAT_BIP32",
-            path: "m/44'/60'/0'/0/0",
-            addressFormat: "ADDRESS_FORMAT_ETHEREUM",
-          },
-        ],
-      },
+      walletName: data.walletName,
+      accounts: [
```
andrewkmin, line 82: tiny nit: we could pull this accounts config out of `sdk-browser` :)

examples/otp-auth/src/pages/index.tsx
```diff
@@ -95,44 +91,26 @@ export default function AuthPage() {
...
-          {
-            curve: "CURVE_SECP256K1",
-            pathFormat: "PATH_FORMAT_BIP32",
-            path: "m/44'/60'/0'/0/0",
-            addressFormat: "ADDRESS_FORMAT_ETHEREUM",
-          },
-        ],
-      },
+      walletName: data.walletName,
+      accounts: [
```
andrewkmin, line 104: same here, but again, a nit

andrewkmin (approved):
very clean!

examples/email-auth/src/pages/index.tsx
moe-dev: Would require installing a new package which may get confusing to users as to why we would require sdk-browser for this example? Perhaps it makes sense to export from react as well

andrewkmin: sorry, meant `sdk-server`. both server and browser offer the helpers: https://github.com/tkhq/sdk/blob/main/packages/sdk-server/src/turnkey-helpers.ts

PR #404, opened 2024-10-19
Change Inquirer/prompts dependency to prompts

omkarshanbhag:
## Summary & Motivation

This PR changes it so that we use the `prompts` library instead of `inquirer/prompts` for consistency with our stackblitz examples -- since it seems friendlier to be used with `pnpm` or `npm`

## How I Tested These Changes

Tested the edited examples locally

## Did you add a changeset?

### DIFF
--- a/examples/export-in-node/src/index.ts
+++ b/examples/export-in-node/src/index.ts
@@ -1,6 +1,6 @@
 import * as dotenv from "dotenv";
 import * as path from "path";
-import { input } from "@inquirer/prompts";
+import prompts from "prompts";
 import { Turnkey } from "@turnkey/sdk-server";
 import { Crypto } from "@peculiar/webcrypto";
 import { generateP256KeyPair, decryptExportBundle } from "@turnkey/crypto";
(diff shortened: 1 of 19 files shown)


examples/with-solana/src/addSignature.ts
```diff
@@ -56,22 +56,22 @@ async function main() {
...
+    await prompts({ message: "Ready to Continue?" });
     // refresh balance...
     balance = await solanaNetwork.balance(connection, solAddress);
   }
 
   print("SOL balance:", `${balance} Lamports`);
 
   // 1. Create, sign, and verify a transfer transaction using `addSignature` as opposed to `signTransaction`
-  const destination = await input({
+  const destination = await prompts({
```
andrewkmin, line 67: was this tested? there are some syntax differences with using `prompts` vs. `inquirer/prompts`. you can refer to the stackblitz example itself https://stackblitz.com/edit/stackblitz-starters-xeb93i?file=src%2Findex.ts

examples/with-solana/src/addSignature.ts
```diff
@@ -56,39 +56,59 @@ async function main() {
...
-    default: TURNKEY_WAR_CHEST,
-  });
+  const { destination } = await prompts([
+    {
+      type: 'text',
+      name: 'destionation',
+      message: `Destination address:`,
+      default: TURNKEY_WAR_CHEST,
+    }
+    ]);
```
omkarshanbhag, line 86: prettier (probably)

examples/export-in-node/src/index.ts
```diff
@@ -1,10 +1,13 @@
 import * as dotenv from "dotenv";
 import * as path from "path";
-import { input } from "@inquirer/prompts";
+import prompts from "prompts";
 import { Turnkey } from "@turnkey/sdk-server";
 import { Crypto } from "@peculiar/webcrypto";
 import { generateP256KeyPair, decryptExportBundle } from "@turnkey/crypto";
-global.crypto = new Crypto();
+if (typeof crypto === "undefined") {
```
omkarshanbhag, line 7: @andrewkmin @moe-dev Tagging for context

This was the fix for the following error that I ran into while running this example: 

`TypeError: Cannot set property crypto of #<Object> which has only a getter`

omkarshanbhag: Same for `import-in-node`

examples/with-solana/src/advanced.ts
```diff
@@ -58,48 +58,69 @@ async function main() {
...
+      type: "text",
+      name: "numTxsStr",
       message: `Number of transactions:`,
-      default: "1",
-    })
-  );
+      initial: "1",
+    },
+  ]);
+  console.log(numTxsStr);
```
andrewkmin, line 82: 🧹

andrewkmin (commented):
nice! this is super sick. great to consolidate on a single interface for using prompts in our examples, and importantly, makes all of these examples more compatible with stackbiltz/other environments 👏 

I'd like to get to the bottom of the `global.crypto` issue before merging as it's one that I hadn't run into when running locally myself. once resolved, let's 🚢 !

examples/export-in-node/src/index.ts
omkarshanbhag: Node version --> `v22.5.1`

Pnpm version --> `8.4.0`

examples/with-solana/src/advanced.ts
omkarshanbhag: Done!

andrewkmin (approved):

examples/export-in-node/src/index.ts
omkarshanbhag: Posting quick follow up for visibility --> We tested this issue and it seems like this is a node versioning issue 

Old code works with node 18 but not with node 20

omkarshanbhag:
Tested import and export successfully!

PR #405, opened 2024-10-21
add zerodev account abstraction example

andrewkmin:
## Summary & Motivation
$title

## How I Tested These Changes
locally ~

## Did you add a changeset?


examples/with-zerodev-aa/src/viem.ts
```diff
@@ -0,0 +1,149 @@
...
+  apiPrivateKey: process.env.API_PRIVATE_KEY!,
+  apiPublicKey: process.env.API_PUBLIC_KEY!,
+  defaultOrganizationId: process.env.ORGANIZATION_ID!,
+});
+
+const main = async () => {
+  if (!process.env.SIGN_WITH) {
+    // If you don't specify a `SIGN_WITH`, we'll create a new wallet for you via calling the Turnkey API.
+    await createNewWallet();
+    return;
```
turnekybc, line 49: Is the expectation that after we create a new wallet that the user then updates `process.env.SIGN_WITH` with the wallet? Or do we want to proceed with the flow using the newly created wallet?

andrewkmin: The latter probably makes the most sense. Unfortunately there are a lot of places we follow this pattern 😅 

On the flip side, having the user set the `SIGN_WITH` env var is helpful so that they set it for subsequent runs of the example

turnekybc (approved):

PR #407, opened 2024-10-23
Fix SubOrgID issue with Passkey Examples

omkarshanbhag:
## Summary & Motivation

This PR fixes issues in examples `with-solana-passkeys` and `with-eth-passkeys-galore` where a passkey client makes sign requests with the orgID of the parent org and not the suborg

## How I Tested These Changes

ran examples locally

## Did you add a changeset?

### DIFF
--- a/examples/with-eth-passkeys-galore/src/pages/index.tsx
+++ b/examples/with-eth-passkeys-galore/src/pages/index.tsx
@@ -132,6 +132,10 @@ export default function Home() {
 
     const response = res.data as TWalletDetails;
     setWallet(response);
+
+    if (passkeyClient && passkeyClient.config) {
+      passkeyClient.config.organizationId = response.subOrgId;
+    }
   };
 
   const login = async () => {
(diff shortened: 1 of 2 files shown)


examples/with-eth-passkeys-galore/src/pages/index.tsx
```diff
@@ -67,6 +67,10 @@ export default function Home() {
   };
 
   const signMessageViem = async (data: signingFormData) => {
+    if (passkeyClient && passkeyClient.config && wallet) {
```
andrewkmin, line 70: we're going to need this for other methods like `signTransactionViem` and `signTransactionEthers` as well. however, in order to avoid repeating ourselves, the easier solution IMO is to do this at the end of the `createSubOrgAndWallet` call right here: https://github.com/tkhq/sdk/blob/402c3bd55a07160ff55e9d53a3ec49aaf5118c89/examples/with-eth-passkeys-galore/src/pages/index.tsx#L141-L142

also, from a syntax perspective, you can use something like `passkeyClient?.config.organizationId = wallet.subOrgId;`

this implicitly does the `if (passkeyClient && ...`) check under the hood

andrewkmin (approved):

PR #409, opened 2024-10-24
Taylor/wallet stamper client

taylorjdawson:
##### `WalletStamper`

  -  Renamed `recoverPublicKey` to `getPublicKey`
    **Reason**: Improves clarity and consistency across wallet interfaces
  
  - Unified `getPublicKey` method signature to take no parameters
   **Reason**: The change in method signature for `getPublicKey` (removing parameters) was necessary to allow the `TurnkeyWalletClient` to access the public key without needing to provide any additional information. This simplifies the process of retrieving the public key for sub-organization creation as [demonstrated](https://github.com/tkhq/sdk/pull/409/files#diff-ab5388bd43f282fd7eed5bef6c0322f402e1f313b02bc50111b3e03e38fbab25R101-R107) in the example.
  
  - Moved `getPublicKey` to `BaseWalletInterface`
   **Reason**: Ensures all wallet types implement this essential method
  
  - Updated `SolanaWalletInterface` and `EthereumWalletInterface`
   **Reason**: Streamlines interfaces by removing redundant method declarations
  
  - Modified `EthereumWallet` implementation
   **Reason**: Aligns with new interface structure and utilizes `signMessage` for public key recovery

##### `TurnkeyWalletClient`
- Added new `TurnkeyWalletClient` to the `@turnkey/sdk-browser`
 **Reason**: Allows using the `WalletStamper` with the browser sdk
- Added `getPublicKey` method to `TurnkeyWalletClient`
 **Reason**: Enables easy access to wallet public key for sub-organization creation and future authentication flows

- Updated `TurnkeyWalletClient` to use new `WalletInterface`
 **Reason**: Ensures compatibility with the updated Wallet Stamper interfaces

##### `AuthClient` (new enum)

- Introduced a new enum to track which client is authenticated (Passkey, Wallet, Iframe)
  **Reason**: To be stored in the new user session local storage object, to determine which client to return in the `@turnkey/sdk-react`

##### `TurnkeyBrowserClient`, `TurnkeyIframeClient`, `TurnkeyPasskeyClient`, `TurnkeyWalletClient`

- Added a static `authClient` property to base `TurnkeyBrowserClient` to be used by the child classes.
  **Reason**: To track which client was used for the initial authentication, enabling proper client retrieval later

- Set specific `authClient` values in each subclass
  **Reason**: To ensure each client type is correctly identified for authentication and retrieval purposes

##### `UserSession` interface
- Added a new `UserSession` interface which is to be stored in local storage to track the authentication state of the user.
  **Reason**: To eliminate the need to store the write and read sessions separately. Improving clarity and maintainability.  
- Added `authenticatedClient` in the session object
  **Reason**: To store the authentication method used in the user's session data. Will be used in the `@turnkey/sdk-react` to determine which client to return.

This is the new `UserSession` object which gets stored in local storage:
```
export interface User {
  userId: string;
  username: string;
  organization: {
    organizationId: string;
    organizationName: string;
  };
  session:
    | {
        read?: ReadOnlySession;
        write?: ReadWriteSession;
        authenticatedClient: AuthClient;
      }
    | undefined;
}
```

##### `StorageKeys` enum
- Added new  versioned `UserSession` key:  `"@turnkey/session/v1"` Added a new storage key to remove reliance on the legacy user session object with the goal of not breaking any existing implementations with past versions of the sdk.
  **Reason**: To facilitate future updates without breaking existing implementations and improve backward compatibility.

##### `login` and `loginWithReadWriteSession` methods

- Updated to use the new `authenticatedClient` property when creating user sessions
  **Reason**: To properly track and store the authentication method used during login

## Did you add a changeset?


examples/with-wallet-stamper/src/components/turnkey-provider.tsx
```diff
@@ -139,7 +130,10 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
   }
 
   async function signInWithWallet(): Promise<User | null> {
-    const publicKey = (wallet as SolanaWalletInterface)?.recoverPublicKey();
+    const publicKey = await wallet?.getPublicKey();
```
taylorjdawson, line 133: Reflects change to the wallet stamper interface, we now use `getPublicKey` for both EVM and Solana.

examples/with-wallet-stamper/src/components/turnkey-provider.tsx
```diff
@@ -101,13 +98,7 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
...
   ) {
     setAuthenticating(true);
-    let publicKey = null;
-    if (chainType === ChainType.SOLANA) {
-      const solanaWallet = wallet as SolanaWalletInterface;
-      publicKey = solanaWallet.recoverPublicKey();
-    } else if (chainType === ChainType.EVM) {
-      const evmWallet = wallet as EvmWalletInterface;
-    }
+    const publicKey = await wallet?.getPublicKey();
```
taylorjdawson, line 101: Reflects change to the wallet stamper interface, we now use `getPublicKey` for both EVM and Solana.

andrewkmin, line 101: in the EVM case, this will require another passkey tap correct?

packages/sdk-browser/README.md
```diff
@@ -12,27 +12,126 @@ Turnkey API documentation lives here: https://docs.turnkey.com.
...
+
+#### Passkey
+
+The Passkey client allows for authentication to Turnkey's API using Passkeys.
+
+```typescript
+const passkeyClient = turnkey.passkeyClient();
+
+// User will be prompted to login with their passkey
+await passkeyClient.login();
```
andrewkmin, line 38: do we want to describe what exactly login is doing? or perhaps this can wait for Omkar's work on documenting all these methods

cc @omkarshanbhag as some of these methods will be getting tweaked

packages/sdk-browser/README.md
```diff
@@ -12,27 +12,126 @@ Turnkey API documentation lives here: https://docs.turnkey.com.
...
+by passing in `https://auth.turnkey.com` as the `iframeUrl`.
+
+```typescript
+const iframeClient = await turnkey.iframeClient({
+  // The container element that will host the iframe
+  iframeContainer: document.getElementById("<iframe container id>"),
+  iframeUrl: "https://auth.turnkey.com",
+});
+
+const response = await iframeClient.injectCredentialBundle(
```
andrewkmin, line 60: tiny nit: let's use something like `injected` here given I don't think the response for `injectCredentialBundle` is meaningful

packages/sdk-browser/README.md
```diff
@@ -12,27 +12,126 @@ Turnkey API documentation lives here: https://docs.turnkey.com.
...
+import { WalletStamper } from "@turnkey/wallet-stamper";
+
+// Create a new wallet client with a JSON-RPC account from the injected provider
+const walletClient = createWalletClient({
+  chain: mainnet,
+  transport: custom(window.ethereum!),
+});
 
-// This config contains parameters including base URLs, iframe URLs, org ID, and rp ID (relying party ID for WebAuthn)
-import turnkeyConfig from "./turnkey.json";
```
andrewkmin, line 23: good catch!

packages/sdk-browser/README.md
```diff
@@ -12,27 +12,126 @@ Turnkey API documentation lives here: https://docs.turnkey.com.
...
+  return walletClient.signMessage({
+    account,
+    message,
+  });
+};
 
-// Use the config to instantiate a Turnkey Client
-const turnkeyClient = new TurnkeyBrowserSDK(turnkeyConfig);
+const getPublicKey = async () => {
+  // Required to recover the secp256k1 public key from the signature
```
andrewkmin, line 112: as the `arbitraryMessage` variabole name indicates, we can explicitly add that the content of the message doesn't matter

packages/sdk-browser/README.md
```diff
@@ -12,27 +12,126 @@ Turnkey API documentation lives here: https://docs.turnkey.com.
...
+  const arbitraryMessage = "getPublicKey";
+  const signature = await signMessage(arbitraryMessage);
 
-// Now you can make authenticated requests!
-const response = await turnkeyClient?.passkeySign.login();
+  const secp256k1PublicKey = recoverPublicKey({
```
andrewkmin, line 116: per the PR description, are we renaming this to `getPublicKey`?

packages/sdk-browser/README.md
```diff
@@ -12,27 +12,126 @@ Turnkey API documentation lives here: https://docs.turnkey.com.
...
+    getPublicKey,
+  },
+});
+
+// Make authenticated requests to Turnkey API, such as listing user's wallets
+// User will be prompted to sign a message to authenticate the request
+const walletsResponse = await walletClient.getWallets();
 ```
 
 ## Helpers
```
andrewkmin, line 135: cc @omkarshanbhag probably makes sense to add links to your upcoming docs here :)

packages/sdk-browser/src/__types__/base.ts
```diff
@@ -105,3 +107,25 @@ export interface IframeClientParams {
...
+/**
+ * The Client used to authenticate the user.
+ */
+export enum AuthClient {
+  Passkey = "passkey",
+  Wallet = "wallet",
+  Iframe = "iframe",
+}
+
+export type SessionResponse = Omit<
```
andrewkmin, line 124: for consistency, we can call this `TSessionResponse`

packages/sdk-browser/src/models.ts
```diff
@@ -1,18 +1,48 @@
+import type { AuthClient } from "./__types__/base";
```
andrewkmin, line 1: cc @omkarshanbhag we'll probably want some level of coverage on this as well

packages/sdk-browser/src/models.ts
```diff
@@ -1,18 +1,48 @@
...
+        // Authenticated client associated with the session.
+        authenticatedClient: AuthClient;
+      }
+    | undefined;
 }
 
 export interface ReadOnlySession {
-  session: string;
-  sessionExpiry: number;
+  token: string;
```
andrewkmin, line 39: should we document where/how this token might be obtained?

packages/sdk-browser/src/models.ts
```diff
@@ -1,18 +1,48 @@
...
+  expiry: number;
 }
 
 export interface ReadWriteSession {
-  authBundle: string;
-  sessionExpiry: number;
+  credentialBundle: string;
```
andrewkmin, line 44: same for this credential bundle

packages/sdk-browser/src/models.ts
```diff
@@ -1,18 +1,48 @@
...
+        authenticatedClient: AuthClient;
+      }
+    | undefined;
 }
 
 export interface ReadOnlySession {
-  session: string;
-  sessionExpiry: number;
+  token: string;
+  expiry: number;
```
andrewkmin, line 40: we can also document how this expiry is defined (i.e. unix timestamp)

packages/sdk-browser/src/sdk-client.ts
```diff
@@ -10,8 +10,10 @@ import type {
   TurnkeySDKClientConfig,
   TurnkeySDKBrowserConfig,
   IframeClientParams,
+  TurnkeyWalletClientConfig,
+  SessionResponse,
 } from "./__types__/base";
-
+import { AuthClient } from "./__types__/base";
```
andrewkmin, line 16: tiny nit: can combine all these imports from `"./__types__/base"`

packages/sdk-browser/src/sdk-client.ts
```diff
@@ -217,15 +234,63 @@ export class TurnkeyBrowserSDK {
...
 export class TurnkeyBrowserClient extends TurnkeySDKClientBase {
-  constructor(config: TurnkeySDKClientConfig) {
+  authClient: AuthClient | undefined;
+
+  constructor(config: TurnkeySDKClientConfig, authClient?: AuthClient) {
     super(config);
+    this.authClient = authClient;
+  }
+
+  async saveSession({
```
andrewkmin, line 252: would be great to have this documented as a helper!

packages/sdk-browser/src/sdk-client.ts
```diff
@@ -217,15 +234,63 @@ export class TurnkeyBrowserSDK {
...
+
+    const userSession: User = {
+      userId,
+      username,
+      organization: {
+        organizationId,
+        organizationName,
+      },
+      session: {
+        authenticatedClient: this.authClient,
```
andrewkmin, line 288: also open to calling `authClient --> authenticatedClient`. don't feel tooo strongly on this though

packages/sdk-browser/src/sdk-client.ts
```diff
@@ -273,21 +326,25 @@ export class TurnkeyBrowserClient extends TurnkeySDKClientBase {
       userId: userId!,
     });
 
+    // Ensure session and sessionExpiry are included in the object
+    const readWriteSessionResultWithSession = {
+      ...readWriteSessionResult,
+      session: readWriteSessionResult.credentialBundle,
```
andrewkmin, line 332: I wonder if we should rename `session` and `sessionExpiry` given how we renamed it earlier under the current User type. or are we intentionally naming it differently here to reduce confusion?

packages/sdk-browser/src/sdk-client.ts
```diff
@@ -392,23 +453,29 @@ export class TurnkeyPasskeyClient extends TurnkeyBrowserClient {
...
+    const session: ReadWriteSession = {
+      credentialBundle: authBundle,
+      expiry: Date.now() + Number(expirationSeconds) * 1000,
     };
 
-    // store auth bundle in local storage
-    await setStorageValue(StorageKeys.ReadWriteSession, readWriteSession);
+    await this.saveSession({
+      organizationId: organizationId!,
+      organizationName: "",
```
andrewkmin, line 463: is this (and `username`) an empty string by default?

packages/sdk-browser/src/storage.ts
```diff
@@ -4,12 +4,14 @@ import WindowWrapper from "./__polyfills__/window";
 export enum StorageKeys {
   AuthBundle = "@turnkey/auth_bundle", // LEGACY
   CurrentUser = "@turnkey/current_user",
```
andrewkmin, line 6: should we consider this deprecated?

packages/wallet-stamper/README.md
```diff
@@ -140,7 +140,6 @@ const wallets = await client.getWallets({
 The primary difference between signing with an Ethereum Wallet and a Solana Wallet is the process of obtaining the public key.
 For Solana, the public key can be directly derived from the wallet. However, for Ethereum, the secp256k1 public key cannot be directly retrieved.
 Instead, you must first obtain a signature from the user and then recover the public key from that signature.
-This additional step is why the Ethereum Wallet (EVM Wallet) interface includes a different method for recovering the public key.
```
andrewkmin, line 143: is this no longer relevant?

packages/wallet-stamper/src/__tests__/wallet-interfaces.ts
```diff
@@ -51,7 +51,9 @@ export class MockEvmWallet implements EvmWalletInterface {
     return signature;
   }
 
-  async recoverPublicKey(message: string, signature: string): Promise<string> {
+  async getPublicKey(): Promise<string> {
+    const message = "hello from TKHQ!";
```
andrewkmin, line 55: I like this message! :D maybe we should also use `hello from TKHQ!` in the sdk-browser README: `packages/sdk-browser/README.md`

andrewkmin (commented):
loooove the progress!

```
Updated SolanaWalletInterface and EvmWalletInterface
Reason: Streamlines interfaces by removing redundant method declarations

Modified EthereumWallet implementation
Reason: Aligns with new interface structure and utilizes signMessage for public key recovery
```

should we consolidate on EVM or Ethereum, not both?

also, would you be able to add a changeset for all the package changes? feel free to make them major updates. It would be great to get some iterations on getting the changes articulated in the changelog :)

examples/with-wallet-stamper/src/components/turnkey-provider.tsx
taylorjdawson: yea unfortunately, I think we could brain storm about maybe saving the key in local storage or in memory?

packages/sdk-browser/README.md
taylorjdawson: @andrewkmin are you suggesting just renaming the variable?

packages/sdk-browser/README.md
taylorjdawson: yea! I need to update this readme to reflect the new implementation

packages/sdk-browser/src/sdk-client.ts
taylorjdawson: well this is because the function takes the TSessionResponse type so I convert the object to that type before passing it in to saveSession
alternatively I could have the saveSession method take the type of `readWriteSessionResultWithSession` as is 
and differentiate between saving a read session vs a write session, what do you think?

packages/sdk-browser/src/sdk-client.ts
```diff
@@ -375,7 +434,7 @@ export class TurnkeyPasskeyClient extends TurnkeyBrowserClient {
     expirationSeconds: string = DEFAULT_SESSION_EXPIRATION,
     organizationId?: string
   ): Promise<ReadWriteSession> => {
-    const localStorageUser = await getStorageValue(StorageKeys.CurrentUser);
+    const localStorageUser = await getStorageValue(StorageKeys.UserSession);
     userId = userId ?? localStorageUser?.userId;
```
taylorjdawson, line 438: @andrewkmin should we make userId optional given that it could comes from the local storage similar to organizationId?

packages/sdk-browser/src/sdk-client.ts
taylorjdawson: Okay I modified it, it is empty if it is not found in local storage

packages/sdk-browser/src/storage.ts
taylorjdawson: yeah should we remove completely? I don't see why that would cause any issues

andrewkmin: I'm down to just mark as // LEGACY or // DEPRECATED for now

packages/sdk-browser/src/sdk-client.ts
```diff
@@ -27,7 +29,7 @@ import {
   StorageKeys,
   getStorageValue,
   removeStorageValue,
-  setStorageValue,
```
andrewkmin, line 30: part of me feels like we should still expose these helpers to manipulate storage, just in case people want to do that

packages/sdk-browser/src/sdk-client.ts
andrewkmin: ohh I see, that makes sense. I'm fine with this approach in that case!

packages/sdk-browser/src/sdk-client.ts
taylorjdawson: @andrewkmin that is still exposed its exported in the index.ts file

examples/with-wallet-stamper/src/components/turnkey-provider.tsx
taylorjdawson: @andrewkmin oh that is interesting, although it looks like is restricted to Local accounts where the privatekey is stored on the user's machine

packages/crypto/src/crypto.ts
```diff
@@ -568,3 +568,46 @@ export const fromDerSignature = (derSignature: string) => {
   // Concatenate and return the raw signature
   return new Uint8Array([...rPadded, ...sPadded]);
 };
+
+export const toDerSignature = (rawSignature: string) => {
```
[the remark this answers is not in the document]
taylorjdawson: @andrewkmin some sources I checked with (including chatGPT)
- [python-ecdsa](https://github.com/tlsfuzzer/python-ecdsa/blob/master/src/ecdsa/der.py)
- [libbitcoin documentation on ECDSA and DER Signatures](https://github.com/libbitcoin/libbitcoin-system/wiki/ECDSA-and-DER-Signatures)

packages/sdk-browser/src/sdk-client.ts
```diff
@@ -140,13 +151,13 @@ export class TurnkeyBrowserSDK {
    */
   currentUserSession = async (): Promise<TurnkeyBrowserClient | undefined> => {
```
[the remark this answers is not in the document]
taylorjdawson: @andrewkmin if we did this is your idea that we would return the `TurnkeyIframeClient` instead of the `TurnkeyBrowserClient` for read-write sessions?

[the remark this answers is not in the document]
andrewkmin: mm yeah that's tricky, i suppose it would have to be in that case

examples/with-wallet-stamper/src/components/turnkey-provider.tsx
andrewkmin: yep, not sure if it's compatible with external signers but interesting nonetheless

packages/sdk-browser/src/sdk-client.ts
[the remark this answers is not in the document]
taylorjdawson: I think we could abstract this in the react sdk
```
const { client } = useTurnkey()
```
we could make this `client` more intelligent by defaulting to creating a read/write session regardless of which client authenticates initially 
and the user would have the option of downgrading or setting a default auth level for the client in the config

package.json
```diff
@@ -59,7 +59,8 @@
       "fast-xml-parser": ">=4.4.1",
       "semver": ">=6.3.1",
       "path-to-regexp@<0.1.10": ">=0.1.10",
-      "secp256k1": ">=4.0.4"
+      "secp256k1": ">=4.0.4",
+      "cross-spawn": ">=7.0.5 "
```
andrewkmin, line 63: tiny nit: trailing whitespace

examples/with-wallet-stamper/src/components/turnkey-provider.tsx
```diff
@@ -139,7 +130,10 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
   }
 
   async function signInWithWallet(): Promise<User | null> {
-    const publicKey = (wallet as SolanaWalletInterface)?.recoverPublicKey();
+    const publicKey = await wallet?.getPublicKey();
+    if (!publicKey) {
+      return null;
```
andrewkmin, line 135: sort of a meta question but do you prefer using `null`s or `undefined`s?

packages/sdk-react/package.json
```diff
@@ -43,7 +43,9 @@
     "typecheck": "tsc -p tsconfig.typecheck.json"
   },
   "dependencies": {
-    "@turnkey/sdk-browser": "workspace:*"
+    "@turnkey/sdk-browser": "workspace:*",
+    "@turnkey/wallet-stamper": "*",
```
andrewkmin, line 47: should this be `workspace:*`?

packages/sdk-react/src/contexts/TurnkeyContext.tsx
```diff
@@ -1,31 +1,45 @@
...
+  client: TurnkeyBrowserClient | undefined;
   turnkey: Turnkey | undefined;
   authIframeClient: TurnkeyIframeClient | undefined;
   passkeyClient: TurnkeyPasskeyClient | undefined;
+  walletClient: TurnkeyWalletClient | undefined;
   getActiveClient: () => Promise<TurnkeyBrowserClient | undefined>;
 }
 
 export const TurnkeyContext = createContext<TurnkeyClientType>({
+  client: undefined,
```
andrewkmin, line 26: if this is the active client, should we name it `activeClient` for consistency with `getActiveClient`?

packages/wallet-stamper/README.md
```diff
@@ -193,12 +167,16 @@ const whoami = await client.getWhoami({
   organizationId: process.env.NEXT_PUBLIC_ORGANIZATION_ID,
 });
 
-// Now that we have the sub organization id, we can make requests using that sub org id
+if (!whoami?.userId) {
+  // User does not yet have a sub organization, so we need to create one
```
andrewkmin, line 171: should we add suborg creation code here?

packages/wallet-stamper/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@turnkey/wallet-stamper",
-  "version": "0.0.5",
+  "version": "1.0.0",
```
andrewkmin, line 3: should we make this initial launch as beta? i.e. `0.1.0`? or are we comfortable full sending as `1.0.0`? I'm fine with either; if we're feeling good about the integration with demo wallet, let's do `1.0.0` :)

packages/wallet-stamper/src/ethereum.ts
andrewkmin, line 1: lots of great goodies here within the inline code comments themselves. we could link to this specifically within the changelog

packages/wallet-stamper/src/stamper.ts
```diff
@@ -27,16 +32,37 @@ export class WalletStamper implements TStamper {
...
 
     let publicKey: string;
     try {
-      publicKey =
-        this.wallet.type === WALLET_TYPE_SOLANA
-          ? this.wallet.recoverPublicKey()
-          : await this.wallet.recoverPublicKey(payload, signature);
+      // For Ethereum, we need to recover the public key from the signature over the payload.
+      // This is because recovering the SECP256K1 public key requires a signed message.
+      // This avoids the an additional call to the wallet to get the public key.
```
andrewkmin, line 43: tiny nit: let's choose one of `the / an`

andrewkmin: can you explain how we avoid the additional call? wouldn't we need an extra one regardless in order to get that public key?

andrewkmin (commented):
wow a herculean effort 👏 just some relatively minor comments

examples/with-wallet-stamper/src/components/turnkey-provider.tsx
taylorjdawson: Generally I'll use `null` to distinguish between uninitialized variables and values that have been intentionally set to an empty value

packages/sdk-react/src/contexts/TurnkeyContext.tsx
taylorjdawson: yeah that makes sense! I was thinking of simplifying the name to abstract away some of the complexity of the multiple clients

packages/wallet-stamper/src/stamper.ts
taylorjdawson: With solana I can call `getPublicKey` without triggering a signature request but with ethereum we need to derive it from the signature, so instead of calling `getPublicKey` we derive it from the signature over the payload

packages/wallet-stamper/src/ethereum.ts
taylorjdawson: good call!

packages/wallet-stamper/package.json
taylorjdawson: Yeah I was on the fence between beta/major version, I decided the API feels solidified -- I have some ideas for future improvements but they change the core API

packages/wallet-stamper/src/stamper.ts
andrewkmin: OH! right right, that's beautiful!

andrewkmin (approved):

PR #414, opened 2024-11-03
allow timestamp overrides

andrewkmin:
## Summary & Motivation
$title

The main use case this serves is when a client may have an incorrect system time, leading to rejections by our servers. See https://docs.turnkey.com/faq#how-long-is-a-signed-activity-request-valid-for for more details.

## How I Tested These Changes
Examples

## Did you add a changeset?


packages/http/src/shared.ts
```diff
@@ -201,3 +201,16 @@ export function getSignedTransactionFromActivity(
...
+ * A utility function to get a live timestmap directly from Turnkey
+ *
+ * @returns string reprenting timestamp in seconds
+ */
+export async function getLiveTimestamp(): Promise<string> {
+  const timestampResponse = await fetch("https://api.turnkey.com/health");
+  const parsedResponse = await timestampResponse.json();
+
+  return parsedResponse.currentTime ?? "";
+}
```
r-n-o, line 216: This might be okay but I wonder how much extra traffic that's going to cause. This code will live "forever" so I'm a bit nervous about the lack of caching.

An idea that we could implement potentially would be a one-time fetch to "set the clock". Then the local timestamps can still be used as a relative source of time.
* SDK init: fetch api.turnkey.com/health, get timestamp
* compare remote with local time, store delta
* when requests need a timestamp, get local time + delta

That's under the assumption that the user clocks are off by an hour or a day or a year...basically a fixed offset. Basically I'm assuming that their local device can still increment time correctly.

If we still see problems with this logic we can periodically "resync" clocks (refresh the delta between remote and local)

This solution is kind of nice because we start with a single init-time call which calls health: this has the benefit of "syncing" time, but also ensure that our API isn't down. If /health fails it's good to know ahead of time before making any API request (perfect place to insert telemetry reporting btw!)

packages/http/src/shared.ts
```diff
@@ -201,3 +201,16 @@ export function getSignedTransactionFromActivity(
...
 }
+
+/**
+ *
+ * A utility function to get a live timestmap directly from Turnkey
+ *
+ * @returns string reprenting timestamp in seconds
+ */
+export async function getLiveTimestamp(): Promise<string> {
+  const timestampResponse = await fetch("https://api.turnkey.com/health");
```
r-n-o, line 212: I think we should make the hostname overridable or configurable? If I set my turnkey API host to preprod, this would also use preprod health.

r-n-o (commented):
Left a few comments below with ideas, but I'm down to ship what you have if you don't have too much time to sink into this. I'm mostly concerned about the extra load this would generate over time, but...all things considered it might be worth the risk.

PR #424, opened 2024-11-14
Release nov14

moe-dev:
## Did you add a changeset?


examples/kitchen-sink/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
...
+
+### Patch Changes
+
+- Updated dependencies [3dd74ac]
+- Updated dependencies [1e36edf]
+- Updated dependencies [4df8914]
+  - @turnkey/sdk-browser@1.9.0
+  - @turnkey/sdk-server@1.6.0
+  - @turnkey/http@2.16.0
+  - @turnkey/sdk-react@1.0.14
```
r-n-o, line 13: I don't think this is correct, our examples all use workspace versions. We didn't update dependencies, they're already updated :)

packages/crypto/CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
...
+
+- 33e8e03: Reorganize into two subparts:
+
+  - `crypto.ts`: core cryptography utilities
+  - `turnkey.ts`: Turnkey-specific cryptography utilities
+
+  Add `verifyStampSignature` method:
+
+  - See in-line code docs for more details + example of usage
+  - This is useful for checking the validity of a stamp (signature) against the request body
```
r-n-o, line 15: This looks a bit weirdly formatted. Let's link to the PR number (instead of the git sha) and cleanup the bullets so it's one list.

packages/eip-1193-provider/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@turnkey/eip-1193-provider",
-  "version": "3.0.0",
+  "version": "4.0.0",
```
r-n-o, line 3: Where did this come from? This is a major change. And I only see "patch changes" in the corresponding CHANGELOG. Something isn't right.

moe-dev: we've also moved from 2 -> 3 on a patch change in the past, should we mess with the autogen?

PR #425, opened 2024-11-14
Allow overriding WebauthnStamper config when creating a passkey client

r-n-o:
## Summary & Motivation
This looks like an oversight. We allow for RPID override but do not let users customize timeout/userVerification/allowCredentials.


moe-dev (approved):
Good catch!

PS: wicked impalas relax

PR #432, opened 2024-11-26
section 1

omkarshanbhag:
## Did you add a changeset?

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,61 @@
 
 [![js-build](https://github.com/tkhq/sdk/actions/workflows/js-build.yml/badge.svg)](https://github.com/tkhq/sdk/actions/workflows/js-build.yml)
 
+## Primary Turnkey SDK Web Packages
+SDK Web Packages
+The following packages expose the functionality required to build Turnkey-Powered applications. 
+- [`@turnkey/sdk-browser`](https://github.com/tkhq/sdk/tree/main/packages/sdk-browser) 
+- [`@turnkey/sdk-server`](https://github.com/tkhq/sdk/tree/main/packages/sdk-server)
+- [`@turnkey/sdk-react`](https://github.com/tkhq/sdk/tree/main/packages/sdk-react)
+- [`@turnkey/sdk-react-native-stamper`](https://www.npmjs.com/package/@turnkey/sdk-react-native-stamper)
+
+The below diagram can help provide an understanding of which package can be used where while developing your Turnkey Powered Application. 
+
+## [DIAGRAM HERE]
+
+### [`@turnkey/sdk-browser`](https://github.com/tkhq/sdk/tree/main/packages/sdk-browser)  
+The [`@turnkey/sdk-browser`](https://github.com/tkhq/sdk/tree/main/packages/sdk-browser) package exposes functionality that lets developers build browser based applications that interact with the Turnkey API with different types of authentication. 
+
+It consists of different clients that enable requests to the API to be authenticated via different auth methods like user sessions, passkeys and iFrames. It also contains methods to manage information and state related to authentication like auth bundles and sessions, retrieving user information and server signing API requests.
+
+[![npm](https://img.shields.io/npm/v/@turnkey/http?color=%234C48FF)](https://www.npmjs.com/package/@turnkey/sdk-browser)
+
+[CHANGELOG](/packages/sdk-browser/CHANGELOG.md)  
+
+Docs Page: https://docs.turnkey.com/sdks/javascript-browser
+
+### [`@turnkey/sdk-server`](https://github.com/tkhq/sdk/tree/main/packages/sdk-server)
+The [`@turnkey/sdk-server`](https://github.com/tkhq/sdk/tree/main/packages/sdk-server) package exposes functionality that lets developers build server-side functionality for applications that interact with the Turnkey API with different types of authentication –  allowing applications to authenticate users, manage sessions, and perform organizational operations securely and efficiently. It consists of an API Client and API Proxies that enable requests to the Turnkey API to be authenticated with the appropriate credentials. Specifically, the API Client manages requests signed by the user’s authentication details, and the API proxies handle requests signed by the parent organization’s authentication details. 
+
+Use the [`@turnkey/sdk-server`](https://www.npmjs.com/package/@turnkey/sdk-server) package to handle server-side interactions for applications that interact with the Turnkey API.
+
+[![npm](https://img.shields.io/npm/v/@turnkey/sdk-server?color=%234C48FF)](https://www.npmjs.com/package/@turnkey/sdk-server)  
+
+[CHANGELOG](/packages/sdk-server/CHANGELOG.md)
+
+Docs Page: https://docs.turnkey.com/sdks/javascript-server
+
+### [`@turnkey/sdk-react`](https://github.com/tkhq/sdk/tree/main/packages/sdk-react)
+The [`@turnkey/sdk-react`](https://github.com/tkhq/sdk/tree/main/packages/sdk-react) package wraps the functionality from the [`@turnkey/sdk-browser`](https://github.com/tkhq/sdk/tree/main/packages/sdk-browser) package to allow developers to build react based applications that interact with the Turnkey API with different types of authentication. It allows developers to use the same clients exposed in [`@turnkey/sdk-browser`](https://github.com/tkhq/sdk/tree/main/packages/sdk-browser) that enable requests to the API to be authenticated via different auth methods like user sessions, passkey and iFrames. It also contains the same methods to manage information and state related to authentication like auth bundles and sessions, retrieving user information and server signing API requests. Use the [`@turnkey/sdk-react`](https://github.com/tkhq/sdk/tree/main/packages/sdk-browser) package when you’re building React-based frontend applications that interact with the Turnkey API.
+
+[![npm](https://img.shields.io/npm/v/@turnkey/sdk-react?color=%234C48FF)](https://www.npmjs.com/package/@turnkey/sdk-react)
+
+[CHANGELOG](/packages/sdk-react/CHANGELOG.md)
+
+Docs Page: https://docs.turnkey.com/sdks/react
+
+### [`@turnkey/sdk-react-native-stamper`](https://www.npmjs.com/package/@turnkey/sdk-react-native-stamper)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,60 +2,106 @@
...
+## Overview
+
+The Turnkey SDK includes functions and clients with methods that construct, authenticate and execute requests to the Turnkey API with support for different ecosystems and methods of authentication. It consists of the [Primary Turnkey Web SDK Packages](#primary-turnkey-web-sdk-packages) which expose the main functionality required to build Turnkey-powered applications in different web environments, the [Chain/Ecosystem-Specific Signing Packages](#chainecosystem-specific-signing-sdk-packages) which expose signers with support for specific ecosystems, built on top of our Web SDK, and the [Advanced Functionality SDK Packages](#advanced-functionality-sdk-packages) which exposes lower level functionality that is leveraged by our Primary Web SDK Packages for those with highly-specific implementations looking to use them.
+
+The below diagram helps visualize the packages in our SDK organized by the functionality they expose.
```
r-n-o, line 9: Nit: "the diagram below" is more common.

README.md
```diff
@@ -2,60 +2,106 @@
...
+## Overview
+
+The Turnkey SDK includes functions and clients with methods that construct, authenticate and execute requests to the Turnkey API with support for different ecosystems and methods of authentication. It consists of the [Primary Turnkey Web SDK Packages](#primary-turnkey-web-sdk-packages) which expose the main functionality required to build Turnkey-powered applications in different web environments, the [Chain/Ecosystem-Specific Signing Packages](#chainecosystem-specific-signing-sdk-packages) which expose signers with support for specific ecosystems, built on top of our Web SDK, and the [Advanced Functionality SDK Packages](#advanced-functionality-sdk-packages) which exposes lower level functionality that is leveraged by our Primary Web SDK Packages for those with highly-specific implementations looking to use them.
```
r-n-o, line 7: I think this overview would benefit from being broken down. The first sentence is A LOT to process for a first time reader. It's long.

The way I'd break it down is with bullets. Strawman below

-----

The Turnkey SDK includes functionality to interact with Turnkey in various contexts and ecosystems. It consists of three main NPM package groups:
* the [Primary Turnkey Web SDK Packages](#primary-turnkey-web-sdk-packages) which expose the main functionality required to build Turnkey-powered applications in different web environments
* the [Chain/Ecosystem-Specific Signing Packages](#chainecosystem-specific-signing-sdk-packages) which expose signers with support for specific ecosystems, built on top of our Web SDK packages
* the [Advanced Functionality SDK Packages](#advanced-functionality-sdk-packages) which exposes lower level functionality that is leveraged by our Primary Web SDK Packages for those with highly-specific implementations looking to use them.

README.md
```diff
@@ -2,60 +2,106 @@
...
+
+The below diagram helps visualize the packages in our SDK organized by the functionality they expose.
+
+<img src="./img/sdk-map.png" alt="homepage screenshot" width="1000px">
```
r-n-o, line 11: I like this diagram a lot but I'd change a couple of things:
* `@turnkey/react-native-passkey-stamper` isn't a "main" package. It's just a stamper so it should be moved next to `@turnkey/webauthn-stamper` and `@turnkey/api-key-stamper`
* The arrows need labeling. I think the vertical arrow should go down and read "uses" or "imports". And the horizontal arrow can be bidirectional and read "integrates with"? Ecosystem packages generally consumes our other packages, but sometimes it's the other way around

There's also a marketing opportunity here with ecosystem/blockchain logos next to each ecosystem package (def nice-to-have!)

README.md
```diff
@@ -2,60 +2,106 @@
...
+
+## Primary Turnkey Web SDK Packages
+
+The following packages expose the main functionality required to build Turnkey-powered applications. Each package exposes functions, and/or client classes with methods that manage the process of authenticating requests to the Turnkey API in the contexts of a generic browser environment, a react or react native client environment, or a server environment.
```
r-n-o, line 15: we do not have a react native client. We simply use the "normal" clients. Passkey stamping has to involve magic native APIs which is why we have a separate package for it, but HTTP calls into Turnkey work the same as on the web.

README.md
```diff
@@ -2,60 +2,106 @@
...
+
+The below diagram helps visualize how each package can be used to devlop the appropriate service in your Turnkey Powered Application, and how Turnkey requests would flow between those services.
```
r-n-o, line 28: "diagram below"

README.md
```diff
@@ -2,60 +2,106 @@
...
-| [`with-federated-passkeys`](/examples/with-federated-passkeys/)        | A NextJS app that demonstrates how to use Turnkey to build a federated, webauthn powered authentication flow                            |
-| [`with-solana-passkeys`](/examples/with-solana-passkeys/)              | A NextJS app powering users to create suborgs, sign messages, and create transactions sponsored by the parent org using @turnkey/solana |
-| [`with-eth-passkey-signers`](/examples/with-eth-passkey-signers/)      | A NextJS app powering users to create suborgs and sign messages via Viem or Ethers                                                      |
-| [`with-eip-1193-provider`](/examples/with-eip-1193-provider/)          | A NextJS app that demonstrates how to use Turnkey the `@turnkey/eip-1193-provider` in your app                                          |
-| [`with-wallet-stamper`](/examples/with-wallet-stamper/)                | A NextJS app that demonstrates how to use Turnkey the `@turnkey/wallet-stamper` in your app                                             |
+### Best First Examples -- Test in Stackblitz
```
r-n-o, line 67: I'd call this section "Instant examples (Stackblitz)"

README.md
```diff
@@ -2,60 +2,106 @@
...
+| [`with-eth-passkey-signers`](/examples/with-eth-passkey-signers/) | A NextJS app powering users to create suborgs and sign messages via Viem or Ethers                                                      | [Test it out on Stackblitz!](https://stackblitz.com/edit/stackblitz-starters-2psu3g) |
+| [`with-solana`](/examples/with-solana/)                           | Create a new Solana address, then sign and broadcast a transaction on Solana's devnet                                                   | [Test it out on Stackblitz!](https://stackblitz.com/edit/stackblitz-starters-xeb93i) |
+| [`with-solana-passkeys`](/examples/with-solana-passkeys/)         | A NextJS app powering users to create suborgs, sign messages, and create transactions sponsored by the parent org using @turnkey/solana | [Test it out on Stackblitz!](https://stackblitz.com/edit/stackblitz-starters-h5pmnu) |
+
+### Other Code Examples
+
```
r-n-o, line 79: (optional) might benefit from a word or two about the fact that these examples require a local installation of node + following instruction in linked READMEs? (since you have an intro in the stackblitz examples section)

README.md
andrewkmin: +1; I think the "Advanced Functionality" label might imply a higher order, when in reality the hierarchy is such that the stampers are prereqs for all the abstractions built atop them (conveying this hierarchy via arrows/etc is huge so thanks for approaching this in such a way). Stampers + `@turnkey/http` could be considered Primitives/foundational packages

Technically, `@turnkey/http` isn't quite a stamper but rather a library containing an http client that allows for (easier) interaction with Turnkey endpoints. And I think the utilities SDK packages could be in a separate bucket of their own.

README.md
omkarshanbhag: This makes sense -- appreciate the color here! Had the same intuition because I noticed that it was just a stamper and that we just say to use the stamper along with the basic http client from @turnkey/http.


I ended up grouping it with the other larger libraries because it gets its own page in our docs after sdk-browser, sdk-server and sdk-react but definitely agree that it should go with the stampers and is more analogous with them

README.md
```diff
@@ -2,60 +2,106 @@
...
+| [@turnkey/sdk-server](https://github.com/tkhq/sdk/tree/main/packages/sdk-server)                                     | The `@turnkey/sdk-server` package exposes functionality that lets developers build server-side functionality for applications that interact with the Turnkey API with different types of authentication – allowing applications to authenticate users, manage sessions, and perform organizational operations securely and efficiently. It consists of an API Client and API Proxies that enable requests to the Turnkey API to be authenticated with the appropriate credentials. Specifically, the API Client manages requests signed by the user's authentication details, and the API proxies handle requests signed by the parent organization's authentication details. Use this package to handle server-side interactions for applications that interact with the Turnkey API. | [![npm](https://img.shields.io/npm/v/@turnkey/sdk-server?color=%234C48FF)](https://www.npmjs.com/package/@turnkey/sdk-server)                                     | [CHANGELOG](/packages/sdk-server/CHANGELOG.md)                   | [Docs](https://docs.turnkey.com/sdks/javascript-server)  |
```
andrewkmin, line 24: > The `@turnkey/sdk-server` package exposes functionality that lets developers build server-side functionality for applications that interact with the Turnkey API with different types of authentication – allowing applications to authenticate users, manage sessions, and perform organizational operations securely and efficiently. It consists of an API Client and API Proxies that enable requests to the Turnkey API to be authenticated with the appropriate credentials. Specifically, the API Client manages requests signed by the user's authentication details, and the API proxies handle requests signed by the parent organization's authentication details. Use this package to handle server-side interactions for applications that interact with the Turnkey API.

The upshot of `@turnkey/sdk-server` is that it allows people to use Turnkey with an API key credential (P256 keypair that can be created using our `@turnkey/crypto` library)

img/sdk-web-diagram.png
andrewkmin, line 1: maybe we can say `Turnkey requests to be (optionally) proxied through Backend`? given not all setups will require a proxy setup

andrewkmin: `End User's sub-org`

andrewkmin: I hesitate to include `@turnkey/sdk-react-native-passkey-stamper` alongside `@turnkey/sdk-browser` and `@turnkey/sdk-react` given it's specifically for stamping only, whereas the other two are a bit more wholistic

omkarshanbhag: some further color on the proxy requests -- def agree that its optional and that its not an expectation. The reason I explicitly included it like that is to show all flows of Turnkey requests between App Client/App Server and the Turnkey API. Agree that including the optionally will help make that clear.

Agree on the other two points as well!

README.md
omkarshanbhag: @r-n-o On the logos -- i put some of them up but it looks a bit lopsided because some of these are actually related to a chain ecosystem with a logo like solana or ethers, but viem for example or eip stamper etc would either also have ethereum logos or none at all. For now I think it would be best to move forward without the logos! Lmk if you have strong opinions otherwise.

r-n-o: I'm down to move forward without. Makes sense 👍

r-n-o (approved):
👏

PR #444, opened 2024-12-09
implement message channel

turnekybc:
## Did you add a changeset?


packages/sdk-react/src/contexts/TurnkeyContext.tsx
```diff
@@ -123,7 +123,7 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
           iframeContainer: document.getElementById(
             TurnkeyAuthIframeContainerId
           ),
-          iframeUrl: "https://auth.turnkey.com",
+          iframeUrl: config.iframeUrl || "https://auth.turnkey.com",
```
r-n-o, line 126: 👏

packages/sdk-browser/src/__types__/base.ts
```diff
@@ -91,6 +91,7 @@ export interface TurnkeySDKBrowserConfig {
   defaultOrganizationId: string;
   rpId?: string;
   serverSignUrl?: string;
+  iframeUrl?: string;
```
r-n-o, line 94: because you're making a change here you'll need to generate a changeset for this PR so that we have a minor version change + changelog entry for sdk-browser

packages/iframe-stamper/src/index.ts
```diff
@@ -109,6 +109,8 @@ export class IframeStamper {
   iframe: HTMLIFrameElement;
   iframeOrigin: string;
   iframePublicKey: string | null;
+  trustedParentOrigin: string;
+  messageChannel: MessageChannel;
```
r-n-o, line 113: Would be nice to explain what these are with a quick comment? Something like:
```js
// used to template the iframe and ensure that iframe events are received by the right (parent) frame, for security reasons.
```

packages/iframe-stamper/src/index.ts
```diff
@@ -132,9 +134,11 @@ export class IframeStamper {
 
     let iframe = window.document.createElement("iframe");
 
+    this.trustedParentOrigin = window.origin;
     // See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#sandbox
     // We do not need any other permission than running scripts for import/export/auth frames.
     iframe.setAttribute("sandbox", "allow-scripts allow-same-origin");
+    iframe.setAttribute("data-trusted-origin", this.trustedParentOrigin);
```
r-n-o, line 141: is there a need to persist `window.origin` on the iframeStamper or can we do `iframe.setAttribute("data-trusted-origin", window.origin)` directly?

packages/iframe-stamper/src/index.ts
```diff
@@ -132,9 +134,11 @@ export class IframeStamper {
 
     let iframe = window.document.createElement("iframe");
 
+    this.trustedParentOrigin = window.origin;
```
r-n-o, line 137: A thought I'm having is that this won't necessarily work in react native (what would "window.origin" resolve to?). But I think that's fine because RN apps don't use iframe stampers do they?

Another thing worth checking is TG apps. But again, they probably don't use iframe stampers?

packages/iframe-stamper/src/index.ts
```diff
@@ -145,29 +149,29 @@ export class IframeStamper {
...
   }
 
   /**
    * Inserts the iframe on the page and returns a promise resolving to the iframe's public key
    */
   async init(): Promise<string> {
     this.container.appendChild(this.iframe);
+    this.iframe.addEventListener("load", () => {
+      // Send a message to the iframe to initialize the message channel
+      this.iframe.contentWindow?.postMessage("init", this.iframeOrigin, [
```
r-n-o, line 163: why can `contentWindow` be null or undefined? Should we throw an error if `this.iframe.contentWindow` is falsy?

packages/iframe-stamper/src/index.ts
```diff
@@ -145,29 +149,29 @@ export class IframeStamper {
...
+        this.messageChannel.port2,
```
r-n-o, line 164: From the MDN docs: "The ownership of these objects [in the transfer array] is given to the destination side and they are no longer usable on the sending side".

So we're essentially giving the iframe the "port2" side of the pipe? Maybe that's worth a comment I'm not sure.

Also: should we do any feature detection at instantiation time to make sure MessageChannel is available?

packages/iframe-stamper/src/index.ts
r-n-o, line 163: Another comment here. Maybe this message should be called "initChannel" or "INIT_CHANNEL" just so it's clear?

r-n-o (commented):
I really like this `MessageChannel`-based solution, it's way better than the global listeners/handlers we had previously! 🚀 

If we have channels in place between parent page and iframes we can completely disregard the notion "trusted origin".  The channel **is opened by a trusted origin**, almost by definition.

The ambiguity which existed before ("is the iframe receiving a message from its parent or from another frame?" and "is the parent receiving from its child iframe or from another frame?"()  doesn't exist anymore because the MessageChannel is privately owned by the iframe stamper:
* script on the parent page or other frames can't write to port1. This means the iframe knows any messages on port2 are from the right sender: the frame which instantiated it 🔒 
* other frames can't write to port2 because it's owned by the iframe. This means the iframe stamper is guaranteed that messages on port1 are from the iframe it instantiated, and nobody else 🔒

packages/iframe-stamper/src/index.ts
turnekybc: Updated this to be `TURNKEY_INIT_MESSAGE_CHANNEL`

packages/iframe-stamper/src/index.ts
turnekybc: Added more comments and links to the MDN docs explaining exactly what's going on.

packages/iframe-stamper/src/index.ts
```diff
@@ -185,6 +208,33 @@ export class IframeStamper {
...
+   */
+  addMessageHandler(): Promise<any> {
+    return new Promise((resolve, reject) => {
+      this.messageChannel.port1.onmessage = (event) => {
+        this.onMessageHandler(event, resolve, reject);
+      };
+    });
+  }
+
+  onMessageHandler(event: MessageEvent, resolve: any, reject: any): void {
```
r-n-o, line 222: clean! ⭐

packages/iframe-stamper/src/index.ts
```diff
@@ -193,31 +243,13 @@ export class IframeStamper {
...
-        },
-        false
-      );
+      this.messageChannel.port1.postMessage({
+        type: IframeEventType.InjectCredentialBundle,
+        value: bundle,
+      });
+      this.messageChannel.port1.onmessage = (event) => {
+        this.onMessageHandler(event, resolve, reject);
+      };
```
r-n-o, line 252: I don't think you need the wrapping promise here? How about modifying this function to be:

```ts
this.messageChannel.port1.postMessage({
    type: IframeEventType.InjectCredentialBundle,
    value: bundle,
});

return addMessageHandler();
```

(looks like you've done that for the other functions, `injectCredentialBundle` is the odd one out!)

packages/iframe-stamper/src/index.ts
```diff
@@ -233,35 +265,14 @@ export class IframeStamper {
     organizationId: string,
     keyFormat?: KeyFormat
   ): Promise<boolean> {
```
r-n-o, line 267: Not sure why TS isn't complaining about this, but I believe there's a type mismatch? The return type is `Promise<boolean>`, but `addMessageHandler` returns `Promise<any>`.

packages/iframe-stamper/src/index.ts
turnekybc: In theory contentWindow should never be `null` or `undefined` but I always add optional chaining to short circuit and prevent an exception. Instead of this failing silently I will throw an exception here

packages/iframe-stamper/src/index.ts
r-n-o: Talked through this on Slack: while odd indeed, we'd break the interface if we made this function synchronous, because it means callers would have to remove `await`s. So, it'd be a cleaner interface, but we're choosing to keep it the way it is for compatibility reasons. Let's not ask people to change their code, there's no real benefit here.

packages/iframe-stamper/src/index.ts
turnekybc: I removed this entirely in favor of the MessageChannel implementation

r-n-o (approved):

packages/iframe-stamper/src/index.ts
turnekybc: Nope. Removed this in favor of the MessageChannel approach.

r-n-o (approved):

packages/iframe-stamper/src/index.ts
```diff
@@ -145,36 +155,58 @@ export class IframeStamper {
...
+     * The iframe will use this port to send messages back to the parent page.
+     * See https://developer.mozilla.org/en-US/docs/Web/API/MessagePort/postMessage#transfer
+     */
+    this.iframe.addEventListener("load", () => {
+      if (
+        !this.iframe.contentWindow ||
+        !this.iframe.contentWindow.postMessage
+      ) {
+        throw new Error(
+          "contentWindow or contentWindow.postMessage does not exist"
```
andrewkmin, line 183: is there a scenario where we'd want granularity over this? in other words, is it particularly weird if `.postMessage` specifically doesn't exist?

packages/iframe-stamper/src/index.ts
```diff
@@ -145,36 +155,58 @@ export class IframeStamper {
...
+
+      this.iframe.contentWindow.postMessage(
+        { type: IframeEventType.TurnkeyInitMessageChannel },
+        this.iframeOrigin,
+        [this.messageChannel.port2]
       );
     });
+
+    return new Promise((resolve, _reject) => {
+      this.messageChannel.port1.onmessage = (event) => {
```
andrewkmin, line 195: non-blocking: a quick note on the importance of the port1 (might be trivial given you've explained port2) could be helpful

packages/iframe-stamper/src/index.ts
andrewkmin: Sorry if I'm missing something obvious — what is the difference in interface between the two? (`injectCredentialBundle` vs `injectKeyExportBundle`), i.e. why is `injectCredentialBundle` the odd one out?

packages/sdk-browser/package.json
```diff
@@ -13,7 +13,7 @@
   },
   "types": "./dist/index.d.ts",
   "license": "Apache-2.0",
-  "description": "Javascript Browser SDK",
+  "description": "JavaScript Browser SDK",
```
andrewkmin, line 16: 🙏

packages/iframe-stamper/src/index.ts
turnekybc: So TypeScript complains if you do `this.iframe.contentWindow.postMessage(...)` because the contentWindow is possibly null. I'm not really aware of a scenario where this would actually be the case in the wild unless there was an underlying bug in the browser itself. `contentWindow` is read-only so you wouldn't be able to null it out if you tried. 

This check is more of a defensive coding pattern in case something is really broken.

packages/iframe-stamper/src/index.ts
turnekybc: `injectCredentialBundle` returns a promise which wraps both the `postMessage` and `onmessage` logic:

```
    return new Promise((resolve, reject) => {
      this.messageChannel.port1.postMessage({
        type: IframeEventType.InjectCredentialBundle,
        value: bundle,
      });
      this.messageChannel.port1.onmessage = (event) => {
        this.onMessageHandler(event, resolve, reject);
      };
    });
```

`injectKeyExportBundle` calls `postMessage` and then returns a promise which handles the `onmessage` logic:

```
    this.messageChannel.port1.postMessage({
      type: IframeEventType.InjectKeyExportBundle,
      value: bundle,
      keyFormat,
      organizationId,
    });

    return this.addMessageHandler();
```

packages/iframe-stamper/src/index.ts
turnekybc: I added a comment below in the constructor for what the MessageChannel is doing. Let me know if that's sufficient.

packages/iframe-stamper/src/index.ts
andrewkmin: 100% agree with being defensive here. was just curious. i'm guessing having this in the wild is a pretty rare occurrence

turnekybc: Yeah I think the internet would be broken

andrewkmin (approved):
🚢

PR #446, opened 2024-12-11
Allow token transfer -- SPL token policy example

omkarshanbhag:
## Summary & Motivation

This PR creates an example that highlights the flow of using our policy engine's support for SPL token transfers! Specifically we show how to allow a user (non root user) to send a Solana SPL token transfer by calculating the "associated token address" of the receiving account, given the mint address of the token that we are transferring. 

The steps included are 
1. Setup - this step creates the on-chain state and SOME of the Turnkey setup state required to make this transfer (notably, it leaves out the creation of the policy that allows the created non root user to sign the SPL transfer)
2. Attempt Transfer - since we're attempting the transfer WITHOUT having created the policy allowing it, this will fail 
3. Create Token Policy - this will create a policy that allows Solana transactions initiated by the correct user to be signed if they contain a single instruction which is an SPL token transfer to the correct token account
4. Attempt Transfer (again) - this will succeed, now that the appropriate policy has been created!

## How I Tested These Changes

Tested the example multiple times

## Did you add a changeset?


andrewkmin (commented):
Looking 🔥 thus far. In terms of next steps:
- minor: feature needs a changeset (`pnpm run changeset`)
- hooking it up with an example in `with-solana` would be amazing
- combining with https://github.com/tkhq/sdk/blob/main/examples/kitchen-sink/src/sdk-server/createPolicy.ts to compute a token address + create a policy based on it would be doubly amazing

stoked about this

omkarshanbhag:
@andrewkmin couple notes here!! 

Given that the associated token address calculating method is readily available by from Solana, I think creating a wrapper function for it and exposing may be less helpful/ a point of confusion/extra weight for us. 

I think what's most helpful is a full example of how to use this with `createPolicy` rather than true SDK support for this. 

Same with combining this with `createPolicy` somehow because there are a variety of policies they could possibly create so don't want to be overly prescriptive there. 

I've created an extensive example with instructions that I think is most helpful in helping folks understand how to best use our SPL support. Let me know if this makes sense overall.

examples/with-solana/README.md
```diff
@@ -166,7 +166,138 @@ Token balance for warchest: 0.0001
 
 Enjoy!
 
-### 6/ Stackblitz Example
+### 6/ Running the "Allow SPL token transfer policy" example
```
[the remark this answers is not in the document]
andrewkmin: instead of `Allow`, let's name it `Create SPL token transfer policy` because the policy can be configured to allow or deny based on parameters

examples/with-solana/README.md
```diff
@@ -166,7 +166,138 @@ Token balance for warchest: 0.0001
...
+
+To best see how policies can be used effectively, we will go through example in the following steps
+
+1. Setup - this step creates the on-chain state and SOME of the Turnkey setup state required to make this transfer (notably, it leaves out the creation of the policy that allows the created non root user to sign the SPL transfer)
+2. Attempt Transfer - since we're attempting the transfer WITHOUT having created the policy allowing it, this will fail
+3. Create Token Policy - this will create a policy that allows Solana transactions initiated by the correct user to be signed if they contain a single instruction which is an SPL token transfer to the correct token account
+4. Attempt Transfer (again) - this will succeed, now that the appropriate policy has been created!
+
+NOTE: An very key piece of understanding to glean from this example is the use of "associated token addresses (or "associated token account address") in the creation of SPL related policies. For further context on ATA's look here: https://spl.solana.com/associated-token-account
```
andrewkmin, line 180: tiny nit: `A`

examples/with-solana/README.md
```diff
@@ -166,7 +166,138 @@ Token balance for warchest: 0.0001
...
+
+```bash
+? Enter Mint account address of token being transferred (created during setup stage):
+```
+
+This should successfully create your policy!
+
+```bash
+New policy created!
+- Name: Let non root user send SPL transfers to the ATA of WARCHEST: 6nDvKk6emwskFLtEpbQgFX6rsMtzbNY4LkvaNnzkvBaq
```
andrewkmin, line 283: should we specify which SPL token we're working with given ATA's are token-specific?

examples/with-solana/package.json
```diff
@@ -6,6 +6,7 @@
     "start": "pnpm tsx src/index.ts",
     "advanced": "pnpm tsx src/advanced.ts",
     "token-transfer": "pnpm tsx src/tokenTransfer.ts",
+    "allow-token-transfer": "pnpm tsx src/allowTokenTransferPolicy.ts",
```
andrewkmin, line 9: maybe a more generic/overarching name for this could be `token-transfer-with-policy`

examples/with-solana/src/requests/createPolicy.ts
```diff
@@ -0,0 +1,51 @@
...
+  consensus: string,
+  condition: string
+): Promise<string> {
+  try {
+    const { policyId } = await turnkeyClient.createPolicy({
+      policyName,
+      condition,
+      consensus,
+      effect,
+      notes: "",
```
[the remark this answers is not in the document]
omkarshanbhag: will do!! -- any particular reason though? I dont think this example needs to use it (will still do it in this round of revisions, just wanted to see if you had something in mind.

examples/with-solana/README.md
```diff
@@ -166,7 +166,138 @@ Token balance for warchest: 0.0001
...
+Token Mint public key: FA7umztm6eYcENE22ndEZhd8MCf9C7myJ2KjfCwobj8F
+Non root user created with user id: 13c4609b-8818-40be-aeac-c8b63da0de4a
+```
+
+#### Step 2. Attempt to Transfer Tokens (This is expected to FAIL)
+
+We will now attempt to transfer these SPL tokens that we've created
+
+```bash
+$ pnpm token-transfer-policy attempt_transfer
```
andrewkmin, line 231: tiny nit: wanna use kebab case here `attempt_transfer` for consistency?

omkarshanbhag: yessir, done

examples/with-solana/README.md
```diff
@@ -166,7 +166,138 @@ Token balance for warchest: 0.0001
 
 Enjoy!
 
-### 6/ Stackblitz Example
+### 6/ Running the "Create SPL token transfer policy" example
```
andrewkmin, line 169: tiny nit: the reason I would include the `with` preposition here is that this example contains more than just creating a policy — we're creating a policy and then using it in practice :)

examples/with-solana/README.md
andrewkmin: did you mean to resolve this one? you could also mention that we're allowing specifically the transfer to the ATA of the warchest wallet associated with the newly created token

andrewkmin (approved):
looks good overall, just some minor comments consider~!

andrewkmin (approved):

PR #456, opened 2024-12-18
Adds postcss-import plugin + fixes import for react phone styles

taylorjdawson:
### CSS Import Rules Error

After importing `import '@turnkey/sdk-react/styles';` in the `layout.tsx` file and running `npm run dev`, an error occurred. When bundling CSS with Rollup using `rollup-plugin-postcss`, `@import` rules are not being correctly placed at the top of the final CSS bundle. This violates CSS specifications, which require `@import` statements to precede all other rules (except `@charset` and `@layer`).

#### Solution

Add `postcss-import` plugin to the PostCSS configuration to the Rollup setup.

The `postcss-import` plugin processes `@import rules` and ensures they appear at the top of the CSS file. By including this in the PostCSS plugins array, all `@import` rules are handled correctly before other CSS rules are processed, preventing build errors.

```bash
@import rules must precede all rules aside from @charset and @layer statements at [project]/node_modules/@turnkey/sdk-react/dist/styles.esm.css:0:6309
```

#### Additional Build Error with postcss-import

When using `postcss-import`, I encountered a build error related to importing CSS from external packages:

```bash
postcss-import: Failed to find 'react-international-phone/style.css'
```

This occurs because `postcss-import` cannot resolve the CSS import from the `node_modules` directory when it's imported within a CSS file.

##### Solution

Instead of importing the styles in `PhoneInput.css`:

```css
@import "react-international-phone/style.css";
```

Move the import to the React component file (`PhoneInput.tsx`) directly:

```tsx
import "react-international-phone/style.css";
```

This approach allows the bundler to correctly resolve the CSS import from `node_modules`, fixing the build error.

## Did you add a changeset?

### DIFF
--- a/packages/sdk-react/rollup.config.mjs
+++ b/packages/sdk-react/rollup.config.mjs
@@ -1,47 +1,54 @@
-import typescript from "@rollup/plugin-typescript";
-import nodeExternals from "rollup-plugin-node-externals";
-import path from "node:path";
-import postcss from "rollup-plugin-postcss";
-import preserveDirectives from "rollup-preserve-directives";
-import url from "@rollup/plugin-url";
-import alias from "@rollup/plugin-alias";
-import copy from "rollup-plugin-copy";
+import typescript from '@rollup/plugin-typescript';
+import nodeExternals from 'rollup-plugin-node-externals';
+import path from 'node:path';
+import postcss from 'rollup-plugin-postcss';
+import preserveDirectives from 'rollup-preserve-directives';
+import url from '@rollup/plugin-url';
+import alias from '@rollup/plugin-alias';
+import copy from 'rollup-plugin-copy';
+import postcssImport from 'postcss-import';
 
 const getFormatConfig = (format) => {
-  const pkgPath = path.join(process.cwd(), "package.json");
+  const pkgPath = path.join(process.cwd(), 'package.json');
   const __dirname = path.dirname(new URL(import.meta.url).pathname);
 
   return {
     input: 'src/index.ts',
     output: {
       format,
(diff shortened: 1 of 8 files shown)


packages/sdk-react/src/components/auth/Auth.tsx
```diff
@@ -60,7 +64,7 @@ interface AuthProps {
     facebookEnabled: boolean;
     googleEnabled: boolean;
   };
-  configOrder: string[];
+  configOrder: AuthSection[];
```
moe-dev, line 67: let's make these changes in another PR - this breaks example build since this type is not exported and will not be consistent with our docs at the moment

moe-dev (approved):

turnekybc (approved):

packages/sdk-react/rollup.config.mjs
```diff
@@ -1,47 +1,54 @@
-import typescript from "@rollup/plugin-typescript";
-import nodeExternals from "rollup-plugin-node-externals";
-import path from "node:path";
-import postcss from "rollup-plugin-postcss";
-import preserveDirectives from "rollup-preserve-directives";
-import url from "@rollup/plugin-url";
-import alias from "@rollup/plugin-alias";
-import copy from "rollup-plugin-copy";
+import typescript from '@rollup/plugin-typescript';
```
andrewkmin, line 1: would expect these to be double quotes

moe-dev: from all the package READMES it shows single quotes actually. It seems like we were incorrect before

PR #462, opened 2024-12-19
Update common error message

zkharit:
## Summary & Motivation
Update a common error message that provides little info as to the root cause
## Did you add a changeset?

### DIFF
--- a/packages/api-key-stamper/src/tink/elliptic_curves.ts
+++ b/packages/api-key-stamper/src/tink/elliptic_curves.ts
@@ -172,7 +172,7 @@ export function pointDecode(point: Uint8Array): JsonWebKey {
     point.length !== uncompressedLength
   ) {
     throw new Error(
-      "Invalid API: Ensure that you are using a valid public and private key for your API key"
+      "Invalid API key: Ensure that you are using a valid public and private key for your API key"
     );
   }
   // Decodes point if its length and first bit match the compressed format
--- a/packages/api-key-stamper/src/tink/elliptic_curves.ts
+++ b/packages/api-key-stamper/src/tink/elliptic_curves.ts
@@ -172,7 +172,7 @@ export function pointDecode(point: Uint8Array): JsonWebKey {
     point.length !== uncompressedLength
   ) {
     throw new Error(
-      "Invalid API key: Ensure that you are using a valid public and private key for your API key"
+      "Invalid length: point is not in compressed or uncompressed format"
     );
   }
   // Decodes point if its length and first bit match the compressed format
--- a/packages/api-key-stamper/src/tink/elliptic_curves.ts
+++ b/packages/api-key-stamper/src/tink/elliptic_curves.ts
@@ -172,7 +172,7 @@ export function pointDecode(point: Uint8Array): JsonWebKey {
     point.length !== uncompressedLength
   ) {
     throw new Error(
-      "Invalid length: point is not in compressed or uncompressed format"
+      "Invalid API: Ensure that you are using a valid public and private key for your API key"
     );
   }
   // Decodes point if its length and first bit match the compressed format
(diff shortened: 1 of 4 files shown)


omkarshanbhag (approved):
Non-blocking suggestion!!

packages/api-key-stamper/src/tink/elliptic_curves.ts
```diff
@@ -172,7 +172,7 @@ export function pointDecode(point: Uint8Array): JsonWebKey {
     point.length !== uncompressedLength
   ) {
     throw new Error(
-      "Invalid length: point is not in compressed or uncompressed format"
+      "Invalid API key: Ensure that you are using a valid public and private key for your API key"
```
omkarshanbhag, line 175: Suggestion: 

This is def a more helpful message than the old one, but maybe we could include some of the old stuff because it points the user in the direction of the fact that the formatting of the API key is what is off. 

Something like: 

`Invalid API key: Ensure that you are using a valid public and private key for your API key, provided in either the compressed or uncompressed point format`

r-n-o, line 175: Sorry to make your life more difficult but I don't think this message should be updated. We're in `src/tink/elliptic_curves.ts` which deals with generic crypto primitives. It's also a vendored file which shouldn't be touched unless we really need to.

My proposal: let's look at the callers of `pointDecode`, catch errors, and re-throw with a different message?

For example, when this fails: https://github.com/tkhq/sdk/blob/d9a046573119022e202fdcbeac70d98b5d66c0a7/packages/sdk-browser/src/utils.ts#L34

This isn't related to API keys. It's because the targetPublicKey is invalid. So you can do something like:
```ts
try {
    const jwk = pointDecode(targetKeyBytes);
} catch (e: Error) {
    if (e.message === "Invalid length: point is not in compressed or uncompressed format") {
        // throw a more helpful error with context
        throw new Error(`target public key is not a valid compressed public key: ${targetPublicKey}`);
    } else {
        // rethrow unmodified
        throw e
    }
}

Comment (zkharit) [reply packages/api-key-stamper/src/tink/elliptic_curves.ts:175]: Hm yeah, that seems like the way to handle it. I didnt realize this was a vendored file, we should def not change it in that case

Comment (zkharit) [reply packages/api-key-stamper/src/tink/elliptic_curves.ts]: I've incorporated some of the original error message in the new message

Comment (zkharit) [reply packages/api-key-stamper/src/tink/elliptic_curves.ts]: I added some try/catches to where decodePoint is being called in the code. However I didn't add the conditional to catch just that error message. All of the error messages that are thrown in decodePoint would be pretty unhelpful imo, so I added general error messages that will lead the user to solving their problem

Comment (zkharit) [reply packages/api-key-stamper/src/tink/elliptic_curves.ts]: Interested in conflicting opinions on this!

File: packages/api-key-stamper/src/utils.ts [hunk 2]
```diff
@@ -11,7 +11,14 @@ export function convertTurnkeyApiKeyToJwk(input: {
 }): JsonWebKey {
   const { uncompressedPrivateKeyHex, compressedPublicKeyHex } = input;
 
-  const jwk = pointDecode(uint8ArrayFromHexString(compressedPublicKeyHex));
+  let jwk;
+  try {
+    jwk = pointDecode(uint8ArrayFromHexString(compressedPublicKeyHex));
+  } catch (e) {
+    throw new Error(
+      `invalid API key: Ensure that you are using a valid public and private key in compressed or uncompressed format and they are not switched`
```
r-n-o, line 19: Do you want to template the value of `compressedPublicKeyHex` in this case?

I think this message should be purely about API public key rather than both public and private keys. Something like `unable to load API key: invalid public key ("${compressedPublicKeyHex}")`?

r-n-o (approved):

PR #467, opened 2025-01-07
add request ids to iframe stamper

andrewkmin:
## Summary & Motivation
$title

iframe counterpart: https://github.com/tkhq/frames/pull/59

This mainly serves the case where there are concurrent, or serial, requests. Request IDs are used to disambiguate between multiple requests.

Latency implications: roughly ~4 seconds for 100 parallel signing requests, compared to ~0.7 seconds if signing 100 payloads using an im-memory API key stamper. The difference comes in overhead cost of the iframe communications.

## How I Tested These Changes
- email auth example (https://github.com/tkhq/sdk/pull/468)
  - tested for backwards compatibility as well

## Did you add a changeset?
- yes

### DIFF
--- a/packages/iframe-stamper/src/index.ts
+++ b/packages/iframe-stamper/src/index.ts
@@ -180,9 +180,6 @@ export class IframeStamper {
   }
 
   onMessageHandler(event: MessageEvent): void {
-    console.log("full event data", event.data);
-    console.log("pending requests", this.pendingRequests);
-
     const { type, value, requestId } = event.data || {};
 
     // Handle messages without requestId (like PUBLIC_KEY_READY)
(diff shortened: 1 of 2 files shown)


packages/iframe-stamper/src/index.ts
```diff
@@ -103,6 +103,16 @@ export type TIframeSettings = {
...
 };
 
+interface PendingRequest {
+  resolve: (value: any) => void;
+  reject: (reason?: any) => void;
+  requestId: string;
+}
+
+function generateUUID(): string {
+  return crypto.randomUUID();
```
andrewkmin, line 113: The iframe stamper will be used strictly in the web browser context, so `crypto` should be an available resource. `randomUUID` is pretty readily available for most modern web browsers.

If we're worried about compatibility, there's an implementation that relies on the older `getRandomValues` (see https://stackoverflow.com/a/2117523)

r-n-o (approved):

PR #469, opened 2025-01-08
EWK Minor Changes 

moe-dev:
## Summary & Motivation
1) Custom session lengths
2) Custom width of auth component
3) Custom generated wallets during create suborg request
4) Some minor css fixes
5) Enter to continue
6) Expose WalletAccount type from SDK Browser/Server as well as all Default accounts

## Did you add a changeset?


packages/sdk-react/src/components/auth/PhoneInput.tsx
```diff
@@ -48,7 +48,6 @@ export const MuiPhone: React.FC<MUIPhoneProps> = ({
       name="phone-input"
       variant="outlined"
       color="primary"
-      autoComplete="off"
```
andrewkmin, line 51: Are we removing because this wasn't working as intended, or because we want to allow for autocomplete?

packages/sdk-react/src/actions/oauth.ts
```diff
@@ -6,6 +6,7 @@ type OauthRequest = {
   suborgID: string;
   oidcToken: string;
   targetPublicKey: string;
+  sessionLength?: number | undefined;
```
andrewkmin, line 9: Nit: we should specify either in name or comment that this is expressed in seconds

Another thing is around type: we're going to cast to string anyways so we could just expect a string all the way through, but at the same time number is probably more intuitive

.changeset/lemon-spoons-own.md
```diff
@@ -0,0 +1,5 @@
+---
+"@turnkey/sdk-react": patch
+---
+
+Add session length customization, more css customization and fixes
```
andrewkmin, line 5: Can you detail the fixes here?

packages/sdk-react/src/actions/oauth.ts
moe-dev: yep agreed! I'm indifferent on number vs string

packages/sdk-react/src/components/auth/PhoneInput.tsx
moe-dev: working as intended but from user stories, and after revisiting this is generally inconsistent with email/phone input fields. If I wrote my email/phone in their respective fields on turnkey before I should be able to see it in the dropdown when I revisit and

packages/sdk-react/src/actions/otpAuth.ts
```diff
@@ -7,7 +7,7 @@ type OtpAuthRequest = {
   otpId: string;
   otpCode: string;
   targetPublicKey: string;
-  sessionLength?: number | undefined;
+  sessionLength?: number | undefined; // Desired expiration time in seconds for the generated API key
```
andrewkmin, line 10: Ah I meant we could add `seconds` to the param name like `sessionLengthSeconds` (similar to `expirationSeconds`). A little verbose, but might be easier for developers when their editor populates params (they might not see the actual code/comment, but should see what params are available).

packages/sdk-react/src/actions/oauth.ts
```diff
@@ -6,6 +6,7 @@ type OauthRequest = {
   suborgID: string;
   oidcToken: string;
   targetPublicKey: string;
+  sessionLengthSeconds?: number | undefined; // Desired expiration time in seconds for the generated API key
```
andrewkmin, line 9: generated API key or auth credential; potato potato

andrewkmin (approved):
🚢

PR #473, opened 2025-01-15
sdk feature release 1/15/2025

andrewkmin:
## Summary & Motivation
Routine (non-mono) release to publish SDK changes.

## Did you add a changeset?


turnekybc (approved):

Ulexus (approved):

packages/sdk-browser/CHANGELOG.md
```diff
@@ -1,5 +1,25 @@
 # @turnkey/sdk-browser
 
+## 1.11.1
```
moe-dev, line 3: Welp a bunch of these patches got squashed into one. Totally fine - but probably no longer a "patch" 😅 Just flagging here so we're aware for next time when we have a bunch of "patches" we want to do without a fast follow release plan

moe-dev (approved):

packages/sdk-browser/CHANGELOG.md
andrewkmin: Yeah I think we have this backlog because of the holidays. Hopefully moving forward we do releases on a per-feature release (i.e. at the time a new feature is merged)

PR #485, opened 2025-01-28
Server Actions

moe-dev:
## Summary & Motivation
Move server actions to sdk-server and consolidate requests that return credential bundles to return a Session object to be used in session abstractions in the near future

## Did you add a changeset?


packages/sdk-server/src/actions.ts
```diff
@@ -0,0 +1,287 @@
+"use server";
+
+import { TurnkeyServerSDK } from "./sdk-client";
+import {
+  DEFAULT_ETHEREUM_ACCOUNTS,
+  DEFAULT_SOLANA_ACCOUNTS,
+  WalletAccount,
+} from "./turnkey-helpers";
+
+type Session = {
```
turnekybc, line 10: I have a couple of thoughts regarding directory layout:
1. There might not be much portability in these types yet but moving them to separate `types.ts` would reduce the size of this file
2. This could be too much from an directory perspective but we could group the actions in their own directory. e.g. `src/passkey/actions.ts`, `src/emailotp/actions.ts`

packages/sdk-server/src/actions.ts
```diff
@@ -0,0 +1,287 @@
...
+      }),
+      ...(request.invalidateExisting && {
+        invalidateExisting: request.invalidateExisting,
+      }),
+    });
+    if (!response.userId) {
+      throw new Error("Expected a non-null userId.");
+    }
+  } catch (error) {
+    console.error(error);
```
turnekybc, line 114: It would be nice to get some telemetry in here at some point

packages/sdk-server/src/actions.ts
moe-dev: I think given that all server actions are related to auth, they can live in one directory for now

turnekybc (approved):

turnekybc (approved):

PR #503, opened 2025-02-13
Patch fix ewk session length & useLocalStorage errors during compling

moe-dev:
## Did you add a changeset?

### DIFF
--- a/.changeset/khaki-glasses-teach.md
+++ b/.changeset/khaki-glasses-teach.md
@@ -1,0 +1,5 @@
+---
+"@turnkey/sdk-react": patch
+---
+
+Patch EWK custom session lengths - previously not working as intended (defaulted to 15 minute sessions only) and fix the following useLocalStorage issue when compiling: [Error: useLocalStorage is a client-only hook]
--- a/packages/sdk-react/package.json
+++ b/packages/sdk-react/package.json
@@ -59,11 +59,11 @@
     "@turnkey/sdk-browser": "workspace:*",
     "@turnkey/sdk-server": "workspace:*",
     "@turnkey/wallet-stamper": "workspace:*",
-    "@uidotdev/usehooks": "^2.4.1",
     "libphonenumber-js": "^1.11.14",
     "next": "^15.0.2",
     "react-apple-login": "^1.1.6",
-    "react-international-phone": "^4.3.0"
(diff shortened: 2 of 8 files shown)


.changeset/khaki-glasses-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+"@turnkey/sdk-react": patch
+---
+
+Patch EWK custom session lengths and fix useLocalStorage issues
```
andrewkmin, line 5: Can we add a little more context here?

i.e. what were the local storage issues we were seeing, and what was the issue with custom session lengths? In other words are we fixing an existing bug/broken state, or adding a net new feature (such as being able to pass through expiration seconds)?

packages/sdk-react/package.json
```diff
@@ -59,11 +59,11 @@
     "@turnkey/sdk-browser": "workspace:*",
     "@turnkey/sdk-server": "workspace:*",
     "@turnkey/wallet-stamper": "workspace:*",
-    "@uidotdev/usehooks": "^2.4.1",
     "libphonenumber-js": "^1.11.14",
     "next": "^15.0.2",
     "react-apple-login": "^1.1.6",
-    "react-international-phone": "^4.3.0"
+    "react-international-phone": "^4.3.0",
+    "usehooks-ts": "^3.1.1"
```
andrewkmin, line 66: Is this dependency change the only necessary fix for the compiler issues?

packages/sdk-react/src/components/auth/Auth.tsx
```diff
@@ -223,12 +229,18 @@ const Auth: React.FC<AuthProps> = ({
 
       const sessionResponse = await passkeyClient?.createReadWriteSession({
         targetPublicKey: authIframeClient?.iframePublicKey!,
+        ...(authConfig.sessionLengthSeconds !== undefined && {
```
andrewkmin, line 232: tiny nit: `authConfig.sessionLengthSeconds &&...` would also be fine given `authConfig.sessionLengthSeconds` would be truthy if it's defined

also could make sense to update the README / have an example of this field being set https://github.com/tkhq/sdk/tree/main/packages/sdk-react#authentication

non-blocking though

moeodeh3 (approved):

PR #504, opened 2025-02-13
sdk-react-native

moeodeh3:
## Summary & Motivation

RN hard

## How I Tested These Changes

I made a local build and tried it in the [react-native-demo-wallet](https://github.com/tkhq/react-native-demo-wallet)

## Did you add a changeset?


packages/sdk-browser/src/sdk-client.ts
```diff
@@ -342,7 +342,6 @@ export class TurnkeyPasskeyClient extends TurnkeyBrowserClient {
   rpId: string;
 
   constructor(config: TurnkeySDKClientConfig) {
-    console.log("config", config);
```
moeodeh3, line 345: I removed this from sdk-client, not sure was this intentional?

moe-dev (commented):
LG at a glance. need to add to the package to code sandbox ci json

packages/react-native-sessions/src/contexts/SessionContext.tsx
```diff
@@ -0,0 +1,206 @@
...
+  // Initializes session state from secure storage on mount
+  useEffect(() => {
+    (async () => {
+      const session = await getSession();
+
+      if (session?.expiry && session.expiry > Date.now()) {
+        setSession(session);
+        config.onSessionCreated?.(session);
+      } else {
+        config.onSessionExpired?.();
```
taylorjdawson, line 46: I like the idea of the `onSessionExpired` callback a lot! but I think as a developer I would expect this to fire once the session expires but this would only fire if the component remounts _after_ the session already expired

It might be out of scope for this PR, but maybe in the future if it makes sense we could do something like I did [here](https://github.com/tkhq/demo-embedded-wallet/pull/18/files#diff-09004c5b49ff536fabc3a8389a27ca33dbba572e098fc840b372a5752e709b0bR391-R424) to fire the callback when the session expires? 

I think at least documenting the behavior of the callback to clarify when the developer can expect this to be fired

taylorjdawson (changes requested):
Looks great! left a few comments

packages/react-native-sessions/src/contexts/SessionContext.tsx
```diff
@@ -0,0 +1,206 @@
...
+import { getPublicKey, decryptCredentialBundle } from "@turnkey/crypto";
+import { generateP256KeyPair } from "@turnkey/crypto";
+import { uint8ArrayToHexString } from "@turnkey/encoding";
+import { OTP_AUTH_DEFAULT_EXPIRATION_SECONDS } from "../constant";
+
+type Session = {
+  publicKey: string;
+  privateKey: string;
+  expiry: number;
+};
```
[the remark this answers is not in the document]
moeodeh3: If we were to do this then I think we should store the entire User object as well. Currently we store the user (and org id) in the turnkey provider here: https://github.com/tkhq/react-native-demo-wallet/blob/main/providers/turnkey.tsx.

Overall I am not strongly opinionated on this, I am fine with whatever decision you make

packages/react-native-sessions/src/contexts/SessionContext.tsx
moeodeh3: very good point, and I like that idea. Going to add it

packages/react-native-sessions/src/contexts/SessionContext.tsx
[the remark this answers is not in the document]
taylorjdawson: if we add it here that would mean we wouldn't need to fetch the orgId [here](https://github.com/tkhq/react-native-demo-wallet/blob/c6dbcc1952e787a88f6d0ae9389fb09b8ccbc49f/providers/turnkey.tsx#L190-L192) I think right?

I would say lets store the User object, since we do this in the sdk-browser? 🤔

taylorjdawson:
@moeodeh3 do we want to rename this to `react-native` or `react-native-sdk`?

packages/react-native-sessions/src/contexts/SessionContext.tsx
```diff
@@ -0,0 +1,390 @@
...
+   * Retrieves the stored embedded key from secure storage.
+   * Optionally deletes the key from storage after retrieval.
+   *
+   * @param deleteKey Whether to remove the embedded key after retrieval. Defaults to `true`.
+   * @returns The embedded private key if found, otherwise `null`.
+   * @throws If retrieving or deleting the key fails.
+   */
+  const getEmbeddedKey = async (deleteKey = true) => {
+    const credentials = await Keychain.getGenericPassword({
+      service: "turnkey-embedded-key",
```
taylorjdawson, line 235: Could we turn this into constant defined in the constants.ts file? and then replace instances of it in this file with the constant?

packages/react-native-sessions/src/contexts/SessionContext.tsx
```diff
@@ -0,0 +1,390 @@
...
+   * @throws If saving the key fails.
+   */
+  const saveEmbeddedKey = async (key: string) => {
+    try {
+      await Keychain.setGenericPassword("turnkey-embedded-key", key, {
+        accessible: Keychain.ACCESSIBLE.WHEN_UNLOCKED_THIS_DEVICE_ONLY,
+        service: "turnkey-embedded-key",
+      });
+    } catch (error) {
+      throw new Error("Could not save the embedded key.");
```
taylorjdawson, line 261: Not a hard requirement but maybe we could define an Error type to give more context about where it is originating? 
for instance we have a [TurnkeyRequestError](https://github.com/tkhq/sdk/blob/20042a3e6c421b4194ecdbbebdde539acfc89916/packages/sdk-browser/src/__types__/base.ts#L32-L49) it might make sense to have like a session error or turnkey react native error? 🤔 
here's another simpler [custom error example](https://github.com/tkhq/sdk/blob/main/packages/wallet-stamper/src/errors.ts) for the wallet stamper

packages/react-native-sessions/src/contexts/SessionContext.tsx
```diff
@@ -0,0 +1,390 @@
...
+
+  /**
+   * Retrieves the stored session from secure storage.
+   *
+   * @returns The stored session or `null` if not found.
+   * @throws If retrieving the session fails.
+   */
+  const getSession = async (): Promise<Session | null> => {
+    const credentials = await Keychain.getGenericPassword({
+      service: "turnkey-session",
```
taylorjdawson, line 289: this could be a constant as well!

taylorjdawson (approved):
Approved with a few suggestions!

PR #512, opened 2025-02-18
Fix phone validation EWK

moe-dev:
## Did you add a changeset?

### DIFF
--- a/packages/sdk-react/src/components/auth/Auth.tsx
+++ b/packages/sdk-react/src/components/auth/Auth.tsx
@@ -17,6 +17,7 @@ import { useTurnkey } from "../../hooks/use-turnkey";
 import { FilterType, OtpType, authErrors } from "./constants";
 import type { WalletAccount } from "@turnkey/sdk-browser";
 import { server } from "@turnkey/sdk-server";
+import parsePhoneNumberFromString from "libphonenumber-js";
 
 const passkeyIcon = (
   <svg xmlns="http://www.w3.org/2000/svg" width="43" height="48" fill="none">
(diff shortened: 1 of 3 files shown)


amircheikh (approved):

packages/sdk-react/src/components/auth/Auth.tsx
```diff
@@ -121,7 +122,10 @@ const Auth: React.FC<AuthProps> = ({
   const isValidEmail = (email: string) =>
     /^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(email);
 
-  const isValidPhone = (phone: string) => /^\+1\d{10}$/.test(phone);
+  const isValidPhone = (phone: string) => {
```
andrewkmin, line 125: these could arguably live in `auth/utils.ts`

packages/sdk-react/src/components/auth/Auth.tsx
```diff
@@ -17,6 +17,7 @@ import { useTurnkey } from "../../hooks/use-turnkey";
 import { FilterType, OtpType, authErrors } from "./constants";
 import type { WalletAccount } from "@turnkey/sdk-browser";
 import { server } from "@turnkey/sdk-server";
+import parsePhoneNumberFromString from "libphonenumber-js";
```
andrewkmin, line 20: do we use `libphonenumber-js` over google's `libphonenumber` because of smaller bundle size? i'd be a bit more comfortable with the latter given it's more robust update/maintenance story

moe-dev: hmm honestly didn't even know of the latter. This seems to be the more popular package

packages/sdk-react/src/components/auth/Auth.tsx
moe-dev: Good call, will follow up on this during phase two of the ewk

PR #516, opened 2025-02-24
getOrCreateSuborg server action

moe-dev:
## Summary & Motivation
Surface a server action that is leveraged to either get or create a suborg (used for continue with... flows)
## Did you add a changeset?

### DIFF
--- a/packages/sdk-server/src/actions.ts
+++ b/packages/sdk-server/src/actions.ts
@@ -7,6 +7,24 @@ import {
   WalletAccount,
 } from "./turnkey-helpers";
 
+type GetOrCreateSuborgRequest = {
+  filterType: FilterType;
+  filterValue: string;
+  additionalData?: {
+    email?: string;
+    phoneNumber?: string;
+    passkey?: Passkey;
+    oauthProviders?: Provider[];
+    customAccounts?: WalletAccount[];
+  };
+};
+
+enum FilterType {
+  Email = "EMAIL",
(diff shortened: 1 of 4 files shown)


packages/sdk-server/src/actions.ts
```diff
@@ -7,6 +7,24 @@ import {
   WalletAccount,
 } from "./turnkey-helpers";
 
+type GetOrCreateSuborgRequest = {
+  filterType: FilterType;
+  filterValue: string;
+  additionalData?: {
```
omkarshanbhag, line 13: do we want to name this field something more descriptive than `additionalData`? My understanding of the info in this field is that if nothing is found by the combination of `FilterType` + `FilterValue` then you use `additionalData` to create a suborg. 

Maybe something like `newSuborgData`?

omkarshanbhag: Aah I'm seeing now that this looks like an internal method and we're mirroring what was used in the older non-server action version. For code readability might still be good to have something like `newSuborgAdditionalData` cause it took me a sec to understand what was going on but I'll defer to you on this.

omkarshanbhag (approved):
Left a quick non-blocking comment but LGTM! 

Question: How was this tested/if we have existing tests that test the broader method lmk where they are!

omkarshanbhag:
Also @moe-dev a brief PR description would be helpful for history/context if possible!

PR #520, opened 2025-02-28
RN multiple sessions [wip]

moeodeh3:
## Summary & Motivation

This PR adds support for multiple sessions by allowing sessions to be created with a custom sessionKey instead of using a single default session. We also introduced the concept of a "selected session," meaning users can switch between sessions by calling `setSelectedSession( {sessionKey: <key here>})` , which updates the client, user, and session info accordingly. Any function calls, like updateUser or createWallet, now apply to the selected session.

## How I Tested These Changes

tested in a custom app using the expo-template :)

## Did you add a changeset?


packages/sdk-react-native/src/constants.ts
```diff
@@ -0,0 +1,9 @@
+// default storage key
+export const TURNKEY_DEFAULT_SESSION_STORAGE = "turnkey-session";
+
+// fixed storage keys (used internally by the SDK)
+export const TURNKEY_EMBEDDED_KEY_STORAGE = "turnkey-embedded-key";
+export const TURNKEY_SESSION_KEYS_INDEX = "turnkey_session_keys_index";
```
andrewkmin, line 6: tiny nit: any particular reason this is using snake case instead of kebab?

packages/sdk-react-native/src/storage.ts
```diff
@@ -0,0 +1,275 @@
+import * as Keychain from "react-native-keychain";
+import {
+  TURNKEY_EMBEDDED_KEY_STORAGE,
```
zkharit, line 3: If I understand the flow correctly, we store a long lived api key at TURNKEY_EMBEDDED_KEY_STORAGE in the Keychain, store session keys in the Keychain in their respective "key" as well as in the key index and then store the selected session in the Keychain as well? It feels like the session index may be redundant and could introduce issues with synchronization? But maybe is there a specific reason we need the index that im missing?

zkharit (commented):
Couple comments, HUGE improvement adding multiple sessions!!

packages/sdk-react-native/package.json
```diff
@@ -50,12 +50,14 @@
...
     "@turnkey/encoding": "workspace:*",
-    "@turnkey/http": "workspace:*",
-    "react": "18.3.1",
-    "react-native": "^0.76.5",
-    "react-native-keychain": "^8.1.0"
+    "@turnkey/http": "workspace:*"
+  },
+  "peerDependencies": {
+    "@types/react": "^18.2.75",
+    "react": "^16.8.0 || ^17.0.0 || ^18.0.0",
```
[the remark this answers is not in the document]
moeodeh3: I just did this to follow what we do in `sdk-react`. I believe because we only use react for hooks, any version above 16.8.0 is sufficient. But we don't want users to think we restrict other major releases so thats why we do this. I could be wrong though - maybe @moe-dev could comment on this. If this is unnecessary I can update both `sdk-react` and `sdk-react-native`

packages/sdk-react-native/src/storage.ts
moeodeh3: the flow looks something like:

1. createEmbeddedKey() - this generates an api key-pair and stores the private key in local storage under TURNKEY_EMBEDDED_KEY_STORAGE, and returns the public key
2. user uses the public key (targetPublicKey) to make a request to turnkey
3. you call createSession() with the credential bundle - this will pull the private key from TURNKEY_EMBEDDED_KEY_STORAGE (also deletes it), decrypt the bundle using the private key, and then saves the session under TURNKEY_DEFAULT_SESSION_STORAGE

so why do we currently store the selected session (TURNKEY_SELECTED_SESSION) and the session index (TURNKEY_SESSION_IDS_INDEX)?

for the useEffect (which runs on app initialization), basically if the user refreshes the app we would like to automatically select the session that was previously selected. Then for the reason we have the TURNKEY_SESSION_IDS_INDEX is because we don't know all the ids we store these sessions under (since the user can set it) so we need to know what service names to pull from in secure storage in the useEffect on app initialization.

I do see what you mean though - let me look into possible ways we can clean this up. Maybe there is a way to get all the service names of stuff in secure storage which would then allow us to remove TURNKEY_SESSION_IDS_INDEX.

packages/sdk-react-native/src/types.ts
```diff
@@ -1,6 +1,7 @@
 import type { TurnkeyApiTypes } from "@turnkey/http";
 
 export type Session = {
+  key: string;
```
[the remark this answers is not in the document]
moeodeh3: good point - changed

packages/sdk-react-native/package.json
```diff
@@ -50,12 +50,14 @@
...
+    "react-native-keychain": "^8.1.0 || ^9.2.2"
   },
   "devDependencies": {
-    "@types/react": "^18.2.6"
+    "react": "^18.2.0"
```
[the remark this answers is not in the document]
moeodeh3: nice catch - fixed

packages/sdk-react-native/README.md
```diff
@@ -119,15 +98,64 @@ const loginWithPasskey = async () => {
 
 ---
 
+## **Functions Provided by the Turnkey Provider**
+
+### **Session Management**
+
+- `createEmbeddedKey()`: Generates a new embedded key pair and securely stores the private key.
+- `createSession({ bundle, expirySeconds?, sessionId? })`: Creates a session. If `sessionId` is provided, it will be stored under that id in secure storage. If no session exists, the first session created is automatically selected.
```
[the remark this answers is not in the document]
andrewkmin: also, to add to the description for `createSession` — what happens when `sessionId` is not provided, and there are multiple existing sessions? is the most recently created session used?

packages/sdk-react-native/README.md
```diff
@@ -119,15 +98,64 @@ const loginWithPasskey = async () => {
...
+
+### **Wallet Management**
+
+- `createWallet({ walletName, accounts, mnemonicLength? })`: Creates a wallet.
+- `importWallet({ walletName, mnemonic, accounts })`: Imports a wallet.
+- `exportWallet({ walletId })`: Exports a wallet mnemonic.
+
+### **Transaction Signing**
+
+- `signRawPayload({ signWith, payload, encoding, hashFunction })`: Signs a payload.
```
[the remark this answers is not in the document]
andrewkmin: optional, but might also be useful to link these endpoints to the appropriate section of our SDK docs https://docs.turnkey.com/api

packages/sdk-react-native/src/constants.ts
```diff
@@ -0,0 +1,9 @@
+// default storage key
+export const TURNKEY_DEFAULT_SESSION_STORAGE = "turnkey-session";
+
+// fixed storage keys (used internally by the SDK)
+export const TURNKEY_EMBEDDED_KEY_STORAGE = "turnkey-embedded-key";
```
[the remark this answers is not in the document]
andrewkmin: however, this goes back to the "id" vs "key" discussion — we should just be consistent with whatever we ultimately go with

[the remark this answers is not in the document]
andrewkmin: do you think it'd be overkill to use enums similar to https://github.com/tkhq/sdk/blob/main/packages/sdk-browser/src/storage.ts ?

packages/sdk-react-native/src/storage.ts
```diff
@@ -0,0 +1,270 @@
...
+};
+
+/**
+ * Retrieves a stored session from secure storage.
+ *
+ * @param sessionId The unique id identifying the session.
+ * @returns The session object if found, otherwise `null`.
+ * @throws If retrieving the session fails.
+ */
+export const getSession = async (sessionId: string): Promise<any | null> => {
```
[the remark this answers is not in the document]
andrewkmin: I'd imagine it would just be `Session`

packages/sdk-react-native/src/turnkey-helpers.ts
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * @param client The authenticated `TurnkeyClient` instance.
+ * @param organizationId The ID of the organization to which the user belongs.
+ * @returns The `User` object containing user details and associated wallets, or `undefined` if the user is not found.
+ * @throws If any API request fails.
+ */
+export const fetchUser = async (
+  client: TurnkeyClient,
+  organizationId: string,
+): Promise<User | undefined> => {
```
andrewkmin, line 44: tiiiny tiny nit: any particular reason you're using `undefined` here as opposed to `null` used elsewhere?

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
```diff
@@ -1,12 +1,12 @@
+import type { ReactNode, FC } from "react";
 import {
-  type ReactNode,
-  type FC,
   createContext,
   useEffect,
   useRef,
   useState,
+  useCallback,
```
andrewkmin, line 7: tiny nit: can we combine this into one single import from "react", or were there type/build errors with that?

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
```diff
@@ -15,12 +15,11 @@ import {
   decryptExportBundle,
 } from "@turnkey/crypto";
 import { uint8ArrayToHexString } from "@turnkey/encoding";
-import { TurnkeyClient } from "@turnkey/http";
+import type { TurnkeyClient } from "..";
```
andrewkmin, line 18: tiny nit: can you use `"../index"` instead? I know the `index` is implied, but just personally prefer it for clarity

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
andrewkmin: also, if that works, then relatedly we can do the same thing (combining imports) here: https://github.com/tkhq/sdk/blob/main/packages/sdk-react-native/src/index.ts#L1-L12

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
```diff
@@ -83,414 +100,350 @@ export const TurnkeyProvider: FC<{
   const [session, setSession] = useState<Session | undefined>(undefined);
   const [client, setClient] = useState<TurnkeyClient | undefined>(undefined);
 
-  const expiryTimeoutRef = useRef<NodeJS.Timeout | null>(null);
+  // map to track expiration timers for each session by sessionId
```
andrewkmin, line 103: seriously nonblocking/tiny nit: we can use `session id` here instead of `sessionId` for consistency (I think `session id` is used more commonly in the codebase/this diff). fine with either, just optimizing for consistency

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
```diff
@@ -83,414 +100,350 @@ export const TurnkeyProvider: FC<{
...
+    async ({ sessionId }: { sessionId: string }) => {
+      const session = await getSession(sessionId);
 
-    const result = await client.updateUser({
-      type: "ACTIVITY_TYPE_UPDATE_USER",
-      timestampMs: Date.now().toString(),
-      organizationId: session.user.organizationId,
-      parameters,
-    });
+      if (session?.expiry && session.expiry > Date.now()) {
```
andrewkmin, line 241: very smol nit: can pull this out into a helper like `isValidSession` or `isLiveSession`

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
```diff
@@ -83,414 +100,350 @@ export const TurnkeyProvider: FC<{
...
-      await Keychain.setGenericPassword(TURNKEY_EMBEDDED_KEY_STORAGE, key, {
-        accessible: Keychain.ACCESSIBLE.WHEN_UNLOCKED_THIS_DEVICE_ONLY,
-        service: TURNKEY_EMBEDDED_KEY_STORAGE,
-      });
-    } catch (error) {
-      throw new TurnkeyReactNativeError(
-        "Could not save the embedded key.",
-        error,
-      );
+  const scheduleSessionExpiration = async (
```
andrewkmin, line 194: the concept of having a bunch of timers all sounds fine. do we want to enforce a limit / maximum on the number of active sessions (and thus timers) that we have? one slight fear (very unlikely) is that somehow we would get into a state where we have a runaway train full of timers getting added to the js event loop, resulting in some slow/buggy behavior. again, very unlikely, but just a thought that popped up from the corner of my mind

packages/sdk-react-native/README.md
```diff
@@ -119,15 +98,64 @@ const loginWithPasskey = async () => {
...
+
+---
+
 ### **Session Storage**
 
-To enable secure authentication, two separate keychain entries are used:
+To enable secure authentication, the following storage ids are used:
```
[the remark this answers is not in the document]
moeodeh3: im indifferent - looking at sdk-browser we use sessionKeys so I changed it back for consistency

packages/sdk-react-native/README.md
```diff
@@ -119,15 +98,64 @@ const loginWithPasskey = async () => {
...
+
+---
+
+## **Handling Multiple Sessions**
+
+Most users won't need multiple sessions, but if your app requires switching between multiple sessions, here’s what you need to know:
+
+This SDK supports **multiple sessions**, allowing you to create and switch between different session ids using `setSelectedSession({ sessionId })`. When a session is selected, the client, user, and session information are updated accordingly, so that all subsequent function calls (like `updateUser` or `createWallet`) apply to the selected session.
+
+- **Creating a Session with a Custom Id**: You can pass a `sessionId` when calling `createSession`. If provided, the session will be stored in secure storage under that id, allowing for multiple sessions.
```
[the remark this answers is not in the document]
moeodeh3: I now throw an error

packages/sdk-react-native/README.md
```diff
@@ -119,15 +98,64 @@ const loginWithPasskey = async () => {
...
-- `turnkey-session`: Stores the session credentials, including the private key, public key, and expiry time, which are decrypted from the credential bundle after a session is created.
+- `turnkey-session`: Default session storage id, storing the session credentials, including the private key, public key, and expiry time, which are decrypted from the credential bundle after a session is created.
+- `turnkey-session-ids-index`: Stores the list of stored session ids.
+- `turnkey-selected-session`: Stores the currently selected session id.
+
+---
+
+## **Handling Multiple Sessions**
+
+Most users won't need multiple sessions, but if your app requires switching between multiple sessions, here’s what you need to know:
```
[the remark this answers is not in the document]
moeodeh3: going to add a brief reason/example

packages/sdk-react-native/src/storage.ts
```diff
@@ -0,0 +1,270 @@
...
+  TURNKEY_SESSION_IDS_INDEX,
+} from "./constants";
+import { TurnkeyReactNativeError } from "./errors";
+import type { Session } from "./types";
+
+/**
+ * Retrieves the stored embedded key from secure storage.
+ * Optionally deletes the key from storage after retrieval.
+ *
+ * @param deleteKey Whether to remove the embedded key after retrieval. Defaults to `true`.
```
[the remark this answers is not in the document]
moeodeh3: changed

packages/sdk-react-native/src/storage.ts
```diff
@@ -0,0 +1,270 @@
...
+  TURNKEY_EMBEDDED_KEY_STORAGE,
+  TURNKEY_SELECTED_SESSION,
+  TURNKEY_SESSION_IDS_INDEX,
+} from "./constants";
+import { TurnkeyReactNativeError } from "./errors";
+import type { Session } from "./types";
+
+/**
+ * Retrieves the stored embedded key from secure storage.
+ * Optionally deletes the key from storage after retrieval.
```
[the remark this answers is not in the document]
moeodeh3: correct

packages/sdk-react-native/src/storage.ts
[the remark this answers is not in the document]
moeodeh3: oh wow don't know how that happened lol - fixed

packages/sdk-react-native/src/turnkey-helpers.ts
moeodeh3: I do this to be consistent throughout the package - I originally had it as null but then I was having to cast stuff from undefined to null so I just changed everything to undefined

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
moeodeh3: I have no idea if this causes build errors lol - I keep getting them and am slowly trying to fix them (I'll change it back for now)

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
moeodeh3: true - I set a max of 10 now

packages/sdk-react-native/README.md
```diff
@@ -56,78 +61,73 @@ export const AppProviders = ({ children }: { children: React.ReactNode }) => {
...
+  - `onSessionSelected`: Called when a session is selected.
+  - `onSessionExpired`: Called when a session expires.
+  - `onSessionCleared`: Called when a session is cleared.
 
-To enable secure authentication, two separate keychain entries are used:
+**When are multiple sessions useful?**
 
-- `turnkey-embedded-key`: Stores the private key that corresponds to the public key used when initiating the session request to Turnkey.
-- `turnkey-session`: Stores the session credentials, including the private key, public key, and expiry time, which are decrypted from the credential bundle after a session is created.
+Using multiple sessions can be beneficial when enabling different authentication methods for various operations. For example, you might authenticate a user with OTP for login while using a passkey-based session for signing transactions.
```
andrewkmin, line 125: 🏅

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
```diff
@@ -83,414 +102,372 @@ export const TurnkeyProvider: FC<{
...
-    privateKey: string,
-    apiBaseUrl: string,
-  ): TurnkeyClient => {
-    const stamper = new ApiKeyStamper({
-      apiPrivateKey: privateKey,
-      apiPublicKey: publicKey,
-    });
-    return new TurnkeyClient({ baseUrl: apiBaseUrl }, stamper);
-  };
+          config.onSessionSelected?.(selectedSession!);
```
andrewkmin, line 154: nice!

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
```diff
@@ -83,414 +102,372 @@ export const TurnkeyProvider: FC<{
...
+      }
+
+      const parameters = {
+        userId: session.user.id,
+        userTagIds: [] as string[],
+        ...(phone?.trim() && { userPhoneNumber: phone }),
+        ...(email?.trim() && { userEmail: email }),
+      };
+
+      const result = await client.updateUser({
```
andrewkmin, line 310: realized we're using the raw http Turnkey client here, instead of something akin to `@turnkey/sdk-server` which abstracts away the need to specify `type/timestamp/org ID`. this is def fine for now, but down the line I wonder if it'd help if we had our React Native SDK also have some similar abstractions

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
```diff
@@ -462,13 +463,47 @@ export const TurnkeyProvider: FC<{
...
+      const sessionKeys = await getSessionKeys();
+
+      for (const key of sessionKeys) {
+        await deleteSession(key);
+        await removeSessionKey(key);
+      }
+
+      await clearSelectedSessionKey();
+
+      setSession(undefined);
```
andrewkmin, line 498: using `undefined` here is helpful — there could also be an argument made for just calling `setSession()`, without any arguments, but I think this works great too

packages/sdk-react-native/src/storage.ts
```diff
@@ -0,0 +1,297 @@
...
+  try {
+    const credentials = await Keychain.getGenericPassword({
+      service: StorageKeys.SessionKeys,
+    });
+
+    let keys: string[] = [];
+
+    if (credentials) {
+      try {
+        keys = JSON.parse(credentials.password);
```
andrewkmin, line 195: probably overkill, but wouldn't be too bad to include a description on how `Keychain.getGenericPassword()` works. i.e. that we're really actually storing the sensitive data (session keys) under `credentials.password` , and that `credentials.password` doesn't actually _have_ to be a literal password

andrewkmin (approved):

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
moeodeh3: noted

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
moeodeh3: setSession() is a state function - I don't think we can do that unless we write a helper function and call it inside. I think I would prefer setSession(undefined) but if you disagree I can change it

packages/sdk-react-native/src/storage.ts
moeodeh3: done

PR #542, opened 2025-03-17
Movement network example

moe-dev:
## Did you add a changeset?


examples/with-movement/README.md
```diff
@@ -0,0 +1,60 @@
...
+$ cp .env.local.example .env.local
+```
+
+Now open `.env.local` and add the missing environment variables:
+
+- `API_PUBLIC_KEY`
+- `API_PRIVATE_KEY`
+- `BASE_URL`
+- `ORGANIZATION_ID`
+- `MOVEMENT_ADDRESS`
```
andrewkmin, line 41: We don't explicitly support `MOVEMENT` addresses. Should they use the `APTOS` address format + derivation path?

examples/with-movement/src/index.ts
```diff
@@ -0,0 +1,146 @@
...
+  });
+
+  const client = new AptosClient(
+    "https://testnet.bardock.movementnetwork.xyz/v1",
+  );
+  const movementAddress = process.env.MOVEMENT_ADDRESS!;
+  const movementPublicKeyHex = process.env.MOVEMENT_PUBLIC_KEY!;
+
+  if (!movementAddress || !movementPublicKeyHex) {
+    throw new Error(
```
andrewkmin, line 26: If it's not too much work, we can add a `createNewWallet` script that creates both of these if not provided

https://github.com/tkhq/sdk/blob/main/examples/with-viem/src/createNewWallet.ts

examples/with-movement/README.md
moe-dev: yep! I'll add a comment here

andrewkmin (approved):

PR #543, opened 2025-03-19
fix: adding missed organizationId param

paco0x:
## Summary & Motivation

In the `@turnkey/viem` package, the `organizationId` parameter is ignored when using a client other than `TurnkeyClient` (e.g., `passkeyClient`). Consequently, the SDK calls the client without the specified `organizationId`, which is unintended.

This change is to align with the Solana SDK code in here: https://github.com/tkhq/sdk/blob/0a97724d265e803067bba74762d4e495855e204c/packages/solana/src/index.ts#L161

## How I Tested These Changes

example in frontend context:

```typescript
// get the passkeyClient
const { passkeyClient } = useTurnkey();
const viemAccount = await createAccount({
  client: passkeyClient,
  organizationId: "YOUR_ORGANIZATION_ID",
  signWith: "YOUR_EVM_ADDRESS",
});

await viemAccount.signMessage({message: "hello"});
```

In the code above, the SDK may send a signing request using an automatically retrieved organization ID, which could differ from the one provided when the signer account was created.

## Did you add a changeset?

### DIFF
--- a/packages/viem/src/index.ts
+++ b/packages/viem/src/index.ts
@@ -403,6 +403,7 @@ async function signTransactionImpl(
     );
   } else {
     const { activity, signedTransaction } = await client.signTransaction({
+      organizationId,
       signWith,
       type: "TRANSACTION_TYPE_ETHEREUM",
       unsignedTransaction: unsignedTransaction,
@@ -480,6 +481,7 @@ async function signMessageImpl(
     result = assertNonNull(activity?.result?.signRawPayloadResult);
   } else {
     const { activity, r, s, v } = await client.signRawPayload({
+      organizationId,
       signWith,
       payload: message,
       encoding: "PAYLOAD_ENCODING_HEXADECIMAL",


moe-dev:
Thanks a tonne for finding this! Moved to https://github.com/tkhq/sdk/pull/544

andrewkmin:
Hi @paco0x , thanks for the contribution. Just to follow up on the need here: 

> In the code above, the SDK may send a signing request using an automatically retrieved organization ID, which could differ from the one provided when the signer account was created.

Is this the case for you? In theory, these two org IDs should be aligned. For example

```
const turnkeyClient = new TurnkeyServerSDK({
    apiBaseUrl: process.env.BASE_URL!,
    apiPrivateKey: process.env.API_PRIVATE_KEY!,
    apiPublicKey: process.env.API_PUBLIC_KEY!,
    defaultOrganizationId: process.env.ORGANIZATION_ID!,
  });
```

```
const turnkeyAccount = await createAccount({
    client: turnkeyClient.apiClient(),
    organizationId: process.env.ORGANIZATION_ID!,
    signWith: process.env.SIGN_WITH!,
  });
```

If the `defaultOrganizationId` passed into `TurnkeyServerSDK` differs from the `organizationId` passed into `createAccount`, then you'll effectively be trying to sign a request using one organization's credentials, that then targets another organization's signing resources

paco0x:
> Hi @paco0x , thanks for the contribution. Just to follow up on the need here:
> 
> > In the code above, the SDK may send a signing request using an automatically retrieved organization ID, which could differ from the one provided when the signer account was created.
> 
> Is this the case for you? In theory, these two org IDs should be aligned. For example
> 
> ```
> const turnkeyClient = new TurnkeyServerSDK({
>     apiBaseUrl: process.env.BASE_URL!,
>     apiPrivateKey: process.env.API_PRIVATE_KEY!,
>     apiPublicKey: process.env.API_PUBLIC_KEY!,
>     defaultOrganizationId: process.env.ORGANIZATION_ID!,
>   });
> ```
> 
> ```
> const turnkeyAccount = await createAccount({
>     client: turnkeyClient.apiClient(),
>     organizationId: process.env.ORGANIZATION_ID!,
>     signWith: process.env.SIGN_WITH!,
>   });
> ```
> 
> If the `defaultOrganizationId` passed into `TurnkeyServerSDK` differs from the `organizationId` passed into `createAccount`, then you'll effectively be trying to sign a request using one organization's credentials, that then targets another organization's signing resources

My use case is like there's a single organization but also registered a bunch of sub-orgs each with a user in the sub-org.

I was testing it in FE, using a user registered in sub-org and signing transactions with the passkey of that user with the turnkey viem SDK. So the code is like:

```
// get the passkeyClient
const { passkeyClient } = useTurnkey();
const viemSigner = await createAccount({
  client: passkeyClient,
  organizationId: ACCOUNT_ORG_ID, // this is the sub-org ID
  signWith: "ACCOUNT_ADDRESS",
});

await viemSigner.signTransaction(...); // this will trigger webauth and I signed it with the user's passkey
```

Then I found that the `organizationId` param in the request sent by the SDK is the parent organization ID, which is incorrect because I signed the transaction with the `viemAccount` created with a specified sub-org ID. Consequently, the response returns an error like "credential not found". Then I found that the `organizationId` passed to `createAccount` in the SDK is unused in this scenario. So the SDK automatically defaults to the parent organization ID, though I haven't investigated the source code to determine the reason for this behavior.

PR #547, opened 2025-03-21
PasskeyConfig Auth component

moe-dev:
## Summary & Motivation
Custom passkey naming for EWK createUserPasskey
## Did you add a changeset?

### DIFF
--- a/.changeset/brown-insects-double.md
+++ b/.changeset/brown-insects-double.md
@@ -1,0 +1,5 @@
+---
+"@turnkey/sdk-react": patch
+---
+
+Add passkeyConfig to EWK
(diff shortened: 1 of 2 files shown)


.changeset/brown-insects-double.md
```diff
@@ -0,0 +1,5 @@
+---
+"@turnkey/sdk-react": patch
+---
+
+Add passkeyConfig to EWK
```
andrewkmin, line 5: non blocking nit: can we add documentation in this PR (or the followup) that describes how to use this config?

essentially digging into

```
export interface PasskeyConfig {
  displayName?: string;
  name?: string;
}
```

a little self explanatory, but I think it's a low-effort way to make things a bit easier for devs! 😄

andrewkmin (approved):

andrewkmin (approved):

PR #566, opened 2025-03-31
Remove diagram which mentions iframes (email auth local storage example)

r-n-o:
## Summary & Motivation
We previously were showing the email auth flow in this example README, but it has references to iframes. That was confusing.

I opted to remove this diagram since we point to our docs and the canonical email auth example, both of which have this diagram already.


turnekybc (approved):

besler613 (approved):
LGTM!

PR #580, opened 2025-04-07
add support for signing type 4 (eip 7702) transactions in `viem`

andrewkmin:
## Summary & Motivation
$title. See changelog

Includes some misc updates
- Foundry version
- Viem version bumps (required)

Also adds multiple scripts for doing flows via Turnkey's SignTransaction _and_ SignRawPayload activities

## How I Tested These Changes
- local

## Did you add a changeset?


packages/viem/src/index.ts
```diff
@@ -500,16 +587,39 @@ async function signMessageImpl(
...
+  sig: TSignature,
+  to: TSignatureFormat = "hex",
+) {
+  const { r: rString, s: sString, v: vString } = sig;
+
+  const r: `0x${string}` = `0x${rString}`;
+  const s: `0x${string}` = `0x${sString}`;
+  const v = BigInt(vString);
+
+  // Turnkey's `v` returned can be used as a proxy for yParity
```
andrewkmin, line 613: this isn't quite scientific, but should suffice for this use case

examples/with-zerodev-aa/src/eip7702.ts
```diff
@@ -0,0 +1,146 @@
...
+      },
+    ]),
+  });
+  console.log("UserOp sent:", userOpHash);
+  console.log("Waiting for UserOp to be completed...");
+
+  const { receipt } = await kernelClient.waitForUserOperationReceipt({
+    hash: userOpHash,
+  });
+  console.log(
```
amircheikh, line 138: Very small and not blocking but, why use `console.log` here and `print` in other places? It also kinda seams a little strange that we have a helper print function anyways but idk 🤷

packages/viem/src/index.ts
```diff
@@ -287,9 +319,59 @@ export async function createApiKeyAccount(
...
+
+  if (to === "object")
+    return {
+      address,
+      chainId,
+      nonce,
+      ...(signature as TSignature),
+    } as any;
+
+  return signature as any;
```
amircheikh, line 372: I thought it was weird to return as `any` here with a defined return type but I guess Viem does this themselves anyways: https://app.unpkg.com/viem@2.23.5/files/accounts/utils/signAuthorization.ts#L55-62

amircheikh (approved):

packages/viem/src/index.ts
andrewkmin: can get this narrowed down the line for sure; thought it was interesting on their end as well

examples/with-zerodev-aa/src/eip7702.ts
andrewkmin: agreed. room for consistency here — will address in a followup

PR #585, opened 2025-04-09
resolve eip 1193 type

andrewkmin:
## Summary & Motivation
$title

## How I Tested These Changes
typecheck

## Did you add a changeset?
yes

### DIFF
--- a/packages/eip-1193-provider/src/utils.ts
+++ b/packages/eip-1193-provider/src/utils.ts
@@ -6,7 +6,7 @@ import {
   serializeTransaction,
 } from "viem";
 
-import type { WalletAddEthereumChain } from "./types";
+import type { WalletAddEthereumChain, TransactionType } from "./types";
 import {
   BlockExplorerUrlError,
   RpcUrlsRequiredError,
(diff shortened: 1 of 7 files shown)


packages/eip-1193-provider/src/utils.ts
```diff
@@ -31,16 +31,20 @@ export const preprocessTransaction = ({
...
+    // TODO: enable these once ready
+    // "0x3": "eip4844",
+    // "0x4": "eip7702",
   };
+
   const processedTransaction: TransactionSerializable = {
     ...transaction,
     // @ts-ignore
     chainId: parseInt(transaction.chainId, 16),
-    type: typeMapping[transaction.type ?? ""] ?? "eip1559",
```
zkharit, line 43: Any reason we defaulted to eip1559 before and now defaulting to legacy?

zkharit (approved):

packages/eip-1193-provider/src/utils.ts
andrewkmin: I was thinking it might be "safer" to default to legacy, but at this point in time it's probably more likely that eip1559 is the most common + likely case. Pushed!

zkharit (approved):

PR #587, opened 2025-04-10
improve ci test stability

andrewkmin:
## Summary & Motivation
$title

Viem tests were very unstable in CI, particularly due to the ERC721 case. Specifically, there were instances where there would be EVM collisions due to the ERC721 contract being deployed, deterministically, to the same address each time. (address and nonce are factors in determining the contract address. You cannot pass a `salt` during contract deployment using viem at this time)

This PR aims to improve test stability by:
- Removing redundant test (that may have been causing evm collisions during contract deployments within viem tests)
- Kill anvil after tests (this is mainly to ensure that processes properly get cleaned up post-test)
- Introducing a separate wallet account address for testing

Interestingly, these race conditions do not happen locally.

If I were to test something else out, there are the following options:
- Have each config use a unique ETH address (currently we reuse some)
- Have unique contracts to deploy per address
- Introduce an effective "jitter"/source of randomness for the number of transactions that an address has performed, thereby controlling its nonce in order to prevent EVM collisions.

## How I Tested These Changes
local, CI

## Did you add a changeset?


.github/workflows/js-build.yml
```diff
@@ -56,6 +56,9 @@ jobs:
           SOLANA_TEST_ORG_API_PRIVATE_KEY: ${{ secrets.SOLANA_TEST_ORG_API_PRIVATE_KEY }}
           WALLET_ID: ${{ secrets.PREPROD_WALLET_ID }}
 
+      - name: Teardown
+        run: lsof -t -i tcp:8545 | xargs kill # kill anvil
```
andrewkmin, line 60: 8545 is the default port for anvil

packages/viem/src/__tests__/index-test.ts
```diff
@@ -92,11 +92,6 @@ describe("TurnkeyAccount", () => {
       signWith: expectedPrivateKeyEthAddress,
       expectedEthAddress: expectedPrivateKeyEthAddress,
     },
-    {
-      configName: "Private Key Address using createAccountWithAddress",
```
andrewkmin, line 96: this is a redundant config. we already test `createAccountWithAddress` using `Wallet Account using createAccountWithAddress`

.changeset/mighty-cycles-press.md
```diff
@@ -0,0 +1,5 @@
+---
+"@turnkey/eip-1193-provider": patch
+---
+
+Tweak unit tests given expectation that the wallet ID used for tests will have 2 associated addresses
```
r-n-o, line 5: Nit: if it's a test-only change and doesn't impact the code that our customers consume whatsoever there's no need for a changelog entry 💅

andrewkmin: sahweet! removed :)

turnekybc (approved):

zkharit (approved):
CI

PR #590, opened 2025-04-10
ensure presence of iframe embedded key

andrewkmin:
## Summary & Motivation
$title

This will ensure that whenever users take auth-related actions, there will be a target embedded key to use

## How I Tested These Changes
- locally, via `examples/react-components`

## Did you add a changeset?
yes

### DIFF
--- a/packages/sdk-react/src/components/auth/Auth.tsx
+++ b/packages/sdk-react/src/components/auth/Auth.tsx
@@ -130,6 +130,12 @@ const Auth: React.FC<AuthProps> = ({
   useEffect(() => {
     if (authIframeClient) {
       setComponentReady(true);
+
+      const updateIframeKey = async () => {
+        await authIframeClient.initEmbeddedKey();
+      };
+
+      updateIframeKey();
     }
   }, [authIframeClient]);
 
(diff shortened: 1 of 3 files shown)


packages/sdk-react/src/components/auth/Auth.tsx
```diff
@@ -199,10 +205,11 @@ const Auth: React.FC<AuthProps> = ({
         }
       }
 
+      const iframePublicKey = await authIframeClient!.initEmbeddedKey();
```
andrewkmin, line 208: Each of these checks could be considered overkill, but at the same time I don't think it would hurt. Unless there's a flow/edge case where the `authIframeClient.iframePublicKey` can be inconsistent across different components?

However, this `useEffect` should help mitigate against that

```
useEffect(() => {
    if (authIframeClient) {
      setComponentReady(true);

      const updateIframeKey = async () => {
        await authIframeClient.initEmbeddedKey();
      };

      updateIframeKey();
    }
  }, [authIframeClient]);
```

moe-dev (approved):

moe-dev (approved):

PR #594, opened 2025-04-11
refreshSession

moeodeh3:
## Summary & Motivation

client asked for abstraction to be able to refresh sessions before they expire (i.e extend the session)

image

went a step up and also added a warning function that runs 15 seconds before the session expiries - to make these an even easier process

## Did you add a changeset?

### DIFF
--- a/packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
+++ b/packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
@@ -497,7 +497,7 @@ export const TurnkeyProvider: FC<{
       const keyToRefresh = sessionKey ?? (await getSelectedSessionKey());
       if (!keyToRefresh) {
         throw new TurnkeyReactNativeError(
-          "Session not found. Either the provided sessionKey is invalid, or no session is currently selected.",
+          "Session not found when refreshing the session. Either the provided sessionKey is invalid, or no session is currently selected.",
         );
       }
 
@@ -530,7 +530,7 @@ export const TurnkeyProvider: FC<{
           ?.credentialBundle;
       if (!bundle) {
         throw new TurnkeyReactNativeError(
-          "Failed to create read/write session.",
+          "Failed to create read/write session when refreshing the session",
         );
       }
 
@@ -539,7 +539,7 @@ export const TurnkeyProvider: FC<{
         StorageKeys.RefreshEmbeddedKey,
       );
       if (!embeddedKey) {
-        throw new TurnkeyReactNativeError("Embedded key not found.");
+        throw new TurnkeyReactNativeError("Embedded key not found when refreshing the session");
       }
 
       const newPrivateKey = decryptCredentialBundle(bundle, embeddedKey);
@@ -553,7 +553,7 @@ export const TurnkeyProvider: FC<{
       );
       const user = await fetchUser(newClient, config.organizationId);
       if (!user) {
-        throw new TurnkeyReactNativeError("User not found.");
+        throw new TurnkeyReactNativeError("User not found when refreshing the session");
       }
 
       const newSession: Session = {
--- a/packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
+++ b/packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
@@ -56,7 +56,7 @@ export interface TurnkeyContextType {
   }) => Promise<Session | undefined>;
   updateUser: (params: { email?: string; phone?: string }) => Promise<Activity>;
   refreshUser: () => Promise<void>;
-  createEmbeddedKey: () => Promise<string>;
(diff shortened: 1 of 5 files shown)


packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
```diff
@@ -203,6 +211,22 @@ export const TurnkeyProvider: FC<{
       clearTimeout(expiryTimeoutsRef.current[sessionKey]);
     }
 
+    // clear existing warning timeouts if it exists
+    if (expiryTimeoutsRef.current[`${sessionKey}-warning`]) {
+      clearTimeout(expiryTimeoutsRef.current[`${sessionKey}-warning`]);
+    }
+
+    const timeUntilExpiry = expiryTime - Date.now();
+    const warningThreshold = SESSION_WARNING_THRESHOLD_SECONDS * 1000;
```
amircheikh, line 220: Would be cool to optionally pass in a threshold here. But not blocking

amircheikh (approved):

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
```diff
@@ -425,6 +468,116 @@ export const TurnkeyProvider: FC<{
...
+      const newPrivateKey = decryptCredentialBundle(bundle, embeddedKey);
+      const newPublicKey = uint8ArrayToHexString(getPublicKey(newPrivateKey));
+      const newExpiry = Date.now() + expirationSeconds * 1000;
+
+      const newClient = createClient(
+        newPublicKey,
+        newPrivateKey,
+        config.apiBaseUrl,
+      );
+      const user = await fetchUser(newClient, config.organizationId);
```
zkharit, line 554: You could do a simple getWhoAmI here, not blocking

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
```diff
@@ -425,6 +468,116 @@ export const TurnkeyProvider: FC<{
...
+      if (!user) {
+        throw new TurnkeyReactNativeError("User not found.");
```
zkharit, line 556: This message should probably be more descriptive to the situation, and include something like "Failed to refresh session"

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
```diff
@@ -425,6 +468,116 @@ export const TurnkeyProvider: FC<{
...
+      const newSession: Session = {
+        key: keyToRefresh,
+        publicKey: newPublicKey,
+        privateKey: newPrivateKey,
+        expiry: newExpiry,
+        user,
+      };
+
+      // if selected session is being refreshed, we update the local state session and client
+      if (session?.key === keyToRefresh) {
```
zkharit, line 568: Will this ever not be true? the sessions `key` is set to `keyToRefresh` right above here

moeodeh3: I believe here we are checking the local state variable (currently selected session) session key - not the new one we just created (the one above). Honestly I should probably write a doc on how this all works - theres a lot of things going on

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
moeodeh3: this is a helper function to basically do that - for RN i decided to have the wallet and wallet accounts in the user object because I thought it made things easier. We also store the user in the session object that way we don't have to keep making requests every time a user leaves the app (the useEffect runs)

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
moeodeh3: will note this down for next release - for now going to skip this as its a bit of work to do since we need to then store it along with the session in secure storage (which would a lot of testing)

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
moeodeh3: good point - especially since this is usually run in the background automatically. Done

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
zkharit: Ah I completely missed the user being used for creating the session, this makes sense!

packages/sdk-react-native/src/contexts/TurnkeyContext.tsx
zkharit: 🤦‍♂️ i mistook `session` for `newSession`

zkharit (approved):

zkharit (approved):

PR #618, opened 2025-05-01
sdk-react-native refreshSession bug fix

moeodeh3:
## Summary & Motivation

removed storing the embeddedKey in secure storage for refreshSession

Whats the reason?
- before we were creating it then getting it from secure storage right after, there is no need to store it
- Injective reported a `TurnkeyReactNativeError: Embedded key not found when refreshing the session` error sometimes. I am suspecting that since the two events are so close together on some slower devices this might be the cause.

slack thread can be found [here](https://turnkeycrypto.slack.com/archives/C08NZQ30FA6/p1746007618850369)

## Did you add a changeset?

yes

### DIFF
--- a/packages/sdk-react-native/src/constants.ts
+++ b/packages/sdk-react-native/src/constants.ts
@@ -1,9 +1,11 @@
 export enum StorageKeys {
   DefaultSession = "@turnkey/session",
   EmbeddedKey = "@turnkey/embedded-key",
-  RefreshEmbeddedKey = "@turnkey/refresh-embedded-key",
   SessionKeys = "@turnkey/session-keys",
   SelectedSession = "@turnkey/selected-session",
+
+  // deprecrated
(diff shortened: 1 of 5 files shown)


packages/sdk-react-native/src/constants.ts
```diff
@@ -1,9 +1,11 @@
 export enum StorageKeys {
   DefaultSession = "@turnkey/session",
   EmbeddedKey = "@turnkey/embedded-key",
-  RefreshEmbeddedKey = "@turnkey/refresh-embedded-key",
   SessionKeys = "@turnkey/session-keys",
   SelectedSession = "@turnkey/selected-session",
+
+  // deprecrated
```
andrewkmin, line 7: let's explicitly note this deprecation in the changelog

andrewkmin (approved):

amircheikh (approved):

PR #633, opened 2025-05-15
feat: add debugging to release workflow

turnekybc:
## Summary & Motivation
 - adding debugging to ensure chaangeset is present to avoid PR creation

## Did you add a changeset?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -142,15 +142,24 @@ jobs:
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
 
-      - name: Version and Publish SDK packages to npm
+      - name: Check for changesets
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -142,15 +142,24 @@ jobs:
...
+            exit 0
+          fi
+
+      - name: Version and Publish
         uses: changesets/action@06245a4e0a36c064a573d4150030f5ec548e4fcc # v1.4.10
         with:
           version: pnpm changeset version # updates package versions and changelogs
-          publish: pnpm publish # publish to npm
+          publish: |
+            echo "//registry.npmjs.org/:_authToken=$NPM_TOKEN" > .npmrc
```
r-n-o, line 157: I don't think we should do this, it'll display the NPM token in plaintext for everyone to see! (this is a public repo, our CI output is public)

PR #659, opened 2025-05-31
export types and models from @turnkey/sdk-browser

turnekybc:
## Summary & Motivation

- export types and models declared in `@turnkey/sdk-browser` for better type support

## How I Tested These Changes

- locally via `examples/react-components`

## Did you add a changeset?

### DIFF
--- a/.changeset/nine-shrimps-listen.md
+++ b/.changeset/nine-shrimps-listen.md
@@ -1,0 +1,7 @@
+---
+"@turnkey/sdk-browser": patch
+"@turnkey/wallet-stamper": patch
+"@turnkey/webauthn-stamper": patch
+---
+
+export types and models from @turnkey/sdk-browser
--- a/packages/sdk-browser/src/index.ts
+++ b/packages/sdk-browser/src/index.ts
@@ -120,11 +120,11 @@ export {
 /** @internal */
 export type { WalletAccount } from "./turnkey-helpers";
 
-export {
-  type TurnkeySDKClientConfig,
-  type TurnkeySDKBrowserConfig,
-  AuthClient,
-} from "./__types__/base";
+// Export all types and values from __types__/base
+export * from "./__types__/base";
+
+// Export all models from models
+export * from "@models";
 
 export { type Session, SessionType } from "@turnkey/sdk-types";
 


.changeset/nine-shrimps-listen.md
```diff
@@ -0,0 +1,7 @@
+---
+"@turnkey/sdk-browser": patch
+"@turnkey/wallet-stamper": patch
+"@turnkey/webauthn-stamper": patch
+---
```
r-n-o, line 5: This looks off no? I was expecting a bump for sdk-browser only

(and this shouldn't be a patch bump: we're adding new functionality with these new exports: so it should be a minor bump)

packages/sdk-browser/src/index.ts
```diff
@@ -120,11 +120,11 @@ export {
...
-export {
-  type TurnkeySDKClientConfig,
-  type TurnkeySDKBrowserConfig,
-  AuthClient,
-} from "./__types__/base";
+// Export all types and values from __types__/base
+export * from "./__types__/base";
+
+// Export all models from models
+export * from "@models";
```
r-n-o, line 127: exporting everything means committing to everything staying stable. Are we...ready for that? 👀

.changeset/nine-shrimps-listen.md
turnekybc: Good call. I will update `@turnkey/sdk-browser` to be a minor bump.

I included `@turnkey/wallet-stamper` and `@turnkey/webauthn-stamper` because the previous release / publish workflow failed to publish `@turnkey/viem`. The wallet-stamper and webauthn-stamper packages get published after viem but this didn't happen because the workflow failed. Out of an abundance of caution I was going to patch bump those two packages to ensure the latest packages are published to NPM.

I can do that in a separate PR if needed.

packages/sdk-browser/src/index.ts
turnekybc: There's only three models that are getting exported from `@models`. The only thing I can see changing is we consolidate their type def into `@turnkey/sdk-types`. We would then simply export those types from `@turnkey/sdk-types` out of `@turnkey/sdk-browser` which would maintain consistency with how consumers of the package are referencing types / models from `@turnkey/sdk-browser`.

.changeset/nine-shrimps-listen.md
r-n-o: Ok ok! Makes sense!

andrewkmin (approved):

packages/sdk-browser/src/index.ts
andrewkmin: @turnekybc in that case, anytime we make any sort of breaking change in `@turnkey/sdk-types`, that would also become a breaking change in `@turnkey/sdk-browser` (if we stick with exporting *)

PR #662, opened 2025-06-02
Add notarizer signature verification (Session verification)

moe-dev:
## Did you add a changeset?


packages/crypto/src/turnkey.ts
```diff
@@ -403,3 +405,62 @@ export const encryptWalletToBundle = async ({
...
+  const [headerB64, payloadB64, signatureB64] = jwt.split(".");
+  if (!signatureB64) throw new Error("invalid JWT: need 3 parts");
+  const signingInput = `${headerB64}.${payloadB64}`;
+
+  /* 2. sha256(sha256(header.payload)) ----------------------------------- */
+  const h1 = sha256(new TextEncoder().encode(signingInput));
+  const msgDigest = sha256(h1); // 32-byte Uint8Array
+
+  /* 3. base64-url decode signature -------------------------------------- */
+  const toB64 = (u: string) =>
```
andrewkmin, line 450: not necessary now but should we move this to `@turnkey/encoding`?

packages/crypto/src/turnkey.ts
```diff
@@ -403,3 +405,62 @@ export const encryptWalletToBundle = async ({
...
+export const verifySessionJwtSignature = async (
+  jwt: string,
+  dangerouslyOverrideNotarizerPublicKey?: string,
+): Promise<boolean> => {
+  const notarizerKeyHex =
+    dangerouslyOverrideNotarizerPublicKey ?? PRODUCTION_NOTARIZER_PUBLIC_KEY;
+
+  /* 1. split JWT -------------------------------------------------------- */
+  const [headerB64, payloadB64, signatureB64] = jwt.split(".");
+  if (!signatureB64) throw new Error("invalid JWT: need 3 parts");
```
andrewkmin, line 442: should we also error if `payloadB64` is somehow unpopulated as well?

packages/crypto/src/turnkey.ts
```diff
@@ -403,3 +405,62 @@ export const encryptWalletToBundle = async ({
...
+ * Verifies that a **session JWT** was signed by Turnkey’s
+ * notarizer key (P-256 / ES256, compact 64-byte r‖s signature).
+ *
+ * How it works
+ * ------------
+ * 1.  Split the JWT into `header.payload.signature`.
+ * 2.  **Double-hash** the string `"header.payload"`:
+ *        `h1 = sha256(header.payload)`
+ *        `msg = sha256(h1)`
+ *     (The Rust signer feeds `h1` into `SigningKey::sign`, which hashes once
```
andrewkmin, line 419: `Rust signer` here == Notarizer correct? (not _the_ Signer enclave)

andrewkmin (approved):

packages/crypto/src/turnkey.ts
andrewkmin: nvm, if `payloadB64` is unpopulated, then `signatureB64` will def be unpopulated as well. so this single check should be sufficient

PR #677, opened 2025-06-09
@turnkey/core and @turnkey/react-wallet-kit 

amircheikh:
A full redesign / rewrite of Turnkey's client side Javascript and React packages.

## [`@turnkey/core`](https://www.npmjs.com/package/@turnkey/core)
A lightweight, framework-agnostic SDK for interacting with Turnkey’s API, replacing `@turnkey/sdk-browser`. It offers a single client with simple, one-line helper functions for authentication, wallet creation, signing, and session management — all working in both web and React Native. Compared to the old SDK, it’s cleaner, more flexible, supports multiple “stampers” for signing, and has better type safety, while removing browser-specific limitations.

## [`@turnkey/react-wallet-kit`](https://www.npmjs.com/package/@turnkey/react-wallet-kit)
A full-featured React wrapper around `@turnkey/core`, replacing `@turnkey/sdk-react`. It adds a provider-based architecture, hooks, and a ready-made modal UI system for login, wallet management, signing, and linking external wallets — all highly customizable in look and behavior. Compared to the old React SDK, it’s more opinionated, comes with built-in UI flows, handles state automatically, and offers much easier integration for common wallet-related tasks.

### DIFF
--- a/packages/core/src/__clients__/core.ts
+++ b/packages/core/src/__clients__/core.ts
@@ -248,7 +248,7 @@ export class TurnkeyClient {
       {
         errorMessage: "Failed to create passkey",
         errorCode: TurnkeyErrorCodes.CREATE_PASSKEY_ERROR,
-        customMessageByMessages: {
+        customErrorByMessages: {
           "timed out or was not allowed": {
             message: "Passkey creation was cancelled by the user.",
             code: TurnkeyErrorCodes.SELECT_PASSKEY_CANCELLED,
@@ -365,7 +365,7 @@ export class TurnkeyClient {
       {
         errorMessage: "Unable to log in with the provided passkey",
         errorCode: TurnkeyErrorCodes.PASSKEY_LOGIN_AUTH_ERROR,
-        customMessageByMessages: {
+        customErrorByMessages: {
           "timed out or was not allowed": {
             message: "Passkey login was cancelled by the user.",
             code: TurnkeyErrorCodes.SELECT_PASSKEY_CANCELLED,
@@ -981,7 +981,7 @@ export class TurnkeyClient {
           {
             errorMessage: "Failed to create stamped request for wallet login",
             errorCode: TurnkeyErrorCodes.WALLET_LOGIN_OR_SIGNUP_ERROR,
-            customMessageByMessages: {
+            customErrorByMessages: {
               "Failed to sign the message": {
                 message: "Wallet auth was cancelled by the user.",
                 code: TurnkeyErrorCodes.CONNECT_WALLET_CANCELLED,
@@ -1171,7 +1171,7 @@ export class TurnkeyClient {
       {
         errorMessage: "Failed to initialize OTP",
         errorCode: TurnkeyErrorCodes.INIT_OTP_ERROR,
-        customMessageByMessages: {
+        customErrorByMessages: {
           "Max number of OTPs have been initiated": {
             message:
               "Maximum number of OTPs has been reached for this contact.",
@@ -1241,7 +1241,7 @@ export class TurnkeyClient {
       {
         errorMessage: "Failed to verify OTP",
         errorCode: TurnkeyErrorCodes.VERIFY_OTP_ERROR,
-        customMessageByMessages: {
+        customErrorByMessages: {
           "Invalid OTP code": {
             message: "The provided OTP code is invalid.",
             code: TurnkeyErrorCodes.INVALID_OTP_CODE,
@@ -1537,7 +1537,7 @@ export class TurnkeyClient {
       oidcToken,
       publicKey,
       createSubOrgParams,
-      providerName = "OpenID Connect Provider" + Date.now(),
+      providerName = "OpenID Connect Provider" + " " + Date.now(),
       sessionKey = SessionKey.DefaultSessionkey,
       invalidateExisting = false,
     } = params;
@@ -1661,7 +1661,7 @@ export class TurnkeyClient {
       {
         errorMessage: "Failed to complete OAuth login",
         errorCode: TurnkeyErrorCodes.OAUTH_LOGIN_ERROR,
-        customMessageByMessages: {
+        customErrorByMessages: {
           "OAUTH disallowed": {
             message:
               "OAuth is disabled on the dashboard for this organization.",
@@ -3630,7 +3630,7 @@ export class TurnkeyClient {
       {
         errorMessage: "Failed to import wallet",
         errorCode: TurnkeyErrorCodes.IMPORT_WALLET_ERROR,
-        customMessageByMessages: {
+        customErrorByMessages: {
           "invalid mnemonic": {
             message: "Invalid mnemonic input",
             code: TurnkeyErrorCodes.BAD_REQUEST,
@@ -3710,7 +3710,7 @@ export class TurnkeyClient {
       {
         errorMessage: "Failed to import wallet",
         errorCode: TurnkeyErrorCodes.IMPORT_WALLET_ERROR,
-        customMessageByMessages: {
+        customErrorByMessages: {
           "invalid mnemonic": {
             message: "Invalid mnemonic input",
             code: TurnkeyErrorCodes.BAD_REQUEST,
--- a/packages/core/src/utils.ts
+++ b/packages/core/src/utils.ts
@@ -917,8 +917,8 @@ export async function getAuthProxyConfig(
  * @param errorOptions.catchFn Optional function to execute in the catch block
  * @param errorOptions.errorMessage The default error message to use if no custom message is found
  * @param errorOptions.errorCode The default error code to use if no custom message is found
- * @param errorOptions.customMessageByCodes Optional mapping of error codes to custom messages, if you're trying to target a specific error code and surface a custom message, use this
- * @param errorOptions.customMessageByMessages Optional mapping of error messages to custom messages, if you're trying to target a specific error message and surface a custom message, use this
+ * @param errorOptions.customErrorByCodes Optional mapping of error codes to custom messages, if you're trying to target a specific error code and surface a custom message, use this
+ * @param errorOptions.customErrorByMessages Optional mapping of error messages to custom messages, if you're trying to target a specific error message and surface a custom message, use this
  * @param finallyFn Optional function to execute in the finally block
  * @returns The result of the async function or throws an error
  */
@@ -928,10 +928,10 @@ export async function withTurnkeyErrorHandling<T>(
     catchFn?: () => Promise<void>;
     errorMessage: string;
     errorCode: TurnkeyErrorCodes;
-    customMessageByCodes?: Partial<
+    customErrorByCodes?: Partial<
       Record<TurnkeyErrorCodes, { message: string; code: TurnkeyErrorCodes }>
     >;
-    customMessageByMessages?: Record<
+    customErrorByMessages?: Record<
       string,
       { message: string; code: TurnkeyErrorCodes }
     >;
@@ -943,8 +943,8 @@ export async function withTurnkeyErrorHandling<T>(
   const {
     errorMessage,
     errorCode,
-    customMessageByCodes,
-    customMessageByMessages,
+    customErrorByCodes,
+    customErrorByMessages,
     catchFn,
   } = catchOptions;
   const finallyFn = finallyOptions?.finallyFn;
@@ -953,7 +953,7 @@ export async function withTurnkeyErrorHandling<T>(
   } catch (error) {
     await catchFn?.();
     if (error instanceof TurnkeyError) {
-      const customCodeMessage = customMessageByCodes?.[error.code!];
+      const customCodeMessage = customErrorByCodes?.[error.code!];
       if (customCodeMessage) {
         throw new TurnkeyError(
           customCodeMessage.message,
@@ -961,20 +961,20 @@ export async function withTurnkeyErrorHandling<T>(
           error,
         );
       }
-      throwMatchingMessage(error.message, customMessageByMessages, error);
+      throwMatchingMessage(error.message, customErrorByMessages, error);
 
       throw error;
     } else if (error instanceof TurnkeyRequestError) {
-      throwMatchingMessage(error.message, customMessageByMessages, error);
+      throwMatchingMessage(error.message, customErrorByMessages, error);
 
       throw new TurnkeyError(errorMessage, errorCode, error);
     } else if (error instanceof Error) {
-      throwMatchingMessage(error.message, customMessageByMessages, error);
+      throwMatchingMessage(error.message, customErrorByMessages, error);
 
       // Wrap other errors in a TurnkeyError
       throw new TurnkeyError(errorMessage, errorCode, error);
     } else {
-      throwMatchingMessage(String(error), customMessageByMessages, error);
+      throwMatchingMessage(String(error), customErrorByMessages, error);
       // Handle non-Error exceptions
       throw new TurnkeyError(String(error), errorCode, error);
     }
@@ -988,25 +988,22 @@ export async function withTurnkeyErrorHandling<T>(
  * If no match is found, it does nothing.
  *
  * @param errorMessage The error message to check against the custom messages.
- * @param customMessageByMessages An object mapping error messages to custom messages and codes.
+ * @param customErrorByMessages An object mapping error messages to custom messages and codes.
  * @param error The original error that triggered this function.
  */
 const throwMatchingMessage = (
   errorMessage: string,
-  customMessageByMessages:
+  customErrorByMessages:
     | Record<string, { message: string; code: TurnkeyErrorCodes }>
     | undefined,
   error: any,
 ) => {
-  if (
-    customMessageByMessages &&
-    Object.keys(customMessageByMessages).length > 0
-  ) {
-    Object.keys(customMessageByMessages).forEach((key) => {
+  if (customErrorByMessages && Object.keys(customErrorByMessages).length > 0) {
+    Object.keys(customErrorByMessages).forEach((key) => {
       if (errorMessage.includes(key)) {
         throw new TurnkeyError(
-          customMessageByMessages[key]!.message,
-          customMessageByMessages[key]!.code,
+          customErrorByMessages[key]!.message,
+          customErrorByMessages[key]!.code,
           error,
         );
       }
--- a/packages/core/src/__tests__/utils-test.ts
+++ b/packages/core/src/__tests__/utils-test.ts
@@ -572,7 +572,7 @@ describe("withTurnkeyErrorHandling", () => {
         {
           errorMessage: DEFAULT_MSG,
           errorCode: DEFAULT_CODE,
-          customMessageByMessages: {
+          customErrorByMessages: {
             ECONNREFUSED: {
               message: "Database unavailable",
               code: TurnkeyErrorCodes.INVALID_REQUEST,
@@ -612,13 +612,13 @@ describe("withTurnkeyErrorHandling", () => {
         withTurnkeyErrorHandling(failWith(original), {
           errorMessage: DEFAULT_MSG,
           errorCode: DEFAULT_CODE,
-          customMessageByCodes: {
+          customErrorByCodes: {
             [DEFAULT_CODE]: {
               message: "override via code",
               code: DEFAULT_CODE,
             },
           },
-          customMessageByMessages: {
+          customErrorByMessages: {
             original: { message: "message-map-hit", code: DEFAULT_CODE },
           },
         }),
@@ -640,8 +640,8 @@ describe("withTurnkeyErrorHandling", () => {
         withTurnkeyErrorHandling(failWith(original), {
           errorMessage: DEFAULT_MSG,
           errorCode: DEFAULT_CODE,
-          // no customMessageByCodes hit
-          customMessageByMessages: {
+          // no customErrorByCodes hit
+          customErrorByMessages: {
             specific: { message: "mapped by message", code: DEFAULT_CODE },
           },
         }),
@@ -681,7 +681,7 @@ describe("withTurnkeyErrorHandling", () => {
         withTurnkeyErrorHandling(failWith(reqErr), {
           errorMessage: DEFAULT_MSG,
           errorCode: DEFAULT_CODE,
-          customMessageByMessages: {
+          customErrorByMessages: {
             expired: {
               message: "Session expired, please re-auth",
               code: DEFAULT_CODE,
@@ -756,7 +756,7 @@ describe("withTurnkeyErrorHandling", () => {
         withTurnkeyErrorHandling(failWith(err), {
           errorMessage: DEFAULT_MSG,
           errorCode: DEFAULT_CODE,
-          customMessageByMessages: {
+          customErrorByMessages: {
             "connection refused": {
               message: "Database unavailable",
               code: DEFAULT_CODE,
@@ -794,7 +794,7 @@ describe("withTurnkeyErrorHandling", () => {
         withTurnkeyErrorHandling(failWith("timeout while fetching"), {
           errorMessage: DEFAULT_MSG,
           errorCode: DEFAULT_CODE,
-          customMessageByMessages: {
+          customErrorByMessages: {
             timeout: { message: "Network timeout", code: DEFAULT_CODE },
           },
         }),


andrewkmin (commented):
quick round of comments, non urgent. will dig into the provider code separately

packages/react-wallet-kit/src/components/export/Export.tsx
```diff
@@ -0,0 +1,149 @@
...
+  } & DefaultParams,
+) {
+  const { exportType, targetPublicKey, stampWith, walletId } = params;
+  const { config } = useTurnkey();
+
+  const [exportIframeVisible, setExportIframeVisible] = useState(false);
+
+  const { closeModal, isMobile } = useModal();
+
+  const apiBaseUrl = config?.apiBaseUrl ?? "http://localhost:8081/";
```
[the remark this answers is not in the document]
ethankonk: No longer need this since we now default to the prod api endpoint url so removing! Great catch!

packages/core/src/__stampers__/api/mobile/stamper.ts
```diff
@@ -0,0 +1,78 @@
...
+    return await Keychain.getAllGenericPasswordServices();
+  }
+
+  async clearKeyPairs(): Promise<void> {
+    const keys = await this.listKeyPairs();
+    for (const key of keys) {
+      await this.deleteKeyPair(key);
+    }
+  }
+  async createKeyPair(externalKeyPair?: {
```
andrewkmin, line 27: tiny nit: let's add a new line above this for consistency

packages/core/src/__stampers__/api/web/stamper.ts
```diff
@@ -0,0 +1,218 @@
...
+  stringToBase64urlString,
+  pointEncode,
+} from "@turnkey/encoding";
+import type { TStamp, ApiKeyStamperBase } from "@types";
+
+const DB_NAME = "TurnkeyStamperDB";
+const DB_STORE = "KeyStore";
+const stampHeaderName = "X-Stamp";
+
+function convertEcdsaIeee1363ToDer(ieee: Uint8Array): Uint8Array {
```
andrewkmin, line 12: feels like these two utility methods belong elsewhere (and also get unit tested)

packages/core/src/__stampers__/api/web/stamper.ts
```diff
@@ -0,0 +1,218 @@
...
+      const extractable = (externalKeyPair.privateKey as any).extractable;
+      if (extractable !== false) {
+        throw new Error("Provided privateKey must be non-extractable.");
+      }
+      privateKey = externalKeyPair.privateKey;
+      publicKey = externalKeyPair.publicKey;
+    } else {
+      const keyPair = await crypto.subtle.generateKey(
+        { name: "ECDSA", namedCurve: "P-256" },
+        false,
```
andrewkmin, line 146: might be useful here (for confidence) to add a comment that this is what's taking care of the non-extractable behavior

packages/core/src/__stampers__/api/web/stamper.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  private async storeKeyPair(
+    publicKeyHex: string,
+    privateKey: CryptoKey,
+  ): Promise<void> {
+    const db = await this.openDb();
+    return new Promise((resolve, reject) => {
+      const tx = db.transaction(DB_STORE, "readwrite");
+      const store = tx.objectStore(DB_STORE);
+      store.put(privateKey, publicKeyHex);
```
andrewkmin, line 89: TIL: indexed db `put` does item, _then_ key!

packages/core/src/__stampers__/passkey/base.ts
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        this.stamper = new PasskeyStamper({
+          ...this.config,
+          rpId: this.config.rpId!,
+          allowCredentials: this.config.allowCredentials?.map((cred) => ({
+            id: uint8ArrayToHexString(cred.id as Uint8Array),
+            type: cred.type,
+            transports: cred.transports,
+          })) as any,
```
andrewkmin, line 53: nonblocking: feels like we have enough information to define this type

packages/core/src/__stampers__/passkey/base.ts
```diff
@@ -0,0 +1,165 @@
...
+            "preferred",
+        },
+      },
+    };
+
+    const attestation = await getWebAuthnAttestation(webauthnConfig);
+
+    return {
+      encodedChallenge: config.publicKey?.challenge
+        ? base64StringToBase64UrlEncodedString(config.publicKey?.challenge)
```
andrewkmin, line 134: feel like if we're in this clause then we would expect `config.publicKey?` to be defined? so 

```suggestion
        ? base64StringToBase64UrlEncodedString(config.publicKey!.challenge)
```

packages/core/src/__stampers__/passkey/base.ts
```diff
@@ -0,0 +1,165 @@
...
+        : encodedChallenge,
+      attestation,
+    };
+  };
+
+  createReactNativePasskey = async (
+    config: Record<any, any> = {},
```
andrewkmin, line 141: nonblocking: I think we should define these optional parameter types sometime soon

packages/core/src/__stampers__/passkey/base.ts
```diff
@@ -0,0 +1,165 @@
...
+      rp: {
+        id: this.config.rpId!,
+        name: this.config.rpName ?? "Turnkey",
+      },
+      user: {
+        id: uuidv4(),
+        name,
+        displayName,
+      },
+      authenticatorName: "End-User Passkey",
```
andrewkmin, line 162: this should be configurable

packages/core/src/__stampers__/api/mobile/stamper.ts
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export class ReactNativeKeychainStamper implements ApiKeyStamperBase {
+  async listKeyPairs(): Promise<string[]> {
+    return await Keychain.getAllGenericPasswordServices();
+  }
+
+  async clearKeyPairs(): Promise<void> {
+    const keys = await this.listKeyPairs();
+    for (const key of keys) {
```
andrewkmin, line 23: tiny style nit: in a similar scenario, `clearAllSessions`, `Promise.all()` is used. any preference on approach?

packages/core/src/__storage__/mobile/storage.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  async clearSession(sessionKey: string): Promise<void> {
+    await this.removeStorageValue(sessionKey);
+
+    const keys = await this.listSessionKeys();
+    const updated = keys.filter((k) => k !== sessionKey);
+    await this.setStorageValue(MobileStorageManager.ALL_SESSION_KEYS, updated);
+
+    const active = await this.getActiveSessionKey();
```
andrewkmin, line 91: could we foresee any use for being able to clear a non-active, but existing, session key? (assuming it's possible to end up in that state. if not, ignore)

packages/core/src/__types__/base.ts
```diff
@@ -0,0 +1,528 @@
...
+  withSecurityKey?: boolean;
+
+  // Option to force platform passkeys on native platforms
+  withPlatformKey?: boolean;
+
+  // Optional extensions. Defaults to empty.
+  extensions?: Record<string, unknown>;
+};
+
+export type TWalletManagerConfig = {
```
andrewkmin, line 372: I see that some types are `T`-prefixed, others aren't. let's be consistent — I personally think we could move away from the prefix, but this isn't a strongly held opinion. wdyt?

packages/core/src/__types__/base.ts
```diff
@@ -0,0 +1,528 @@
...
+    provider: WalletProvider,
+    chainOrId: string | SwitchableChain,
+  ) => Promise<void>;
+}
+
+/**
+ * Solana wallets can directly access the public key without needing a signed message.
+ * @interface SolanaWalletInterface
+ * @extends BaseWalletInterface
+ * @property {function(): string} getPublicKey - Returns the public key, which is the ED25519 hex encoded public key from your Solana wallet public key.
```
andrewkmin, line 491: tiny nit here and in a few other places in this file: it might be more accurate to say `hex-encoded Ed25519 public key`

packages/core/src/__wallet__/base.ts
```diff
@@ -0,0 +1,27 @@
+import { WebWalletManager } from "./web/manager";
+import { isReactNative, isWeb } from "@utils";
+import type { TWalletManagerConfig, WalletManagerBase } from "@types";
+
+export async function createWalletManager(
+  cfg: TWalletManagerConfig,
+): Promise<WalletManagerBase> {
+  if (isReactNative()) {
+    try {
+      // const { MobileWalletManager } = await import("../mobile/base");
```
andrewkmin, line 10: let's add a comment indicating this is intentionally left in, or remove this

packages/core/src/__stampers__/api/base.ts
```diff
@@ -0,0 +1,74 @@
...
+  clearOverridePublicKey(): void {
+    this.publicKeyOverride = undefined;
+  }
+
+  async stamp(payload: string): Promise<TStamp> {
+    let publicKeyHex = this.publicKeyOverride;
+    if (!publicKeyHex) {
+      const session = await this.storageManager.getActiveSession();
+      if (!session) {
+        throw new Error("No active session or token available.");
```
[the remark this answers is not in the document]
amircheikh: https://turnkey-0e7c1f5b-amir-wallet-kit-docs.mintlify.app/sdks/react/troubleshooting#5-no-active-session-or-token-available

I have exactly this lol

packages/core/src/__polyfills__/window.ts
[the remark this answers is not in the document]
amircheikh: I honestly don't know why we do it like this. This is something I ripped from sdk-browser back when this project was first starting out. The person who added it is no longer at Turnkey so I can't even ask them 😢

packages/core/src/__wallet__/web/connector/wallet-connect/base.ts
```diff
@@ -0,0 +1,475 @@
...
+        "At least one namespace must be enabled for WalletConnect",
+      );
+    }
+
+    const session = this.client.getSession();
+    if (hasConnectedAccounts(session)) {
+      return;
+    }
+
+    const namespaces: Record<string, any> = {};
```
andrewkmin, line 74: nonblocking nit: would be nice to replace the `any` with some predefined type possibilities (`eip155, solana`, etc)

packages/core/src/__wallet__/web/connector/wallet-connect/base.ts
```diff
@@ -0,0 +1,475 @@
...
+    if (!hasConnectedAccounts(session))
+      throw new Error("No account found in session");
+  }
+
+  /**
+   * Switch the user’s WalletConnect session to a new EVM chain.
+   *
+   * @param provider   – the WalletProvider returned by getProviders()
+   * @param chainOrId  – either:
+   *                     • a CAIP string ("eip155:1", "eip155:137", …)
```
andrewkmin, line 143: CAIP is just "chain agnostic improvement proposal". I think we meant a specific one — CAIP-2? https://chainagnostic.org/CAIPs/caip-2 . let's define or link CAIP somewhere for the uninitiated.

also, seeing as "•" is a bit nonstandard in my experience, can we replace them with dashes (-) instead?

packages/core/src/__wallet__/web/connector/wallet-connect/base.ts
```diff
@@ -0,0 +1,475 @@
...
+
+    await this.client.pair(namespaces).then((newUri) => {
+      this.uri = newUri;
+    });
+  }
+
+  /**
+   * Builds a lightweight provider interface for the given chain.
+   *
+   * @param chainId - The namespace chain ID (e.g., eip155:1)
```
andrewkmin, line 381: let's reference CAIP here too?

packages/core/src/__stampers__/api/base.ts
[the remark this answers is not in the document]
andrewkmin: amaze. on top of it 🏅

packages/core/src/__stampers__/api/mobile/stamper.ts
andrewkmin: same lol. let's do `Promise.all()` then?

andrewkmin: there's `all()` and `allSettled()`. the first will error out if _any_ of the promises fail, which is our desired behavior I believe

packages/core/src/__clients__/core.ts
```diff
@@ -0,0 +1,3480 @@
...
+   * @returns A promise resolving to a `v1SignRawPayloadResult` containing the signature and metadata.
+   * @throws {TurnkeyError} If signing fails, if the wallet account does not support signing, or if the response is invalid.
+   */
+  signMessage = async (params: {
+    message: string;
+    walletAccount: WalletAccount;
+    encoding?: v1PayloadEncoding;
+    hashFunction?: v1HashFunction;
+    stampWith?: StamperType | undefined;
+    addEthereumPrefix?: boolean;
```
taylorjdawson, line 1841: @moeodeh3 Could you elaborate on the use cases for this param? `addEthereumPrefix`

packages/core/src/__clients__/core.ts
```diff
@@ -0,0 +1,3480 @@
...
+   * - Automatically handles message encoding and hashing based on the wallet account’s address format,
+   *   unless explicitly overridden.
+   *
+   * @param params.message - message to sign.
+   * @param params.walletAccount - wallet account to use for signing.
+   * @param params.encoding - override for the payload encoding (defaults to the encoding appropriate for the address type).
+   * @param params.hashFunction - override for the hash function (defaults to the hash function appropriate for the address type).
+   * @param params.stampWith - stamper to tag the signing request (e.g., Passkey, ApiKey, or Wallet).
+   * @param params.addEthereumPrefix - whether to prefix the message with Ethereum's `"\x19Ethereum Signed Message:\n"` string.
+   *   - If `true` (default for Ethereum), the message is prefixed before signing.
```
taylorjdawson, line 1827: @moeodeh3 This says that we default to true for ethereum but I don't see below in the function where we are setting this value to true in the event that the address format is ethereum

r-n-o (dismissed):
First batch of comments; will continue later! So far it looks really good I have to say 👏

r-n-o:
> First batch of comments; will continue later! So far it looks really good I have to say 👏

Whoops sorry not meant to be an approval yet 🙈

packages/core/scripts/codegen.js
```diff
@@ -0,0 +1,526 @@
...
+
+/**
+ * @param {string} methodName
+ * @returns {string}
+ */
+function methodTypeFromMethodName(methodName) {
+  if (["approveActivity", "rejectActivity"].includes(methodName)) {
+    return "activityDecision";
+  }
+  if (methodName.startsWith("nOOP")) {
```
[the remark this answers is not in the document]
ethankonk: 🤷 this was how `sdk-browser` and `sdk-server` have it in their codegen

packages/core/scripts/codegen.js
```diff
@@ -0,0 +1,526 @@
...
+  imports.push('import type * as SdkTypes from "@turnkey/sdk-types";');
+  imports.push(
+    'import { TurnkeyError, TurnkeyErrorCodes } from "@turnkey/sdk-types";',
+  );
+
+  imports.push('import { parseSession } from "../utils";');
+
+  imports.push('import { StamperType } from "../__types__/base";');
+
+  codeBuffer.push(`
```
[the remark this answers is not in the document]
ethankonk: Agreed, but our other codegens from `sdk-browser` and `sdk-server` use pure js as well, could it have something to do with how we sync API changes from mono?

packages/core/src/__stampers__/api/mobile/stamper.ts
```diff
@@ -0,0 +1,78 @@
...
+    if (!privateKey) {
+      throw new Error(`No private key found for public key: ${publicKeyHex}`);
+    }
+    const stamper = new ApiKeyStamper({
+      apiPublicKey: publicKeyHex,
+      apiPrivateKey: privateKey,
+    });
+    const { stampHeaderName, stampHeaderValue } = await stamper.stamp(payload);
+    return { stampHeaderName, stampHeaderValue };
+  }
```
r-n-o, line 77: Interesting. For each stamp() operation we have to get the key out? I wonder what the security benefits are vs. have the `stamper` persisted as a private instance property on the Mobile stamper?

When there are lots of signature I don't know what the performance or UX for `getGenericPassword` is. Does it trigger faceID or similar?

packages/core/src/__stampers__/api/web/stamper.ts
```diff
@@ -0,0 +1,238 @@
...
+  async createKeyPair(externalKeyPair?: CryptoKeyPair): Promise<string> {
+    let privateKey: CryptoKey;
+    let publicKey: CryptoKey;
+    if (externalKeyPair) {
+      const extractable = (externalKeyPair.privateKey as any).extractable;
+      if (extractable !== false) {
+        throw new Error("Provided privateKey must be non-extractable.");
+      }
+      privateKey = externalKeyPair.privateKey;
+      publicKey = externalKeyPair.publicKey;
```
r-n-o, line 162: We should validate that  the private and public key provided here represent a valid P256 key pair

packages/core/src/__stampers__/api/web/stamper.ts
```diff
@@ -0,0 +1,238 @@
...
+      new Uint8Array(signatureIeee1363),
+    );
+    return uint8ArrayToHexString(signatureDer);
+  }
+
+  async stamp(payload: string, publicKeyHex: string): Promise<TStamp> {
+    const signature = await this.sign(payload, publicKeyHex);
+    const stamp = {
+      publicKey: publicKeyHex,
+      scheme: "SIGNATURE_SCHEME_TK_API_P256",
```
r-n-o, line 230: I feel like this should be brought up as a top-level comment: this indexedDB stamper manages P256 key pairs only.

packages/core/src/__stampers__/api/web/stamper.ts
```diff
@@ -0,0 +1,238 @@
...
+      store.clear();
+      tx.oncomplete = () => {
+        db.close();
+        resolve();
+      };
+      tx.onerror = () => reject(tx.error);
+    });
+  }
+
+  private async sign(payload: string, publicKeyHex: string): Promise<string> {
```
r-n-o, line 209: nit: bare signature interfaces should always take bytes (Uint8Array) instead of string. Otherwise you end up having to make assumptions about string encoding, that's no bueno :D 

(it's not a big deal here because it's a private function, and there's only one caller)

packages/core/src/__storage__/base.ts
```diff
@@ -0,0 +1,22 @@
...
+      throw new Error(
+        `Failed to load storage manager for react-native: ${error}`,
+      );
+    }
+  } else if (isWeb()) {
+    return new WebStorageManager();
+  } else {
+    throw new Error("Unsupported environment for storage manager.");
+  }
+}
```
r-n-o, line 22: It'd be nice to avoid making assumptions when it comes to what users will / will not import. Do you think it's possible to let users import their own storage manager as they need it instead of have a createStorageManager which just wraps the thing that they'll need?

The value we're providing with this thin layer is very low (and can be frustrating for users if we get the `isWeb` or `isReactNative` checks wrong)

r-n-o (commented):
Took another look, this time focusing on stampers / storage!

r-n-o (commented):
Finished reviewing `packages/core/src/__clients__/core.ts`, top to bottom 😤

packages/core/src/__stampers__/api/base.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  clearKeyPairs?(): Promise<void> {
+    return this.stamper.clearKeyPairs();
+  }
+
+  // This allows forcing a specific public key to find the key pair for stamping. The key pair must already exist in indexedDB / Keychain.
+  // This is useful if you need to stamp with a specific key pair without having an active session.
+  // See "signUpWithPasskey" function in core.ts for usage
+  setPublicKeyOverride(publicKeyHex: string | undefined): void {
```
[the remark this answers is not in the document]
ethankonk: Renamed to temporaryPublicKey instead to keep it a little less verbose. Let me know if that sounds good!

packages/core/src/__clients__/core.ts
```diff
@@ -0,0 +1,3795 @@
...
+                displayName,
+              },
+            },
+          });
+          if (!res) {
+            throw new TurnkeyError(
+              "Failed to create React Native passkey",
+              TurnkeyErrorCodes.INTERNAL_ERROR,
+            );
+          }
```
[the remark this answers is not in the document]
ethankonk: withTurnkeyErrorHandling function rethrows errors that are already wrapped in a TurnkeyError (with a code attached), so errors from createWebPasskey should retain their og cause

packages/core/src/utils.ts
```diff
@@ -0,0 +1,908 @@
...
+  catchOptions: {
+    catchFn?: () => Promise<void>;
+    errorMessage: string;
+    errorCode: TurnkeyErrorCodes;
+    customMessageByCodes?: Partial<
+      Record<TurnkeyErrorCodes, { message: string; code: TurnkeyErrorCodes }>
+    >;
+    customMessageByMessages?: Record<
+      string,
+      { message: string; code: TurnkeyErrorCodes }
```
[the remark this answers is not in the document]
ethankonk: This essential allows you to pass specific messages you want to catch and wrap it in a new (more readable) error + code. Example of this is if you're trying to catch a user canceling their passkey creation, you can catch the specific message.

packages/core/src/utils.ts
```diff
@@ -0,0 +1,908 @@
...
+    } else if (error instanceof Error) {
+      throwMatchingMessage(error.message, customMessageByMessages, error);
+
+      // Wrap other errors in a TurnkeyError
+      throw new TurnkeyError(errorMessage, errorCode, error);
+    } else {
+      throwMatchingMessage(String(error), customMessageByMessages, error);
+      // Handle non-Error exceptions
+      throw new TurnkeyError(String(error), errorCode, error);
+    }
```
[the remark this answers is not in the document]
ethankonk: Will add the test cases! But essentially what I was goinging for is if the error is an instance of some other Error type (which should be fairly rare ex: if some library we use wraps their errors in some other way) their message is thrown instead to try to preserve whatever message they were going for (this way we can catch specific messages in `react-wallet-kit`). But this is also kind of a convoluted case and could have some strange drawbacks, but in the end, the original error should still be preserved in the error cause (the third parameter you pointed out)

packages/core/src/__clients__/core.ts
```diff
@@ -0,0 +1,3795 @@
...
+   * @param params.createSubOrgParams - parameters for creating a sub-organization (e.g., authenticators, user metadata).
+   * @param params.sessionKey - session key to use for storing the session (defaults to the default session key).
+   * @param params.passkeyDisplayName - display name for the passkey (defaults to a generated name based on the current timestamp).
+   * @param params.expirationSeconds - session expiration time in seconds (defaults to the configured default).
+   * @returns A promise that resolves to a signed JWT session token for the new sub-organization.
+   * @throws {TurnkeyError} If there is an error during passkey creation, sub-organization creation, or session storage.
+   */
+  signUpWithPasskey = async (params?: {
+    createSubOrgParams?: CreateSubOrgParams;
+    sessionKey?: string;
```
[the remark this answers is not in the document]
ethankonk: Not too sure what you mean, is it passing custom CreateSubOrgParams? If so, we need it to allow for wallet creation on sub-org create time and stuff like that. If its session keys that would mainly be for multi session stuff

packages/core/src/__clients__/core.ts
```diff
@@ -0,0 +1,3795 @@
...
+          const signUpBody = buildSignUpBody({
+            createSubOrgParams: {
+              ...createSubOrgParams,
+              apiKeys: [
+                {
+                  apiKeyName: `wallet-auth:${publicKey}`,
+                  publicKey: publicKey,
+                  curveType: isEthereumProvider(walletProvider)
+                    ? ("API_KEY_CURVE_SECP256K1" as const)
+                    : ("API_KEY_CURVE_ED25519" as const),
```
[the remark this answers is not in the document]
moeodeh3: bit🌽

packages/core/src/__clients__/core.ts
```diff
@@ -0,0 +1,3795 @@
...
+   * Signs a message using the specified wallet account.
+   *
+   * - Supports both embedded and connected wallets.
+   * - For **connected wallets**:
+   *   - Delegates signing to the wallet provider’s native signing method.
+   *   - **Important:** For Ethereum wallets (e.g., MetaMask), signatures follow [EIP-191](https://eips.ethereum.org/EIPS/eip-191).
+   *     The message is automatically prefixed with `"\x19Ethereum Signed Message:\n" + message length`
+   *     before signing. As a result, this signature **cannot be used as a raw transaction signature**
+   *     or broadcast on-chain.
+   * - For **embedded wallets**, uses the Turnkey API to sign the message directly.
```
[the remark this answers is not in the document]
moeodeh3: hmm yeah, this is a conversation I had with @andrewkmin. It was about how signing works so differently between connected and embedded wallets. My end goal was to unify the code paths so there’s no divergence between connected and embedded wallets. I don’t like the idea of developers having to implement separate code paths for embedded wallets - in my ideal experience, they should look the same to the developer. Of course, there are small differences we can’t always hide, like this prefix thing. My solution there was to provide extremely clear error codes so developers know exactly what went wrong without needing to spam support

the risk though is that someone only tests with embedded wallets and not with connected ones, and then misses something like this (e.g., the prefix not being enabled), which then makes it to production. At that point, I’d put the responsibility on the developer for not testing properly. That said, I see how this could be confusing, and I’m open to separating them. But I really do envision one unified experience for both connected and embedded wallets

moe-dev (dismissed):
The day has come (again)

andrewkmin (approved):

andrewkmin (approved):
let's ship it.

moe-dev (approved):

PR #699, opened 2025-06-23
update der signature parsing

andrewkmin:
## Summary & Motivation
$title, per audit.

## How I Tested These Changes
- unit

## Did you add a changeset?

### DIFF
--- a/packages/crypto/src/__tests__/crypto-test.ts
+++ b/packages/crypto/src/__tests__/crypto-test.ts
@@ -16,6 +16,7 @@ import {
   formatHpkeBuf,
   verifyStampSignature,
   verifySessionJwtSignature,
+  fromDerSignature,
 } from "../";
 
 // Mock data for testing
@@ -31,7 +32,7 @@ describe("HPKE Encryption and Decryption", () => {
     const senderKeyPair = generateP256KeyPair();
     const receiverKeyPair = generateP256KeyPair();
     const receiverPublicKeyUncompressed = uncompressRawPublicKey(
-      uint8ArrayFromHexString(receiverKeyPair.publicKey),
+      uint8ArrayFromHexString(receiverKeyPair.publicKey)
     );
 
     const textEncoder = new TextEncoder();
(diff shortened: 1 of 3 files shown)


packages/crypto/src/__tests__/crypto-test.ts
```diff
@@ -257,6 +258,201 @@ describe("Turnkey Crypto Primitives", () => {
...
+        0x02, // Correct tag for r
+        0x21, // length of INTEGER // 33 -- this is valid
+        ...new Array(33).fill(0x01), // r -- this is invalid, as the first byte in a 33 byte sequence is a non-padding byte
+        0x02, // Correct tag for s
+        0x20, // length of INTEGER
+        ...new Array(32).fill(0x02), // s
+      ]);
+
+      expect(() => fromDerSignature(invalidRTagHex)).toThrow(
+        /invalid number of starting zeroes/,
```
andrewkmin, line 399: this error log is technically thrown from the downstream `encoding` library (`normalizePadding`) — if the shape of that error message is changed (fairly unlikely), this will have to change as well.

packages/crypto/src/crypto.ts
```diff
@@ -536,31 +536,90 @@ const bigIntToHex = (num: bigint, length: number): string => {
...
+  let index = 1;
+  const lengthByte = derSignatureBuf[index]!;
+  if (lengthByte <= 0x7f) {
+    // Short form: single byte length
+
+    // directly take the consumed value as length and check against buffer
+    // buffer length: initial header bytes + claimed remaining length
+    if (derSignatureBuf.length < 1 + 1 + lengthByte) {
+      throw new Error(
+        "failed to convert DER-encoded signature: insufficient length",
```
r-n-o, line 563: I think this message should be updated to be different than the first message? Maybe ("expected X, got Y")...just so there's a difference between passing a < 2 byte value and messing up the length byte encoding.

r-n-o (approved):
2 small things flagged above, but don't let that stop you. The version you have here is much better than before; ty for the extra test coverage! ❤️

r-n-o (approved):

PR #716, opened 2025-06-30
bs58check

moeodeh3:
## Summary & Motivation

- upgrades bs58check to 4.0.0 in our entire codebase because of a security vulnerability reported
- fixes decryptCredentialBundle not working in react-native

why did this happen? Basically the latest version of bs58check introduced ESM/CJS export conditions using the exports field in package.json, which React Native Metro doesn’t resolve correctly.

## How I Tested These Changes

- Called `decryptCredentialBundle` in the `react-native-demo-wallet` referencing the CI build and it worked 
- For web relying on unit tests

## Did you add a changeset?

### DIFF
--- a/packages/crypto/src/bs58check.ts
+++ b/packages/crypto/src/bs58check.ts
@@ -1,0 +1,34 @@
+// This is a temporary shim for bs58check@4.0.0
+//
+// See: https://github.com/bitcoinjs/bs58check/issues/47
+//
+// bs58check v4.0.0 uses ESM with only a default export, which causes compatibility
+// issues with Metro (React Native). When importing the package using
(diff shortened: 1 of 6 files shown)


packages/crypto/src/bs58check.ts
```diff
@@ -0,0 +1,34 @@
...
+
+type Bs58Check = {
+  encode(buf: Uint8Array): string;
+  decode(str: string): Uint8Array;
+  decodeUnsafe(str: string): Uint8Array | undefined;
+};
+
+function unwrap(obj: any): any {
+  let cur = obj;
+  while (cur && !cur.decode && cur.default) {
```
andrewkmin, line 26: any reason we shouldn't include checks for `encode` and `decodeUnsafe` as well? verbosity?

PR #720, opened 2025-07-01
fix decryptExportBundle not working in some environments

moeodeh3:
## Summary & Motivation

- upgrades bs58 to 6.0.0 in our entire codebase because of a security vulnerability reported
- fixes decryptExportBundle not working in some environments

why did this happen? Basically the latest version of bs58 introduced ESM/CJS export conditions using the exports field in package.json, which some environments don't resolve correctly.

## Did you add a changeset?

### DIFF
--- a/.changeset/fluffy-taxis-flow.md
+++ b/.changeset/fluffy-taxis-flow.md
@@ -1,0 +1,5 @@
+---
+"@turnkey/crypto": patch
+---
+
+Fixed `decryptExportBundle` not working in some environments by adding a shim to handle `bs58`'s ESM-only export.


packages/solana/package.json
```diff
@@ -56,7 +56,7 @@
   },
   "devDependencies": {
     "@turnkey/api-key-stamper": "workspace:*",
-    "bs58": "^5.0.0",
+    "bs58": "^6.0.0",
```
moeodeh3, line 59: since this is a dev dependency I don't think we need a changeset

andrewkmin (approved):
lgtm.

realizing the bs58 codepath isn't currently covered in our unit tests https://github.com/tkhq/sdk/blob/1e3dbf956ddac507252bb0cc8b53b73ac84e48e2/packages/crypto/src/__tests__/crypto-test.ts#L141-L147

that wouldn't necessarily have caught this earlier for us, but as we've talked about, more coverage there would be 💯

PR #724, opened 2025-07-07
Add a sign and broadcast Stacks tx example

t-vila:
Example showing how to sign a Stacks transaction with Turnkey

### DIFF
--- a/examples/with-stacks/.env.local.example
+++ b/examples/with-stacks/.env.local.example
@@ -1,0 +1,4 @@
+TURNKEY_API_PUBLIC_KEY="<Turnkey API Public Key (that starts with 02 or 03)>"
+TURNKEY_API_PRIVATE_KEY="<Turnkey API Private Key>"
+TURNKEY_BASE_URL="https://api.turnkey.com"
+TURNKEY_ORGANIZATION_ID="<Turnkey organization ID>"
--- a/examples/with-stacks/package.json
+++ b/examples/with-stacks/package.json
@@ -1,0 +1,19 @@
+{
+  "name": "@turnkey/with-stacks",
+  "version": "0.0.1",
+  "private": true,
+  "scripts": {
+    "build": "pnpm -w run build-all",
+    "start": "tsx src/index.ts",
+    "validate": "tsx src/validate.ts",
+    "clean": "rimraf ./dist ./.cache",
+    "typecheck": "tsc --noEmit"
+  },
+  "dependencies": {
+    "@stacks/common": "^7.0.2",
+    "@stacks/transactions": "^7.1.0",
+    "@turnkey/sdk-server": "workspace:*",
+    "dotenv": "^16.0.3",
+    "path": "^0.12.7"
+  }
+}
--- a/examples/with-stacks/src/index.ts
+++ b/examples/with-stacks/src/index.ts
@@ -1,0 +1,156 @@
+import { Turnkey as TurnkeyServerSDK } from "@turnkey/sdk-server";
+import {
+  broadcastTransaction,
+  createMessageSignature,
+  createStacksPublicKey,
+  leftPadHex,
+  makeUnsignedContractCall,
+  PubKeyEncoding,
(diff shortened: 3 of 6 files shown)


examples/with-stacks/src/index.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  let nextSigHash = txidFromData(sigHashBytes);
+
+  return nextSigHash;
+};
+
+const signStacksTx = async () => {
+  try {
+    const stacksPublicKey =
```
r-n-o, line 93: same here, I think this public key should come from an env var if we think people running this example will need to override it

examples/with-stacks/src/index.ts
```diff
@@ -0,0 +1,156 @@
...
+    const signature = await client?.apiClient().signRawPayload({
+      payload,
+      signWith: stacksPublicKey,
+      encoding: "PAYLOAD_ENCODING_HEXADECIMAL",
+      hashFunction: "HASH_FUNCTION_NO_OP",
+    });
+
+    // r and s values returned are in hex format
+    // may need to padStart r and s values
+    // v should be "00" for Stacks but the returned "01" also works
```
r-n-o, line 115: this is not always 01 or 00. v is important for ecrecover. I think you're doing the right thing here by inserting `signature.v` rather than hardcoding "00". So let's remove this line of comment?

examples/with-stacks/src/index.ts
```diff
@@ -0,0 +1,156 @@
...
+    )}${signature!.s.padStart(64, "0")}`;
+
+    let nextSigHash = generatePostSignSigHash(
+      stacksPublicKey!,
+      preSignSigHash,
+      nextSig,
+    );
+
+    // Reassign current sigHash with `nextSigHash`
+    stacksTxSigner.sigHash = nextSigHash;
```
r-n-o, line 128: This feels confusing. Essentially it's the "combine" step where you take the raw signature and attach it to the unsigned transaction. I'm surprised there's no SDK function for this already

r-n-o (commented):
Wow nice job figuring this out!

examples/with-stacks/src/index.ts
```diff
@@ -110,9 +111,7 @@ const signStacksTx = async () => {
       hashFunction: "HASH_FUNCTION_NO_OP",
     });
 
-    // r and s values returned are in hex format
-    // may need to padStart r and s values
-    // v should be "00" for Stacks but the returned "01" also works
+    // r and s values returned are in hex format, padStart r and s values
     const nextSig = `${signature!.v}${signature!.r.padStart(
```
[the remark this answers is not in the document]
r-n-o: Oh whoops. My bad! Consider this resolved. I see it included now.

examples/with-stacks/.env.local.example
```diff
@@ -0,0 +1,6 @@
+TURNKEY_API_PUBLIC_KEY="<Turnkey API Public Key (that starts with 02 or 03)>"
+TURNKEY_API_PRIVATE_KEY="<Turnkey API Private Key>"
+TURNKEY_BASE_URL="https://api.turnkey.com"
+TURNKEY_ORGANIZATION_ID="<Turnkey organization ID>"
+TURNKEY_SIGNER_PUBLIC_KEY="<Turnkey signing public key"
```
r-n-o, line 5: nit: missing closing `>` on this line

examples/with-stacks/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@turnkey/with-stacks",
+  "version": "0.0.1",
+  "private": true,
+  "scripts": {
+    "build": "pnpm -w run build-all",
+    "start": "tsx src/index.ts",
+    "validate": "tsx src/validate.ts",
```
r-n-o, line 8: remove this, there's no `validate.ts` file here :)

r-n-o (approved):

PR #847, opened 2025-08-20
update viem implementation to use PAYLOAD_ENCODING_EIP7702_AUTHORIZATION

andrewkmin:
## Summary & Motivation
$title

## How I Tested These Changes
locally

## Did you add a changeset?
yes

### DIFF
--- a/packages/viem/src/index.ts
+++ b/packages/viem/src/index.ts
@@ -22,7 +22,6 @@ import type {
   TransactionSerializable,
   TypedData,
 } from "viem";
-import { hashAuthorization } from "viem/utils";
 import { secp256k1 } from "@noble/curves/secp256k1";
 
 import {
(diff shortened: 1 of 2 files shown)


packages/viem/src/index.ts
```diff
@@ -360,18 +359,16 @@ export async function signAuthorization(
...
-  const hashedAuthorization = hashAuthorization({
-    address,
-    chainId,
-    nonce,
-  });
-
   const signature = await signMessageWithErrorWrapping(
     client,
-    hashedAuthorization,
+    JSON.stringify({
```
moeodeh3, line 364: so `PAYLOAD_ENCODING_EIP7702_AUTHORIZATION` means that Turnkey does the hashing internally?

moeodeh3, line 364: this doesn't apply here but just knowledge sharing, recently learned that `JSON.stringify` throws for BigInts in node <20

PR #874, opened 2025-08-26
Adding a with-aave example

t-vila:
## Did you add a changeset?

### DIFF
--- a/examples/with-aave/README.md
+++ b/examples/with-aave/README.md
@@ -1,0 +1,85 @@
+# Example: `with-aave`
+
+This example shows how to use a Turnkey wallet to interact with Aave v3 on Base mainnet, supplying and withdrawing USDC. It provides the following scripts:
+
+- `deposit.ts` approves the Aave Pool contract to spend USDC and deposits USDC into Aave
+- `balance.ts` reads the aUSDC balance (eg. the deposit receipt)
+- `withdraw.ts` is withdrawing an USDC amount from the pool
+
+On top of it, we showcase the power of the Turnkey policy engine by allowing a non-root Turnkey user to only sign the specific transactions to the USDC and Aave Pool contract addresses:
+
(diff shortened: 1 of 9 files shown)


examples/with-aave/README.md
```diff
@@ -0,0 +1,85 @@
+# Example: `with-aave`
+
+This example shows how to use a Turnkey wallet to interact with Aave v3 on Base mainnet, supplying and withdrawing USDC. It provides the following scripts:
+
+- `deposit.ts` approves the Aave Pool contract to spend USDC and deposits USDC into Aave
+- `balance.ts` reads the aUSDC balance (eg. the deposit receipt)
+- `withdraw.ts` is withdrawing an USDC amount from the pool
```
andrewkmin, line 7: tiny nit: same grammatical case would be `is withdrawing --> withdraws`

examples/with-aave/README.md
```diff
@@ -0,0 +1,85 @@
...
+```
+
+### 2/ Setting up Turnkey
+
+The first step is to set up your Turnkey organization and account. By following the [Quickstart](https://docs.turnkey.com/getting-started/quickstart) guide, you should have:
+
+- A root user with a public/private API key pair within the Turnkey parent organization
+- An organization ID
+
+The next step is to create another user within the organization with a different API key and remove it from the root quorum. You can do this from the Turnkey [dashboard](https://app.turnkey.com/dashboard/security/updateRootQuorum) or [API](https://docs.turnkey.com/api-reference/activities/update-root-quorum). Here's a simple [script](https://github.com/tkhq/sdk/blob/main/examples/kitchen-sink/src/sdk-server/updateRootQuorum.ts) that shows how to update the root quorum using `@turnkey/sdk-server`.
```
andrewkmin, line 41: would an easier way be creating a non-root API-only user via dashboard? 
image

andrewkmin (dismissed):
minor comments; haven't been able to test locally but lgtm overall

examples/with-aave/README.md
t-vila: yeah that's pretty much the assumption as one would have already done through the quickstart to create the root user. I just laid out all the available methods to remove it from the quorum.

PR #946, opened 2025-09-15
WalletConnect bug fixes

moeodeh3:
## How I Tested These Changes
- fix when authenticating, double sign prompt for already connected WalletConnect wallets in dev mode

## Did you add a changeset?

### DIFF
--- a/packages/react-wallet-kit/src/providers/client/Types.tsx
+++ b/packages/react-wallet-kit/src/providers/client/Types.tsx
@@ -4,6 +4,7 @@ import type {
   TurnkeySDKClientBase,
   TurnkeyClientMethods,
   Wallet,
+  WalletProvider,
 } from "@turnkey/core";
 import type {
   Session,


packages/react-wallet-kit/src/components/auth/Wallet.tsx
```diff
@@ -438,28 +466,35 @@ export function WalletConnectScreen(props: WalletConnectScreenProps) {
...
     try {
-      await onDisconnect?.(walletConnectProvider ?? provider);
-      const newProviders = await fetchWalletProviders();
-      setWalletConnectProvider(
-        newProviders.find((p) => p.interfaceType === provider.interfaceType),
-      );
+      // Use the current provider from ref for disconnect
+      await onDisconnect?.(latestProviderRef.current!);
     } catch (err) {
+      console.error("Error disconnecting wallet:", err);
```
amircheikh, line 487: Do we wanna keep the console.error? I dont think we do this in other places

packages/react-wallet-kit/src/providers/modal/Provider.tsx
```diff
@@ -26,13 +26,21 @@ export const ModalContext = createContext<ModalContextType | undefined>(
...
   const [modalStack, setModalStack] = useState<ModalPage[]>([]);
-
   const { isMobile, width } = useScreenSize();
-  const openModal = (page: ModalPage) => setModalStack([page]);
-  const pushPage = (page: ModalPage) =>
-    setModalStack((prev) => [...prev, page]);
-  const popPage = () => setModalStack((prev) => prev.slice(0, -1));
-  const closeModal = () => setModalStack([]);
+
+  const openModal = useCallback((page: ModalPage) => setModalStack([page]), []);
```
amircheikh, line 31: Make sure this actually doesnt break anything

moeodeh3: removed it for safe measure

PR #974, opened 2025-09-24
Proof verification

narimonf:
## Summary & Motivation

Logic for verifying an app proof in conjunction with its boot proof.

## How I Tested These Changes

Unit tests with fixtures.

## Did you add a changeset?

Yes


packages/crypto/src/proof.ts
```diff
@@ -0,0 +1,263 @@
...
+
+  // Verify ECDSA signature
+  const isValid = p256.verify(signatureBytes, payloadDigest, signingKeyBytes);
+  if (!isValid) {
+    throw new Error("Signature verification failed");
+  }
+}
+
+// Pinned AWS Nitro Enclaves Root
+const AWS_ROOT_CERT_PEM = `-----BEGIN CERTIFICATE-----
```
amircheikh, line 136: Prolly should move this to constants.ts

packages/crypto/src/proof.ts
```diff
@@ -0,0 +1,263 @@
...
+
+function bytesEq(a: ArrayBuffer, b: ArrayBuffer) {
+  const A = new Uint8Array(a),
+    B = new Uint8Array(b);
+  if (A.length !== B.length) return false;
+  for (let i = 0; i < A.length; i++) if (A[i] !== B[i]) return false;
+  return true;
+}
+
+export function verifyCoseSign1Sig(coseSign1: any, leaf: Uint8Array): void {
```
amircheikh, line 241: Not sure if you can type `coseSign1`. Is it the same type that `CBOR.decode()` returns?

packages/crypto/package.json
```diff
@@ -45,12 +45,17 @@
     "@noble/ciphers": "1.3.0",
     "@noble/curves": "1.9.0",
     "@noble/hashes": "1.8.0",
+    "@peculiar/webcrypto": "^1.5.0",
```
amircheikh, line 48: Should probably pin this version just in case. Just remove the '^'. Seems like this can be a good target package for bad people to try and mess with.

packages/crypto/src/proof.ts
```diff
@@ -0,0 +1,263 @@
...
+      if (!c) throw new Error("Invalid certificate data in cabundle");
+      return new x509.X509Certificate(c);
+    });
+    const leaf = new x509.X509Certificate(leafCert);
+
+    // Build path leaf → intermediates → root, with our hardcoded known root certificate
+    const builder = new x509.X509ChainBuilder({
+      certificates: [rootX509, ...intermediatesX509],
+    });
+    const path = await builder.build(leaf);
```
emostov, line 187: It looks like build actually does verification:

Build iterates through chain https://github.com/PeculiarVentures/x509/blob/08f3fd86824d5b890fed349f64554cc2f413c721/src/x509_chain_builder.ts#L45

and calls findIssuer

https://github.com/PeculiarVentures/x509/blob/08f3fd86824d5b890fed349f64554cc2f413c721/src/x509_chain_builder.ts#L61

which verifies the cert with signatureOnly
https://github.com/PeculiarVentures/x509/blob/08f3fd86824d5b890fed349f64554cc2f413c721/src/x509_chain_builder.ts#L86

packages/crypto/src/proof.ts
```diff
@@ -0,0 +1,263 @@
...
+import * as x509 from "@peculiar/x509";
+import { Crypto as PeculiarCrypto } from "@peculiar/webcrypto";
+import { createHash, createPublicKey, createVerify } from "node:crypto";
+
+x509.cryptoProvider.set(new PeculiarCrypto());
+
+export async function verify(
+  appProof: v1AppProof,
+  bootProof: v1BootProof,
+): Promise<void> {
```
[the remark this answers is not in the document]
narimonf: Just a bool isn't enough information in case of failure. But I'm not sure what the right pattern is here -- is it to throw? or is there a pattern of returning errors?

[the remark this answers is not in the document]
narimonf: @amircheikh

packages/crypto/src/__tests__/proof-tests.ts
```diff
@@ -0,0 +1,115 @@
...
+    createdAt: {
+      // 2025-09-16 21:24:01.448145+00
+      seconds: "1758057841",
+      nanos: "448145000",
+    },
+  };
+
+  test("should verify valid app proof signatures", () => {
+    expect(() => verifyAppProofSignature(testAppProof1)).not.toThrow();
+    expect(() => verifyAppProofSignature(testAppProof2)).not.toThrow();
```
r-n-o, line 65: "not throwing" is certainly good, but what is the return value? I think we should have a stricter assertion (`toEqual`?)

packages/crypto/src/__tests__/proof-tests.ts
```diff
@@ -0,0 +1,115 @@
...
+    await expect(verify(testAppProof1, testBootProof1)).resolves.not.toThrow();
+    await expect(verify(testAppProof2, testBootProof2)).resolves.not.toThrow();
+  });
+
+  test("should NOT verify with malformed app proofs", () => {
+    let malformedAppProof2 = { ...testAppProof2 };
+
+    // Wrong publicKey
+    malformedAppProof2.publicKey = testAppProof1.publicKey;
+    expect(() => verifyAppProofSignature(malformedAppProof2)).toThrow();
```
r-n-o, line 78: same thing here. I think we want to ensure:
1) the error messages are reasonable and easy to understand
2) the test is failing for the right reasons! (say this test failed with a good old "undefined is not a function" error...we don't want this test to pass!)

packages/crypto/src/cbor-js.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "cbor-js" {
+  export function encode(value: any): ArrayBuffer;
+  export function decode(data: ArrayBuffer): any;
+}
```
r-n-o, line 4: I don't really get why this is here. Isn't this something we should import instead of vendoring/checking in?

Here they are: https://www.npmjs.com/package/@types/cbor-js

packages/crypto/src/proof.ts
```diff
@@ -0,0 +1,254 @@
...
+import { AWS_ROOT_CERT_PEM, AWS_ROOT_CERT_SHA256 } from "./constants";
+
+x509.cryptoProvider.set(new PeculiarCrypto());
+
+// verify goes through the following verification steps for an app proof & boot proof pair:
+//  - Verify app proof signature
+//  - Verify the boot proof
+//    - Attestation doc was signed by AWS
+//    - Attestation doc's `user_data` is the hash of the qos manifest
+//  - Verify the connection between the app proof & boot proof i.e. that the ephemeral key's match
```
r-n-o, line 18: ephemeral keys match

packages/crypto/src/proof.ts
```diff
@@ -0,0 +1,254 @@
...
+import * as CBOR from "cbor-js";
+import * as x509 from "@peculiar/x509";
+import { Crypto as PeculiarCrypto } from "@peculiar/webcrypto";
+import { createHash, createPublicKey, createVerify } from "node:crypto";
+import { AWS_ROOT_CERT_PEM, AWS_ROOT_CERT_SHA256 } from "./constants";
+
+x509.cryptoProvider.set(new PeculiarCrypto());
+
+// verify goes through the following verification steps for an app proof & boot proof pair:
+//  - Verify app proof signature
```
r-n-o, line 14: (nit) I think the standard for function-level JS docs is:
```ts
/**
 * bla bla bla
 * ...
 */
```

https://jsdoc.app/about-getting-started

packages/crypto/src/proof.ts
```diff
@@ -0,0 +1,254 @@
...
+  const decodedBootProofManifest = Uint8Array.from(
+    atob(bootProof.qosManifestB64)
+      .split("")
+      .map((c) => c.charCodeAt(0)),
+  );
+  const manifestHash = sha256(decodedBootProofManifest);
+  if (!bytesEq(manifestHash, attestationDoc.user_data)) {
+    throw new Error(
+      "attestationDoc's user_data doesn't match the hash of the manifest",
+    );
```
r-n-o, line 63: Would be good to template the actual hashes? (expected vs actual)

(could be useful for debugging)

packages/crypto/src/proof.ts
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  // 3. Verify that all the ephemeral public keys match: app proof, boot proof structure, actual attestation doc
+  const u8 = new Uint8Array(attestationDoc.public_key);
```
r-n-o, line 67: nit: `publicKeyBytes` (instead of "u8")

packages/crypto/src/proof.ts
```diff
@@ -0,0 +1,254 @@
...
+  const attestationPubKey = Buffer.from(u8).toString("hex");
```
r-n-o, line 68: Is `Buffer` going to work in browser contexts? I don't think so, unfortunately:

Image

packages/crypto/src/proof.ts
```diff
@@ -0,0 +1,254 @@
...
+  ) {
+    throw new Error(
+      `Ephemeral pub keys from app proof: ${appProof.publicKey}, boot proof structure ${bootProof.ephemeralPublicKeyHex}, and attestation doc ${attestationPubKey} should all match`,
+    );
+  }
+}
+
+export function verifyAppProofSignature(appProof: v1AppProof): void {
+  if (!appProof) {
+    throw new Error("App proof cannot be null");
```
r-n-o, line 81: nit: "cannot be falsy" (if `!appProof` is true, `appProof` can be `null`, `undefined`, the number 0, the empty string....and more. JS is [whacky](https://dorey.github.io/JavaScript-Equality-Table/))

packages/crypto/src/proof.ts
```diff
@@ -0,0 +1,254 @@
...
+  let publicKeyBytes: Uint8Array;
+  try {
+    publicKeyBytes = uint8ArrayFromHexString(appProof.publicKey);
+  } catch (error) {
+    throw new Error("Failed to decode public key");
+  }
+
+  if (publicKeyBytes.length !== 130) {
+    throw new Error(
+      `Expected 130 bytes (encryption + signing keys), got ${publicKeyBytes.length} bytes`,
```
r-n-o, line 98: (nit) encryption + signing public keys

packages/crypto/src/proof.ts
```diff
@@ -0,0 +1,254 @@
...
+        });
+        if (!ok)
+          throw new Error("Pinned root failed self-signature verification");
+      } else {
+        // Verify signature against issuer
+        const issuer = chain[i + 1];
+        if (!issuer) throw new Error("Issuer can't be null");
+
+        // Attestation docs technically expire after 3 hours, so an app proof generated 3+ hours after an enclave
+        // boots up will fail verification due to certificate expiration. To prevent this failure, we set `signatureOnly: true`.
```
r-n-o, line 205: Worth mentioning that this is okay to do because enclaves are immutable. Meaning that even if the boot proof is technically invalid because the cert is invalid, the code contained within it cannot change.

packages/crypto/src/proof.ts
narimonf: removed all calls

amircheikh (dismissed):

r-n-o (dismissed):
Great job on the proof verification code, looks good! (re-reviewed `packages/core/crypto/*`, had a cursory look at the rest)

One thing for housekeeping purposes: you'll need a changeset to make sure `@turnkey/crypto` gets a minor bump on release, with a changelog entry.

amircheikh (approved):

.changeset/thick-squids-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+"@turnkey/crypto": major
```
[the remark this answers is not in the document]
amircheikh: Mb, should've caught this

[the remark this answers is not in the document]
narimonf: changing this as we speak, I messaged amir about it offline :)

PR #975, opened 2025-09-25
Set up delegated access from the client side

t-vila:
## Did you add a changeset?


examples/with-delegated/src/app/dashboard/page.tsx
```diff
@@ -0,0 +1,331 @@
...
+          return;
+        }
+
+        const embeddedWallets = wallets.filter(
+          (w: Wallet) => w.source === WalletSource.Embedded,
+        );
+
+        const results: AccountsBlock[] = [];
+        for (const w of embeddedWallets) {
+          const accounts = await fetchWalletAccounts({ wallet: w });
```
moeodeh3, line 81: our wallets object from useTurnkey() includes both the wallet and its wallet accounts, so there should be no need to have to fetch here

if you're curios, heres how we fetch wallets in core: https://github.com/tkhq/sdk/blob/4e439bf2973ea13b51d981d7c24a4841d4e5fd5f/packages/core/src/__clients__/core.ts#L1788-L1933

examples/with-delegated/src/app/dashboard/page.tsx
```diff
@@ -0,0 +1,331 @@
...
+  const isEthAddress = (addr: string) =>
+    /^0x[a-fA-F0-9]{40}$/.test(addr.trim());
+
+  // Compressed secp256k1 pubkeys are 33 bytes (66 hex chars)
+  const isHexCompressedPubKey = (key: string) =>
+    /^[0-9a-fA-F]{66}$/.test(key.trim());
+
+  // Delegated Access setup
+  const handleDaSetup = async () => {
+    if (!isHexCompressedPubKey(daPublicKey)) {
```
moeodeh3, line 111: nice! seeing this I am wondering if we should include this check in `fetchOrCreateP256ApiKeyUser()`?

examples/with-delegated/.env.local.example
```diff
@@ -0,0 +1,9 @@
+NEXT_PUBLIC_BASE_URL="https://api.turnkey.com"
+NEXT_PUBLIC_ORGANIZATION_ID="<Turnkey parent organization ID>"
+NEXT_PUBLIC_AUTH_PROXY_CONFIG_ID="<Auth Proxy Config ID>"
+
+# The DA keypir is stored here just for demo purposes
+# In production environments, they should be securely stored within your backend system
```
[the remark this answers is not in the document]
moeodeh3: after looking no that is not the case. IMO we don't really need this header but it would be nice to seperate the client-side vs server-side env vars here

examples/with-delegated/src/app/dashboard/page.tsx
```diff
@@ -0,0 +1,331 @@
...
+          {/* Sub-organization ID */}
+          <p className="mt-2 text-md">
+            <span className="font-medium">Your sub-organization id:</span>{" "}
+            <span className="font-mono">
+              {session?.organizationId ?? "Not found"}
+            </span>
+          </p>
+        </div>
+
+        {/* Embedded Wallet Accounts */}
```
moeodeh3, line 206: IMO it might be also nice to see the users in the subOrg here as well? That way you can visually see the DA user being created. Less important, but also maybe show policies as well? (although maybe that might make the example feel too "clutered")

examples/with-delegated/src/tests/validatePolicy.ts
```diff
@@ -0,0 +1,46 @@
+// This is a simple script to validate the suborganization configured policy
+import { Turnkey } from "@turnkey/sdk-server";
+import * as dotenv from "dotenv";
+import * as path from "path";
+
+// Load environment variables from `.env.local`
+dotenv.config({ path: path.resolve(process.cwd(), ".env.local") });
+
+async function main() {
+  const turnkeyDelegated = new Turnkey({
```
moeodeh3, line 10: it might be worth additionally exposing something like this in a serverless next.js action to the frontend

PR #981, opened 2025-09-28
feat: add `examples/with-porto`

bevensteven:
## Summary & Motivation
https://x.com/gakonst/status/1970295861310824697

Demonstrates upgrading a Turnkey EOA to a Porto account followed by an onchain interaction.

## How I Tested These Changes
```
cd ./examples/with-porto
pnpm i
pnpm start
```

## Did you add a changeset?
No, don't think this needs a changeset since it just adds another example(?)


examples/with-porto/src/viem.ts
```diff
@@ -0,0 +1,134 @@
...
+  });
+
+  // Create Porto client
+  const client = createClient({
+    chain: TARGET_CHAIN, // or your target chain
+    transport: http("https://rpc.porto.sh"),
+  });
+
+  // Create admin key for the upgraded account
+  const adminKey = Key.createSecp256k1({ role: "admin" });
```
jxom, line 56: shouldn't need this if we can use the root key

```suggestion
```

andrewkmin:
a wild 🐐 has appeared

bevensteven:
@andrewkmin ptal 🙇

PR #1044, opened 2025-10-26
fixes: eip-1193-provider should only return ethereum wallet accounts

andrewkmin:
## Summary & Motivation
$title

## How I Tested These Changes
via CSB

## Did you add a changeset?
yes!

### DIFF
--- a/.changeset/curly-windows-help.md
+++ b/.changeset/curly-windows-help.md
@@ -1,0 +1,5 @@
+---
+"@turnkey/eip-1193-provider": patch
+---
+
+Only use relevant (Ethereum) wallet accounts, and better handle multiple chain IDs


packages/eip-1193-provider/src/utils.ts
```diff
@@ -123,7 +122,8 @@ export const validateChain = (
   const { rpcUrls, blockExplorerUrls, chainId, nativeCurrency } = chain;
 
   if (addedChains.some((c) => c.chainId === chainId)) {
```
moeodeh3: can't we just remove this?

this is opinionated, but IMO if someone passes a `chainId` that already exists, but with an invalid RPC, we should throw an error about the RPC being invalid

right now, it'll silently succeeds without any action

moeodeh3 (approved):
small non-blocking nits :)

PR #1050, opened 2025-10-28
Removed peculiar/webcrypto

amircheikh:
## Summary & Motivation
Removed @peculiar/webcrypto as a dep for @turnkey/crypto. We used to include it in the `verify` function because tests were running with an older version of node that required a webcrypto polyfill. Now that we use node v20, we don't need this anymore. 

@peculiar/webcrypto was causing an issue with production builds on react-native since it was trying to `import crypto from 'crypto';` which would fail on react-native. Even tho @peculiar/webcrypto was being dynamically imported to avoid this, some expo production builds will bundle this package anyways and fail to build. On react-native, @peculiar/webcrypto required a webcrypto polyfill anyways to work so removing this package does not negatively impact react-native users.

Now, if you try to use the `verify` function on an environment that does not have webcrypto (old node or react-native), you will simply get this error: ` "Web Crypto API is not available in this environment. You may need to polyfill it."`

## How I Tested These Changes
with-react-native-wallet-kit example using codesandbox build
and
react-wallet-kit example for a sanity check.

## Did you add a changeset?
yes

### DIFF
--- a/.changeset/lovely-streets-like.md
+++ b/.changeset/lovely-streets-like.md
@@ -1,0 +1,5 @@
+---
+"@turnkey/crypto": patch
+---
+
+- Removed `@peculiar/webcrypto` dependancy. This will fix build errors in environments where `webcrypto` is not defined but will still require a polyfill if you use a function where `webcrypto` is required.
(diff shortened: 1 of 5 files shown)


moeodeh3 (approved):
LGTM - might be worth noting in the README that if you want to use `verify()` in React Native or Node < 19, you'll need to include a polyfill

PR #1070, opened 2025-11-05
add with-yield-xyz example

radusandor:
## Summary & Motivation
It was requested by the Yield team to add a cookbook recipe in our docs
## How I Tested These Changes
Tested locally on Base mainnet all 4 scripts end to end.

### DIFF
--- a/examples/with-yield-xyz/src/exit.ts
+++ b/examples/with-yield-xyz/src/exit.ts
@@ -32,7 +32,7 @@ async function main() {
   const provider = new ethers.JsonRpcProvider(process.env.RPC_URL!);
   const connectedSigner = turnkeySigner.connect(provider);
   const exitPayload = {
-    yieldId: process.env.YIELD_ID,
+    yieldId: process.env.YIELD_ID!,
     address: turnkeyAccount.address,
     arguments: { amount: "0.1" },
   };


examples/with-yield-xyz/.env.local.example
```diff
@@ -0,0 +1,8 @@
+TURNKEY_API_PUBLIC_KEY="<Turnkey API Public Key of the root user (that starts with 02 or 03)>"
+TURNKEY_API_PRIVATE_KEY="<Turnkey API Private Key of the root user>"
+TURNKEY_ORG_ID="<Turnkey organization ID>"
+TURNKEY_BASE_URL="https://api.turnkey.com"
+TURNKEY_WALLET_ADDRESS= "<Turnkey organization wallet Base address 0x...>"
+YIELD_ID="base-usdc-gtusdcf-0x236919f11ff9ea9550a4287696c2fc9e18e6e890-4626-vault"
+YIELD_API_KEY="<Yield.xyz project API Key>"
+RPC_URL="<Infura RPC URL for Base mainnet>"
```
t-vila: nit, maybe we should remove Infura :D

examples/with-yield-xyz/src/exit.ts
```diff
@@ -0,0 +1,65 @@
...
+    { baseUrl: process.env.TURNKEY_BASE_URL! },
+    new ApiKeyStamper({
+      apiPublicKey: process.env.TURNKEY_API_PUBLIC_KEY!,
+      apiPrivateKey: process.env.TURNKEY_API_PRIVATE_KEY!,
+    }),
+  );
+
+  // Replace with your wallet address or fetched wallet from Turnkey
+  const turnkeyAccount = {
+    address: process.env.TURNKEY_WALLET_ADDRESS!,
```
t-vila: same here

PR #1203, opened 2026-02-13
added generic client type that only requires specifc methods from our clients, that way version bumps are cross compatible

ethankonk:
## Did you add a changeset?

### DIFF
--- a/packages/viem/src/index.ts
+++ b/packages/viem/src/index.ts
@@ -42,16 +42,12 @@ import type { TurnkeySDKApiTypes as BrowserApiTypes } from "@turnkey/sdk-browser
 import type { TurnkeySDKApiTypes as ServerApiTypes } from "@turnkey/sdk-server";
 import type {
   TGetPrivateKeyBody as SdkTypesGetPrivateKeyBody,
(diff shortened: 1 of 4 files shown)


moeodeh3 (changes requested):

moeodeh3 (dismissed):

packages/viem/src/index.ts
```diff
@@ -35,11 +35,86 @@ import {
...
+ */
+export interface TurnkeyClientInterface {
+  getPrivateKey(input: TGetPrivateKeyBody): Promise<TGetPrivateKeyResponse>;
+  signTransaction(
+    input: TSignTransactionBody,
+  ): Promise<TSignTransactionResponse>;
+  signRawPayload(input: TSignRawPayloadBody): Promise<TSignRawPayloadResponse>;
+}
+
+export type TTurnkeyClient = TurnkeyClientInterface;
```
janjakubnanista, line 113: I think you can get rid of all this code rewriting by doing something like:

```typescript
export type TurnkeyClientInterface= Pick<TurnkeyClient, "getPrivateKey" | "signTransaction" | "signRawPayload">;
```

PR #1237, opened 2026-03-06
added captcha support in the react-wallet-kit

ethankonk:
## Did you add a changeset?


justinformentin:
### PR Review Summary

This PR integrates Cloudflare Turnstile CAPTCHA bot-detection into the `react-wallet-kit` authentication flows. It fetches the Turnstile site key from a new auth-proxy endpoint (`/v1/wallet_kit_client_params`), pre-warms an invisible Turnstile widget in the Provider, and threads a `captchaToken` through every initiation point (`initOtp`, `signUpWithPasskey`, `signUpWithWallet`, OAuth redirects, OTP verification/resend). A second Turnstile widget is rendered inline in the AuthComponent for interaction-only mode when no pre-warmed token is already available.

### Confidence Score: 4/5

Needs changes before merge. The PR contains a build-breaking import error (`React.RefObject` used without a React import in `captcha.ts`), a security issue (captchaToken embedded in the OAuth redirect URL state), a correctness gap (captchaToken not forwarded to `loginWithOtp` so returning users bypass CAPTCHA), and OAuth social buttons are not disabled while awaiting CAPTCHA.

---

### Findings

#### 🔴 P0 — Critical

- **React.RefObject used without React import — build error** (`packages/react-wallet-kit/src/utils/captcha.ts:39`) — `captcha.ts` uses `React.RefObject<TurnstileInstance | null>` but only imports from `@marsidev/react-turnstile`. With `skipLibCheck: true` removed, TypeScript reports `Cannot find name 'React'` — this is the root cause of the failing CI build. Fix: `import type { RefObject } from 'react'` and use `RefObject<TurnstileInstance | null>`.

- **captchaToken written into OAuth redirect URL state** (`packages/react-wallet-kit/src/providers/client/Provider.tsx`) — The captchaToken is embedded in `additionalState` which becomes part of the OAuth redirect URL, exposing it in browser history, referrer headers, and server logs. Fix: store the token in `sessionStorage` before redirect (like the PKCE verifier) and retrieve it on return.

#### 🟠 P1 — Important

- **captchaToken not forwarded to loginWithOtp — returning users bypass CAPTCHA** (`packages/core/src/__clients__/core.ts:1778`) — In `completeOtp`, the login path calls `loginWithOtp` without `captchaToken`. Only the signup path passes it. CAPTCHA is not enforced for returning users.

- **OAuth social buttons not disabled while CAPTCHA pending** (`packages/react-wallet-kit/src/components/auth/index.tsx:318`) — `authEnabled` gates email/SMS/passkey/wallet, but `OAuthButton` has no `disabled` prop and is not guarded by `authEnabled`. Social buttons can be clicked before CAPTCHA resolves. (This is what the author's own `"BAD!"` self-comment was flagging.)

- **authEnabled not reset on Turnstile error** (`packages/react-wallet-kit/src/components/auth/index.tsx:487`) — When `hadTokenOnMount=true`, `authEnabled` starts `true`. If the token expires and `onError` fires, `setAuthEnabled(false)` is never called, leaving buttons enabled.

- **Resend button has no disabled state while awaiting new CAPTCHA** (`packages/react-wallet-kit/src/components/auth/OTP.tsx:88`) — After consuming the token for resend, a second resend before the new challenge completes silently sends no token with no user feedback.

#### 🟡 P2 — Moderate

- **getTurnstileToken/setTurnstileToken exposed on public context** (`packages/react-wallet-kit/src/providers/client/Types.tsx:64`) — Any consumer can read or overwrite the live captcha token via `useTurnkey()`. Mark as `@internal` or pass via props.

- **Pre-warm Turnstile disabled when authenticated — token gap on session renewal** (`packages/react-wallet-kit/src/providers/client/Provider.tsx`) — The invisible pre-warm widget is gated by `authState !== AuthState.Authenticated`, so re-auth flows always start cold.

- **getClientParams called on every config change** (`packages/react-wallet-kit/src/providers/client/Provider.tsx`) — Guard it: only call when `clientParamsRef.current === null`.

- **Codegen and generated base diverge on captchaToken** (`packages/core/scripts/codegen.js`) — `authProxyRequest` has hand-maintained captchaToken support not covered by codegen; a future regeneration could silently lose it.

- **consumeCaptchaToken silently swallows timeout** (`packages/react-wallet-kit/src/utils/captcha.ts:36`) — Timeout returns `{}`, so requests proceed with no token and fail server-side with no client-side indication.

- **No tests for captcha utilities** — `waitForCaptchaToken`, `consumeCaptchaToken`, and CAPTCHA-threaded auth flows have zero test coverage.

#### 🔵 P3 — Nits

- **OTP Turnstile widget missing `id` attribute** (`packages/react-wallet-kit/src/components/auth/OTP.tsx:176`) — Add `id='otp-component-turnstile'` to avoid widget conflicts when AuthComponent and OTP are both mounted.

- **`var` instead of `const` for headers** (`packages/core/src/utils.ts`) — `getClientParams` uses `var headers`, inconsistent with codebase style.

- **OTP digit label doesn't account for alphanumeric OTPs** (`packages/react-wallet-kit/src/components/auth/OTP.tsx:128`) — Says "6-digit code" even when `alphanumeric=true`.

---

### Important Files Changed

| Filename | Overview |
|----------|----------|
| `packages/react-wallet-kit/src/utils/captcha.ts` | New file: `waitForCaptchaToken` + `consumeCaptchaToken` helpers; build-breaking missing React import |
| `packages/react-wallet-kit/src/components/auth/index.tsx` | Adds inline Turnstile widget + `authEnabled` gate; OAuth buttons NOT gated |
| `packages/react-wallet-kit/src/components/auth/OTP.tsx` | Adds Turnstile widget for OTP verify/resend; missing `id`, no resend disabled state |
| `packages/react-wallet-kit/src/providers/client/Provider.tsx` | Fetches client params, pre-warms invisible widget, wires captchaToken into OAuth handlers |
| `packages/react-wallet-kit/src/providers/client/Types.tsx` | Adds `getTurnstileToken`/`setTurnstileToken` to public context |
| `packages/react-wallet-kit/src/types/method-types.ts` | Adds `captchaToken?: string` to all OAuth handler param types |
| `packages/react-wallet-kit/src/types/base.ts` | Adds `turnstileSiteKey?: string` to `TurnkeyProviderConfig` |
| `packages/core/src/__types__/method-types/shared.ts` | Adds `captchaToken?` to `InitOtpParams`, `CompleteOtpParams`, `SignUpWithPasskeyParams`, etc. |
| `packages/core/src/__clients__/core.ts` | Threads captchaToken through all auth methods; missing from `loginWithOtp` |
| `packages/core/src/__generated__/sdk-client-base.ts` | Adds captchaToken to `authProxyRequest` (`X-Captcha-Token` header), new `proxyGetWalletKitClientParams` |
| `packages/core/scripts/codegen.js` | Defines `CAPTCHA_PROTECTED_METHODS`; generates `captchaToken?` for those methods |
| `packages/core/src/utils.ts` | Adds `getClientParams()` for new auth-proxy endpoint |
| `packages/react-wallet-kit/src/utils/oauth/url.ts` | Parses captchaToken from OAuth state string on redirect return |
| `packages/react-wallet-kit/src/utils/oauth/completion.ts` | Threads captchaToken through all OAuth completion flows |
| `packages/sdk-types/src/__generated__/types.ts` | Adds `v1GetWalletKitClientParamsRequest/Response` types |
| `packages/sdk-react/package.json` | Bumps `@types/react` devDep 18.2.14 → 18.3.24 |

### Sequence Diagram

```mermaid
sequenceDiagram
    participant User
    participant AuthComponent
    participant Provider
    participant OTPComponent
    participant TurnkeyCore
    participant AuthProxy
    participant Turnstile as Cloudflare Turnstile

    Provider->>AuthProxy: GET /v1/wallet_kit_client_params
    AuthProxy-->>Provider: { turnstileSiteKey }
    Provider->>Turnstile: render invisible widget (pre-warm)
    Turnstile-->>Provider: onSuccess(token) stored in turnstileTokenRef

    User->>AuthComponent: opens login modal
    AuthComponent->>AuthComponent: hadTokenOnMount = !!getTurnstileToken()
    alt token already in ref
        AuthComponent->>AuthComponent: authEnabled = true
    else no token
        AuthComponent->>Turnstile: render interaction-only widget
        Turnstile-->>AuthComponent: onSuccess(token) setAuthEnabled(true)
    end

    User->>AuthComponent: submits email
    AuthComponent->>AuthComponent: consumeToken() reads and clears token
    AuthComponent->>TurnkeyCore: initOtp({ email, captchaToken })
    TurnkeyCore->>AuthProxy: POST /v1/otp_init_v2 (X-Captcha-Token)
    AuthProxy->>Turnstile: validate token
    AuthProxy-->>TurnkeyCore: { otpId }
    AuthComponent->>OTPComponent: pushPage(OtpVerification)

    User->>OTPComponent: enters OTP code
    OTPComponent->>OTPComponent: consumeToken() from inline widget
    OTPComponent->>TurnkeyCore: completeOtp({ otpCode, captchaToken })
    TurnkeyCore->>AuthProxy: POST /v1/otp_login_v2 or /v1/signup_v2
    AuthProxy-->>TurnkeyCore: session
    TurnkeyCore-->>OTPComponent: auth result
```

packages/react-wallet-kit/src/utils/captcha.ts
```diff
@@ -0,0 +1,48 @@
...
+};
+
+/**
+ * Waits for a captcha token, then consumes it (clears state and resets the widget for a fresh token).
+ * Returns `{ captchaToken }` if a token was available, or `{}` if timed out.
+ */
+export const consumeCaptchaToken = async (
+  getTurnstileToken: () => string | null,
+  setTurnstileToken: (token: string | null) => void,
+  turnstileRef?: React.RefObject<TurnstileInstance | null>,
```
justinformentin: **🔴 P0 Critical — React.RefObject used without React import (build error)**

This file uses `React.RefObject<TurnstileInstance | null>` but never imports `React`. With `skipLibCheck: true` removed in the 'fixed?' commit, TypeScript will report `Cannot find name 'React'` here — this is the root cause of the failing CI build.

Fix: add `import type { RefObject } from 'react';` at the top of the file and change the parameter type:

```ts
// Before
turnstileRef?: React.RefObject<TurnstileInstance | null>,

// After
turnstileRef?: RefObject<TurnstileInstance | null>,
```

packages/react-wallet-kit/src/utils/captcha.ts
```diff
@@ -0,0 +1,48 @@
...
+      }
+    }, interval);
+  });
+};
+
+/**
+ * Waits for a captcha token, then consumes it (clears state and resets the widget for a fresh token).
+ * Returns `{ captchaToken }` if a token was available, or `{}` if timed out.
+ */
+export const consumeCaptchaToken = async (
```
justinformentin, line 38: **🟡 P2 Moderate — consumeCaptchaToken silently swallows token timeout**

When `waitForCaptchaToken` times out, this returns `{}`. Callers spread the result with `...(await consumeToken())`, so requests go out with no `captchaToken`. If the server enforces CAPTCHA, this causes a server-side 4xx with no meaningful client-side indication to the user.

Consider at minimum a `console.warn` on timeout, or returning a sentinel that callers can check to show a user-facing message like 'Please complete the CAPTCHA verification before continuing.'

packages/react-wallet-kit/src/components/auth/OTP.tsx
```diff
@@ -81,6 +91,7 @@ export function OtpVerification(props: OtpVerificationProps) {
       const { otpId, otpEncryptionTargetBundle } = await initOtp({
         otpType,
         contact,
+        ...(await consumeToken()),
```
justinformentin: **🟠 P1 Important — Resend has no disabled/loading state while awaiting new CAPTCHA token**

After `consumeToken()` is called here for resend, the Turnstile widget resets. If the user clicks Resend again before the new challenge completes, `waitForCaptchaToken` times out (5s) and the request fires with no `captchaToken`. There is no loading state, disabled state, or user feedback during this window — the user can silently bypass CAPTCHA protection on resend.

Consider tracking a `captchaReady` state (set `false` on consume, `true` on widget `onSuccess`) and wiring it to the Resend button's `disabled` prop.

packages/react-wallet-kit/src/components/auth/index.tsx
```diff
@@ -375,19 +402,26 @@ export function AuthComponent({
   const methodComponents: Record<string, JSX.Element | null> = {
     socials: oauthBlock,
     email: methods.emailOtpAuthEnabled ? (
-      <EmailInput onContinue={handleEmailSubmit} />
+      <EmailInput onContinue={handleEmailSubmit} disabled={!authEnabled} />
```
justinformentin, line 395: **🟠 P1 Important — OAuth social buttons not disabled while CAPTCHA pending**

The `disabled={!authEnabled}` prop is correctly applied to email/SMS/passkey/wallet inputs here, but `OAuthButton` (used for Google, Apple, Facebook, X, Discord) has no `disabled` prop and is not guarded by `authEnabled`. A user can click any social login button before the Turnstile challenge resolves — `consumeToken()` returns `{}` and the OAuth flow proceeds with no `captchaToken`.

This is what the author's own `"BAD!"` self-review comment was flagging. Add a `disabled` prop to `OAuthButton` and pass `disabled={!authEnabled}` to all entries in `oauthButtonMap`.

packages/react-wallet-kit/src/components/auth/index.tsx
```diff
@@ -450,6 +484,43 @@ export function AuthComponent({
...
+          <Turnstile
+            ref={turnstileRef}
+            id="auth-component-turnstile"
+            siteKey={config.turnstileSiteKey}
+            className="!w-full !block [&>iframe]:!w-full [&>iframe]:!bg-transparent"
+            onSuccess={(token) => {
+              setTurnstileToken(token);
+              setAuthEnabled(true);
+            }}
+            onError={() => {
```
justinformentin: **🟠 P1 Important — authEnabled not reset when Turnstile errors after token was present on mount**

When `hadTokenOnMount=true`, `authEnabled` starts as `true`. If the token expires between mount and submission and `onError` fires here, `setShowTurnstileError(true)` and `setTurnstileToken(null)` are called but `setAuthEnabled(false)` is not. All auth buttons remain enabled while the widget is in an error state.

Add `setAuthEnabled(false)` to this `onError` handler. The `onSuccess` handler already calls `setAuthEnabled(true)`, so this correctly re-enables auth once the user completes the re-challenge.

packages/core/src/__clients__/core.ts
```diff
@@ -1745,6 +1766,7 @@ export class TurnkeyClient {
             ...(createSubOrgParams && { createSubOrgParams }),
             ...(invalidateExisting && { invalidateExisting }),
             ...(sessionKey && { sessionKey }),
+            ...(captchaToken && { captchaToken }),
```
justinformentin, line 1772: **🟠 P1 Important — captchaToken forwarded here for signup, but NOT for login (line ~1778)**

`captchaToken` is correctly passed to `signUpWithOtp` for new users. However, the `else` branch below (line 1778) calls `loginWithOtp` **without** `captchaToken`. CAPTCHA enforcement applies only to signups — returning users logging in via OTP bypass the protection entirely.

Fix: pass `captchaToken` to `loginWithOtp` as well, add it to `LoginWithOtpParams`, and thread it through to `proxyOtpLoginV2`:

```ts
const loginRes = await this.loginWithOtp({
  verificationToken,
  ...(invalidateExisting && { invalidateExisting }),
  ...(sessionKey && { sessionKey }),
  ...(captchaToken && { captchaToken }),  // add this
});
```

packages/react-wallet-kit/src/providers/client/Types.tsx
```diff
@@ -820,6 +820,11 @@ export interface ClientContextType
   handleSendErc20Transfer: (
     params: HandleSendErc20TransferParams,
   ) => Promise<void>;
+
+  /** @internal Get the current Turnstile captcha token */
+  getTurnstileToken: () => string | null;
```
justinformentin, line 825: **🟡 P2 Moderate — getTurnstileToken/setTurnstileToken exposed on public context**

Both accessors are now part of `ClientContextType` and accessible via `useTurnkey()`. Any SDK consumer or third-party script can call `getTurnstileToken()` to read (and potentially replay) the live Turnstile token, or call `setTurnstileToken()` to inject an arbitrary value and bypass bot detection.

Consider marking both as `/** @internal */` in JSDoc to signal they are not public API, or removing them from the context entirely and passing them as props to only the components that need them (`AuthComponent`, `OtpVerification`).

packages/react-wallet-kit/src/components/auth/OTP.tsx
```diff
@@ -155,6 +166,37 @@ export function OtpVerification(props: OtpVerificationProps) {
...
         </div>
       )}
+      {config?.turnstileSiteKey && !submitting && (
+        <div className="mt-3 flex flex-col text-left w-full">
+          {showTurnstilePrompt && (
+            <p className="text-icon-text-light/70 dark:text-icon-text-dark/70 text-sm mb-0.5">
+              Let us know you're human
+            </p>
+          )}
+          <Turnstile
```
justinformentin: **🔵 P3 Nit — OTP Turnstile widget missing `id` attribute**

The AuthComponent Turnstile has `id="auth-component-turnstile"`, but this widget has no `id`. If both components are simultaneously mounted on the modal page stack (which is possible since OTP is pushed onto the stack while AuthComponent may still be in the stack), two widgets without distinct ids could conflict in how `@marsidev/react-turnstile` mounts/unmounts the iframe.

Add `id="otp-component-turnstile"` to this `<Turnstile>` element.

justinformentin: Ignore. https://github.com/tkhq/sdk/pull/1422 changes the Turnstile Component.

packages/react-wallet-kit/src/components/auth/index.tsx
ethankonk: This one is a bit of a weird case, we only actually require captchas on sign ups and OTP requests, if the captcha widget errors out, logging in via OAuth. passkey, and wallet are still doable. Do we want to block _all_ operations if this is the case?

Personally I don't think we should..

packages/core/src/__clients__/core.ts
ethankonk: This is expected, we only require captcha tokens on Sign up and OTP requests

packages/react-wallet-kit/src/providers/client/Types.tsx
ethankonk: I don't think this is a real issue, Turnstile tokens are one time use and validated server side by Cloudflare, a arbitrary token won't pass server side validation failing the request

justinformentin (dismissed):
LGTM

justinformentin (approved):
LGTM

PR #1240, opened 2026-03-18
MFA and Scoped Sessions for SDK

amircheikh:
## Summary & Motivation
MFA and scoped sessions SDK integrations

Includes:
- A way to pause execution and await an MFA approval via `onMfaRequired` callback in `httpClient`
- Configurable MFA handler (changes the `onMfaRequired`) in `@turnkey/react-(native-)wallet-kit`
- Attested stamper in `@turnkey/core` that gets auto-populated with an `attestedIdentity` after calling `verifyOtp` or a `handlePROVIDEROauth` function
- Some tests in `react-wallet-kit-playground` example
- Started `with-mfa-scoped-sessions` example

## How I Tested These Changes

locally
## Did you add a changeset?

noooooooooo


examples/demos/with-mfa-and-scoped-sessions/src/app/scenarios/Scenario2.tsx
```diff
@@ -0,0 +1,42 @@
...
+export const SESSION_KEY = "scenario-2";
+
+export default function Scenario2() {
+  const { handleLogin, logout, allSessions, clientState } = useTurnkey();
+  const session = allSessions?.[SESSION_KEY];
+
+  return (
+    <div className="flex flex-col items-center justify-center gap-4">
+      <h2 className="text-xl font-semibold">Scenario 2</h2>
+      <p className="text-sm text-gray-500">Placeholder text for scenario 2.</p>
```
moeodeh3, line 14: thank you

examples/demos/with-mfa-and-scoped-sessions/src/app/scenarios/Scenario3.tsx
```diff
@@ -0,0 +1,42 @@
...
+export const SESSION_KEY = "scenario-3";
+
+export default function Scenario3() {
+  const { handleLogin, logout, allSessions, clientState } = useTurnkey();
+  const session = allSessions?.[SESSION_KEY];
+
+  return (
+    <div className="flex flex-col items-center justify-center gap-4">
+      <h2 className="text-xl font-semibold">Scenario 3</h2>
+      <p className="text-sm text-gray-500">Placeholder text for scenario 3.</p>
```
moeodeh3, line 14: really setting us up for success! 😜

examples/demos/with-mfa-and-scoped-sessions/README.md
```diff
@@ -0,0 +1 @@
+Hello my name is readme
```
moeodeh3, line 1: hello readme, my name is updateme!

packages/core/scripts/codegen.js
```diff
@@ -377,6 +385,73 @@ const generateSDKClientFromSwagger = async (
...
+     * @internal
+     */
+    private async handleMfaIfNeeded(
+        activityData: TActivityResponse,
+        stampWith?: StamperType
+    ): Promise<TActivityResponse> {
+        const AUTHENTICATORS_NEEDED: TActivityStatus = "ACTIVITY_STATUS_AUTHENTICATORS_NEEDED";
+        
+        // We need to check for CONSENSUS_NEEDED as well, since that can also involve MFA (user is NOT the proposer but has MFA).
+        const CONSENSUS_NEEDED: TActivityStatus = "ACTIVITY_STATUS_CONSENSUS_NEEDED";
```
moeodeh3: nit: having to keep doing this tells me that we need an enum in `@turnkey/sdk-types` of activity statuses

packages/core/scripts/codegen.js
```diff
@@ -377,6 +385,73 @@ const generateSDKClientFromSwagger = async (
...
+        return activityData;
+        }
+
+        // Fetch the MFA statuses for this activity.
+        const { mfaStatuses } = await this.getMfaStatus(
+        {
+            activityId: activityData.activity.id,
+            organizationId: activityData.activity.organizationId,
+        },
+        stampWith
```
moeodeh3, line 539: just flagging that this will be a bit awkward for passkeys and wallets because requests will require two stamps whenever MFA or consensus is needed

technically we already have this problem today with polling, it just hasn't really been a problem yet because we almost never have to poll (like 99% of the time)

if you want to stamp the request with a passkey or wallet but handle the MFA step using your session, that's still possible by doing:

```
stamp<Activity>()
sendSignedRequest()
```

packages/core/src/__clients__/core.ts
```diff
@@ -324,6 +331,71 @@ export class TurnkeyClient {
...
+          this.attestedStamper.setAttestedIdentity(oidcToken);
+          this.attestedStamper.setScheme("STAMP_ATTESTED_SCHEME_P256_OIDC");
+        } else {
+          // If neither token is provided, clear the attestation identity and public key
+          this.attestedStamper.clearAttestedIdentity();
+          this.attestedStamper.clearPublicKey();
+        }
+
+        // Set the public key that signs the attested stamp.
+        if (publicKey !== undefined) {
```
moeodeh3: we have some weird edge cases around `publicKey`. For example: you can set the token without setting the `publicKey`, or clear the `publicKey` without clearing the token. That doesn't really make sense since the two are coupled

maybe we do this instead:

```
overrideAttestedStamper = async (
    params: OverrideAttestedStamperParams,
  ): Promise<void> => {
    return withTurnkeyErrorHandling(
      async () => {
        if (!this.attestedStamper) {
          throw new TurnkeyError(
            "Attested stamper is not initialized",
            TurnkeyErrorCodes.INTERNAL_ERROR,
          );
        }

        const { verificationToken, oidcToken, publicKey } = params;

        // we enforce that you cannot set both a verificationToken and an oidcToken
        if (verificationToken && oidcToken) {
          throw new TurnkeyError(
            "Cannot set both verificationToken and oidcToken. Please provide only one.",
            TurnkeyErrorCodes.INVALID_REQUEST,
          );
        }

        // we require a publicKey if either a verificationToken or an oidcToken is provided
        if ((verificationToken || oidcToken) && !publicKey) {
          throw new TurnkeyError(
            "A publicKey must be provided when setting a verificationToken or oidcToken.",
            TurnkeyErrorCodes.INVALID_REQUEST,
          );
        }

        if (verificationToken) {
          this.attestedStamper.setAttestedIdentity(verificationToken);
          this.attestedStamper.setScheme(
            "STAMP_ATTESTED_SCHEME_P256_VERIFICATION_TOKEN",
          );
          this.attestedStamper.setPublicKey(publicKey!);
        } else if (oidcToken) {
          this.attestedStamper.setAttestedIdentity(oidcToken);
          this.attestedStamper.setScheme("STAMP_ATTESTED_SCHEME_P256_OIDC");
          this.attestedStamper.setPublicKey(publicKey!);
        } else {
          this.attestedStamper.clearAttestedIdentity();
          this.attestedStamper.clearPublicKey();
        }
      },
      {
        errorMessage: "Failed to override attested stamper",
        errorCode: TurnkeyErrorCodes.INTERNAL_ERROR,
      },
    );
  };
 ```
fixes the issue and is a lot easier to read too!

File: packages/core/src/__clients__/core.ts
Comment (moeodeh3) [inline packages/core/src/__clients__/core.ts:5060]: I think there's an issue here: if you create a session using a `sessionProfile`, our session management system will create a full read-write session when it's time to refresh it

I think `refreshSession()` should recreate the same type of session, right now its secretly upgrading. I think it should be done here (we update the `Session` type in `@turnkey/sdk-types`). We should not rely on callers to provide it, where we secretly upgrade them if they don't

I think that means we shouldn't need a `sessionProfileId` param here

Comment (moeodeh3) [inline packages/core/src/__clients__/core.ts:1581]: hate to expand scope: but we don't support session profiles for otp or oauth...

maybe we just replace the otpLogin with stampLogin now 🤔

File: packages/core/src/__stampers__/attested/base.ts [hunk 7]
```diff
@@ -0,0 +1,88 @@
+import type { TStamp, TStamper } from "@turnkey/sdk-types";
+import { TurnkeyError, TurnkeyErrorCodes } from "@turnkey/sdk-types";
+import type { CrossPlatformApiKeyStamper } from "../api/base";
+import { SignatureFormat } from "@turnkey/api-key-stamper";
+import { stringToBase64urlString } from "@turnkey/encoding";
+
+type AttestedScheme =
```
moeodeh3: nit: enum?

packages/core/src/__stampers__/attested/base.ts
```diff
@@ -0,0 +1,88 @@
...
+  public attestedIdentity?: string | undefined;
+  public publicKey?: string | undefined;
+  private scheme: AttestedScheme =
+    "STAMP_ATTESTED_SCHEME_P256_VERIFICATION_TOKEN";
+
+  constructor(apiKeyStamper: CrossPlatformApiKeyStamper) {
+    this.apiKeyStamper = apiKeyStamper;
+  }
+
+  setAttestedIdentity(attestedIdentity: string): void {
```
moeodeh3: hmm, `overrideAttestedStamper()` always sets these 3 fields together, why not set them all together then?

something like
```
configure(params: { attestedIdentity: string; publicKey: string; scheme: AttestedScheme }): void {
  this.attestedIdentity = params.attestedIdentity;
  this.publicKey = params.publicKey;
  this.scheme = params.scheme;
}
```

then in `overrideAttestedStamper()` we can do:
```
if (verificationToken) {
  this.attestedStamper.configure({
    attestedIdentity: verificationToken,
    publicKey: publicKey!,
    scheme: "STAMP_ATTESTED_SCHEME_P256_VERIFICATION_TOKEN",
  });
} else if (oidcToken) {
  this.attestedStamper.configure({
    attestedIdentity: oidcToken,
    publicKey: publicKey!,
    scheme: "STAMP_ATTESTED_SCHEME_P256_OIDC",
  });
} else {
  this.attestedStamper.clear();
}
```

packages/core/src/__types__/method-types/shared.ts
```diff
@@ -451,6 +471,7 @@ export type RefreshSessionParams = {
   publicKey?: string;
   sessionKey?: string;
   invalidateExisitng?: boolean;
+  sessionProfileId?: string;
```
moeodeh3: shouldn't need this. `refreshSession()` should just match whatever type of session it's refreshing automatically

packages/core/src/__types__/method-types/shared.ts
```diff
@@ -97,13 +108,15 @@ export type LoginWithWalletParams = {
   sessionKey?: string;
   expirationSeconds?: string;
   organizationId?: string;
+  sessionProfileId?: string;
```
moeodeh3, line 116: do we need this to be returned? the caller is passing it in already

packages/http/src/async.ts
```diff
@@ -28,6 +28,11 @@ export function withAsyncPolling<
         case "ACTIVITY_STATUS_COMPLETED": {
           return activity;
         }
+        case "ACTIVITY_STATUS_AUTHENTICATORS_NEEDED": {
```
moeodeh3: I think we should treat this like we do consensus and throw an error

packages/http/src/async.ts
```diff
@@ -121,6 +126,11 @@ export function createActivityPoller<
...
```
moeodeh3: same thing here

packages/react-wallet-kit/src/providers/client/Provider.tsx
```diff
@@ -1785,6 +1805,7 @@ export const ClientProvider: React.FC<ClientProviderProps> = ({
           callbacks?.beforeSessionExpiry?.({ sessionKey });
 
           if (masterConfig?.auth?.autoRefreshSession) {
+            // TODO (Amir/Moe): Once the sessionProfileId is stored in the session, we need to pass it in here so that we get a new session with the same sessionProfileId
```
moeodeh3: called this out later, except we shouldn't pass it in I think? should happen inside of core's `refreshSession()`

moeodeh3: (also we forgot about RN)

packages/react-wallet-kit/src/providers/client/Provider.tsx
```diff
@@ -3573,6 +3606,14 @@ export const ClientProvider: React.FC<ClientProviderProps> = ({
                 authWindow.close();
                 clearInterval(interval);
 
+                // Set the attested stamper with the oidcToken
+                if (result.idToken) {
```
moeodeh3: this doesn't make sense for discord? I think we should remove it because:
- its a duplicate, we already do this inside of `exchangeCodeForToken`
- oauth2's response doesn't have an idToken, so this never will run

packages/react-wallet-kit/src/providers/client/Provider.tsx
```diff
@@ -3719,6 +3768,14 @@ export const ClientProvider: React.FC<ClientProviderProps> = ({
...
```
moeodeh3: same thing here for twitter

packages/react-wallet-kit/src/providers/client/Provider.tsx
```diff
@@ -4129,6 +4210,14 @@ export const ClientProvider: React.FC<ClientProviderProps> = ({
...
```
moeodeh3: same thing for facebook

packages/react-wallet-kit/src/types/method-types.ts
```diff
@@ -314,3 +315,8 @@ export type HandleSendErc20TransferParams = {
   icon?: React.ReactNode;
   stampWith?: StamperType;
 };
+
+export type HandleMfaParams = {
```
moeodeh3: I don't think this is used, can we delete this?

packages/sdk-types/src/index.ts
```diff
@@ -120,6 +120,7 @@ export enum TurnkeyErrorCodes {
   SWITCH_WALLET_CHAIN_ERROR = "SWITCH_WALLET_CHAIN_ERROR",
   ONRAMP_ERROR = "ONRAMP_ERROR",
   MAX_OTP_INITIATED_ERROR = "MAX_OTP_INITIATED_ERROR",
+  MFA_COMPLETION_ERROR = "MFA_COMPLETION_ERROR",
```
moeodeh3: I don't think this is used either, lets delete it

examples/demos/with-mfa-and-scoped-sessions/README.md
amircheikh: ladnon

packages/core/src/__types__/method-types/shared.ts
amircheikh: this is params

packages/http/src/async.ts
amircheikh: Ah. It was there a few lines below but I have this duplicate on accident.

packages/core/src/__types__/method-types/shared.ts
moeodeh3: github diff tricked me

moeodeh3 (approved):

hadrelandon (approved):
LGTM!

PR #1249, opened 2026-03-21
Add @turnkey/agent-auth package (Phase 1/3)

natefikru:
## PRD

[Agent Identity Provisioning API — Full PRD](https://gist.github.com/natefikru/b4b75aabfe0531659c2dcf7d02617dd6)

## Summary

New SDK package (`@turnkey/agent-auth`) that provisions isolated, policy-gated cryptographic identities for AI agents using existing Turnkey primitives. No backend changes required.

**Two main functions:**
- `createAgentSession()`: Provisions a complete agent identity in 4-5 API calls
- `deleteAgentSession()`: Tears down an agent session and all associated resources

### How it works

1. Creates an isolated sub-organization for each agent
2. Creates a root admin user (orchestrator holds key for management/deletion)
3. Creates a non-root agent user with a two-key pattern:
   - **Anchor key** (non-expiring): satisfies Turnkey's Notarizer validation, private key discarded immediately
   - **Session key** (expiring): given to the agent, TTL from request, enforced by auth middleware
4. Creates ALLOW policies scoped to the agent user (implicit deny for everything else)
5. Optionally creates wallet accounts (generic config with presets for git/JWT/ETH signing)
6. Optionally exports wallet keys via HPKE

### Architecture

```
Parent Org (customer's Turnkey org)
  +-- Sub-Org: agent-session-001
        +-- Root User: admin (orchestrator, non-expiring, lifecycle = sub-org deletion)
        +-- Non-Root User: agent-001
        |     +-- Anchor API key (non-expiring, satisfies Notarizer, private key discarded)
        |     +-- Session API key (expiring, TTL from request, given to agent)
        +-- Wallet: signing-keys (configurable accounts via presets)
        +-- Policy: allow-sign-raw-payload (scoped to agent-001)
        +-- Default: implicit deny for everything else
```

### Security model

- Agents receive **non-root user** credentials. Non-root users go through full UMP policy evaluation in the enclave with no bypass path.
- Root quorum bypass (which skips policies entirely) only applies to root users, never the agent.
- Only ALLOW policies are created. DenyExplicit has absolute precedence over Allow, so a deny-all policy would block everything including allow rules. Implicit deny handles the rest.
- Agent session key has enforced TTL via Turnkey's auth middleware.
- Sub-org isolation prevents cross-agent access (structurally impossible).

### Key discoveries from implementation

1. **Notarizer requires non-expiring credentials**: Every user must have at least one "long-lived" API key (no expirationSeconds). Solved with the two-key anchor pattern.
2. **Sub-org deletion requires sub-org auth**: Parent org has read-only access. Deletion must be initiated by the sub-org's own root user. The admin key is returned in the result for this purpose.
3. **DenyExplicit always wins**: Cannot use deny-all + allow-specific pattern. Use only ALLOW policies.
4. **Ed25519 requires HASH_FUNCTION_NOT_APPLICABLE**: Per RFC 8032, configurable hashing is not supported.
5. **SDK CJS exports use `Turnkey` not `TurnkeyServerSDK`**: Dynamic imports use fallback pattern.

## Test plan

**Unit tests (36 passing):**
- [x] Full orchestration: sub-org creation, user creation, session key, policies
- [x] Two-key pattern: anchor key non-expiring, session key with TTL
- [x] Error cleanup: createUsers, createApiKeys, createPolicies failures all trigger sub-org deletion
- [x] Cleanup failure swallowed, original error surfaces
- [x] Two-client pattern: parent org vs sub-org admin
- [x] HPKE export with uncompressed public key
- [x] Export failure non-fatal
- [x] Multi-curve accounts (P256 + Ed25519 + secp256k1)
- [x] Custom policies with placeholder resolution
- [x] Edge cases: empty userIds, undefined policyIds, apiBaseUrl override, fewer addresses than accounts
- [x] Delete: admin client creation, deleteWithoutExport
- [x] Presets: all three with defaults and options
- [x] Policies: signing policy, transaction policy, placeholder resolution

**Smoke test (16/16 passing against local Turnkey infra):**
- [x] Full session lifecycle (create, use, delete)
- [x] P256 signing (JWT) works
- [x] Ed25519 signing (git) works
- [x] HPKE export bundle present
- [x] createUsers blocked by implicit deny (policy enforcement verified)
- [x] Session deletion works
- [x] Double-delete fails gracefully
- [x] Agent key revoked after delete

**Build verification:**
- [x] `pnpm run typecheck` passes (tsc --noEmit, strict mode)
- [x] `pnpm run build` passes (rollup, dual ESM/CJS output)


natefikru:
Closing

PR #1250, opened 2026-03-22
SDK V2 - Breaking changes compilation

moeodeh3:
## Summary & Motivation

Originally built off https://github.com/tkhq/sdk/pull/1221

This PR includes a whole set of breaking changes which results in a major bump to all of our SDKs

- API sync: Breaking OtpE activities are now defaulted in all generated clients across all SDKs
- Core/React/React-Native wallet kit: Secondary client ids are now supported which uses the new `oauthClaims` to insert additional unverified clientIds at suborg creation / oauth adding time. Configs and input params have been modified accordingly.
- React-native wallet kit: Apple native oauth is now supported. The old web flow is deprecated but the function is left in for legacy flows. The config and input params for Apple oauth is now slightly different from the other providers, since we need the serviceID and bundleID to support iOS and Android
- Core/React/React-native wallet kit: `sendSignedRequest` helper has been removed since it was deprecated before. `httpClient.sendSignedRequest` replaces this.
- React wallet kit: Structure of the config has changed slightly by moving auth method settings to the UI section.

All of these are detailed in the changesets

## How I Tested These Changes
locally

## Did you add a changeset?

YES


.changeset/tasty-dingos-relate.md
```diff
@@ -0,0 +1,183 @@
...
+**What changed:** `oidcToken` and `oidcClaims` are now a oneOf union — you must provide exactly one.
+
+```ts
+// before — v1OauthProviderParams
+{
+  providerName: string;
+  oidcToken: string;
+}
+
+// after — v1OauthProviderParamsV2
```
andrewkmin: we could add a note explicitly mentioning that this change feeds into the changes below as well

packages/sdk-browser/src/__inputs__/public_api.types.ts
```diff
@@ -1650,21 +1742,27 @@ export type definitions = {
...
      */
     nonce?: number;
     /** @description Optional encrypted pull secret to authorize Turnkey to pull the pivot container image. If your image is public, leave this empty. */
     pivotContainerEncryptedPullSecret?: string;
-    /** @description Optional encrypted pull secret to authorize Turnkey to pull the host container image. If your image is public, leave this empty. */
-    hostContainerEncryptedPullSecret?: string;
+    /** @description Optional flag to indicate whether to deploy the TVC app in debug mode, which includes additional logging and debugging tools. Default is false. */
+    debugMode?: boolean;
+    /** @description Heath check type (TVC_HEALTH_CHECK_TYPE_HTTP or TVC_HEALTH_CHECK_TYPE_GRPC). HTTP health checks are made with a GET request on /health, and gRPC health checks follow the standard gRPC health checking protocol. */
+    healthCheckType: definitions["v1TvcHealthCheckType"];
```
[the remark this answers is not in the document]
moeodeh3: fixed in mono

packages/core/scripts/codegen.js
```diff
@@ -84,14 +84,24 @@ const VERSIONED_ACTIVITY_TYPES = {
...
+    "ACTIVITY_TYPE_OTP_LOGIN_V2",
+    "v1OtpLoginIntentV2",
+    "v1OtpLoginResult",
   ],
   ACTIVITY_TYPE_CREATE_OAUTH_PROVIDERS: [
-    "ACTIVITY_TYPE_CREATE_OAUTH_PROVIDERS",
-    "v1CreateOauthProvidersIntent",
-    "v1CreateOauthProvidersResult",
+    "ACTIVITY_TYPE_CREATE_OAUTH_PROVIDERS_V2",
+    "v1CreateOAuthProvidersIntentV2",
```
ethankonk: `v1CreateOauthProvidersIntentV2` + `v1CreateOauthProvidersResultV2`

packages/core/scripts/codegen.js
```diff
@@ -547,15 +557,7 @@ const generateSDKClientFromSwagger = async (
         );
       }
 
-      // If this activity type is capped at a specific version via VERSIONED_ACTIVITY_TYPES,
-      // use the capped result type name instead of the latest from swagger
-      const cappedResultType =
-        VERSIONED_ACTIVITY_TYPES[unversionedActivityType]?.[2];
-      const versionedMethodName = cappedResultType
-        ? cappedResultType
```
ethankonk, line 598: This needs to be kept in!

packages/core/src/__clients__/core.ts
```diff
@@ -1342,20 +1368,18 @@ export class TurnkeyClient {
...
 
     return withTurnkeyErrorHandling(
       async () => {
-        const { message, publicKey: clientSignaturePublicKey } =
-          getClientSignatureMessageForLogin({
-            verificationToken,
-            sessionPublicKey: publicKey!,
-          });
+        const { message, publicKey: verificationPublicKey } =
+          getClientSignatureMessageForLogin({ verificationToken });
```
amircheikh, line 1377: WARNING! `VerificationToken` type needs to be updated with the organizationId field.

Right now it looks like this:

```ts
export type VerificationToken = {
  contact: string;
  exp: number;
  id: string;
  public_key?: string;
  verification_type: string;
};
```

it needs to look like this:

```ts
export type VerificationToken = {
  contact: string;
  exp: number;
  id: string;
  public_key?: string;
  verification_type: string;
  organization_id: string;
};
```

DO NOT FORGET!

packages/react-native-wallet-kit/src/types/base.ts
```diff
@@ -14,7 +14,9 @@ import type {
 export type OauthProviderConfig =
   | boolean
   | {
-      clientId?: string;
+      // TODO (Amir/Moe): Should we separate this for each provider? We need to include comments somehow since different providers may work differently
```
moeodeh3: yes, google is going to need two primary clientIds: ios and web

packages/core/src/__clients__/core.ts
moeodeh3: yes but it actually needs to look like:

```
export type VerificationToken = {
  contact: string;
  exp: number;
  id: string;
  public_key: string;
  verification_type: string;
  organization_id: string;
};
```

packages/core/src/__clients__/core.ts
```diff
@@ -606,7 +605,7 @@ export class TurnkeyClient {
 
         const sessionResponse = await this.httpClient.stampLogin({
           publicKey: newGeneratedKeyPair!,
-          organizationId: organizationId ?? this.config.organizationId,
+          organizationId: this.config.organizationId,
```
moeodeh3: for signup we can do `res.organizationId` since we just created the suborg!

packages/react-wallet-kit/src/providers/client/Provider.tsx
```diff
@@ -474,7 +502,10 @@ export const ClientProvider: React.FC<ClientProviderProps> = ({
 
       // FACEBOOK
       if (provider === OAuthProviders.FACEBOOK) {
-        const clientId = masterConfig?.auth?.oauthConfig?.facebookClientId;
+        const clientId =
```
moeodeh3: when:
- `openInPage` is set to false, and
- someone called `handleFacebookOauth` and passed a primary or secondaryClientId(s), then these would get ignored and we'd default back to the master config params!

packages/react-wallet-kit/src/providers/client/Provider.tsx
```diff
@@ -497,14 +528,18 @@ export const ClientProvider: React.FC<ClientProviderProps> = ({
...
           );
         }
         return;
       }
 
       // DISCORD
       if (provider === OAuthProviders.DISCORD) {
-        const clientId = masterConfig?.auth?.oauthConfig?.discordClientId;
+        const clientId =
+          masterConfig?.auth?.oauthConfig?.discord?.primaryClientId;
```
moeodeh3: same thing here

packages/react-native-wallet-kit/src/utils/oauth.ts
```diff
@@ -697,3 +703,22 @@ export async function exchangeCodeForToken(
 
   return await response.json();
 }
+
+/**
+ * Builds secondary OAuth provider entries from secondary client IDs.
+ * Decodes the OIDC token to extract `iss` and `sub`, then creates
+ * an `oidcClaims`-based provider entry for each secondary client ID (used as `aud`).
+ */
+export function buildSecondaryOauthProviders(
```
moeodeh3: if we have to define this in both `@turnkey/react-wallet-kit` and `@turnkey/react-native-wallet-kit` then why don't we just define it in `@turnkey/core` 🤔

packages/react-wallet-kit/src/providers/client/Provider.tsx
moeodeh3: for all of them

packages/react-wallet-kit/src/providers/client/Provider.tsx
```diff
@@ -5048,6 +5259,25 @@ export const ClientProvider: React.FC<ClientProviderProps> = ({
         successPageDuration = 2000,
         openInPage = masterConfig?.auth?.oauthConfig?.openOauthInPage ?? false,
       } = params;
+
+      // Resolve the provider's per-provider config, then apply per-call overrides.
+      const providerConfig =
```
moeodeh3: there _has_ to be a cleaner way to do this, maybe a switch statement?

packages/core/src/__clients__/core.ts
moeodeh3: fixed this

ethankonk (approved):

PR #1268, opened 2026-04-09
Spark Signing Example: SO Authentication and Token Transfer

omkarshanbhag:
## Summary & Motivation

This PR demonstrates support for Spark address derivations and Schnorr signing via the `SIGN_RAW_PAYLOAD_V2` endpoint. 

The example goes through the following flows:
- Wallet initialization (including SO authentication under the hood -- requires Turnkey Schnorr signature) 
- Token minting 
- Token transfer (requires Turnkey Schnorr signature)

## How I Tested These Changes
- Tested example 
- Example regtest transaction: https://sparkscan.io/token/btknrt1s0yxfk4nhvzgf9q66z90ksfdwr2mxf0h3v7x53f7x4tqyprzmhlsgq7xkm?network=regtest

## Did you add a changeset?


examples/with-spark-schnorr/README.md
```diff
@@ -0,0 +1,63 @@
...
+```bash
+cp .env.local.example .env.local
+```
+
+| Variable                                                         | Required | Notes                                                |
+| ---------------------------------------------------------------- | -------- | ---------------------------------------------------- |
+| `MNEMONIC`                                                       | No       | Leave blank to auto-generate                         |
+| `SPARK_NETWORK`                                                  | No       | `REGTEST` (default) or `MAINNET`                     |
+| `RECEIVER_SPARK_ADDRESS`                                         | No       | Spark transfer target                                |
+| `TRANSFER_AMOUNT_SATS`                                           | No       | Default 1000                                         |
```
taylorjdawson: Same here, I think it should be `TRANSFER_AMOUNT`

examples/with-spark-schnorr/.env.local.example
```diff
@@ -0,0 +1,40 @@
...
+# Fund the P2TR deposit address printed at startup, then fill these in.
+# -------------------------------------------------------------------
+UTXO_TXID=
+UTXO_VOUT=0
+UTXO_VALUE=
+DESTINATION_ADDRESS=
+FEE_SATS=300
+
+# Amount in sats to send in the Spark transfer test (default: 1000)
+TRANSFER_AMOUNT_SATS=1000
```
taylorjdawson: I noticed that this env var, `TRANSFER_AMOUNT_SATS, isn't used in the code but rather the code reads the `TRANSFER_AMOUNT` var

examples/with-spark-schnorr/.env.local.example
```diff
@@ -0,0 +1,40 @@
...
+
+# -------------------------------------------------------------------
+# Optional: Bitcoin P2TR transaction signing test (regtest)
+# Fund the P2TR deposit address printed at startup, then fill these in.
+# -------------------------------------------------------------------
+UTXO_TXID=
+UTXO_VOUT=0
+UTXO_VALUE=
+DESTINATION_ADDRESS=
+FEE_SATS=300
```
taylorjdawson: I don't see these used in the code either, should we keep them or remove them?

examples/with-spark-schnorr/README.md
```diff
@@ -0,0 +1,63 @@
+# Example: `with-spark-schnorr`
+
+Tests native Schnorr signing via the [Spark SDK](https://spark.money) (`DefaultSparkSigner`) and optionally compares the output against a Turnkey-backed signer — useful for verifying that a custom `SparkSigner` implementation produces identical signatures.
+
+The script covers:
+
+1. **Mnemonic → Spark keys**: derives the identity key and deposit (P2TR) key from a BIP-39 mnemonic using `DefaultSparkSigner`.
+2. **Raw Schnorr (identity key)**: calls `signSchnorrWithIdentityKey` and optionally compares against `Turnkey.signRawPayload`.
+3. **Bitcoin P2TR transaction (deposit key)**: if a regtest UTXO is supplied, builds a `@scure/btc-signer` transaction, computes the sighash with `getSigHashFromTx`, and signs with `signTransactionIndex`. The resulting hex can be broadcast via `bitcoin-cli`.
```
taylorjdawson: From the code it looks like `signTransactionIndex` isn't implemented, is that work planned? if not we should update the readme to reflect what's in the code

examples/with-spark-schnorr/src/index.ts
```diff
@@ -0,0 +1,186 @@
...
+  const network = (process.env.SPARK_NETWORK ?? "REGTEST") as SparkNetwork;
+
+  const requiredVars = [
+    "API_PUBLIC_KEY",
+    "API_PRIVATE_KEY",
+    "ORGANIZATION_ID",
+    "TURNKEY_IDENTITY_ADDRESS",
+    "IDENTITY_PUBLIC_KEY_HEX",
+    "RECEIVER_SPARK_ADDRESS",
+  ] as const;
```
taylorjdawson, line 53: I noticed that these env vars are denoted as optional in the README and in the .env.local.example
but here they are required
we should reconcile that discrepancy

taylorjdawson (dismissed):

examples/with-spark-schnorr/src/index.ts
```diff
@@ -0,0 +1,186 @@
...
+  const maxSupply = BigInt(process.env.TOKEN_SUPPLY ?? "1000000");
+  const mintAmount = BigInt(process.env.MINT_AMOUNT ?? String(maxSupply));
+  const transferAmount = BigInt(
+    process.env.TRANSFER_AMOUNT ?? String(mintAmount),
+  );
+  const receiverSparkAddress = process.env.RECEIVER_SPARK_ADDRESS!;
+
+  // Initialize Turnkey client
+  const turnkeyClient = new TurnkeyServerSDK({
+    apiBaseUrl: process.env.BASE_URL ?? "https://api.turnkey.com",
```
janjakubnanista, line 73: I know this is just an example so no action required per se: `??` falls back on the second value only if the first one is not defined - it takes the first value for empty strings.

examples/with-spark-schnorr/src/index.ts
```diff
@@ -0,0 +1,186 @@
...
+    apiPrivateKey: process.env.API_PRIVATE_KEY!,
+    apiPublicKey: process.env.API_PUBLIC_KEY!,
+    defaultOrganizationId: process.env.ORGANIZATION_ID!,
+  });
+
+  // Create the Turnkey-backed signer
+  const signer = new TurnkeySparkSigner(
+    turnkeyClient,
+    process.env.TURNKEY_IDENTITY_ADDRESS!,
+    process.env.IDENTITY_PUBLIC_KEY_HEX!,
```
janjakubnanista: I'd smack these up above with the rest for prettiness

examples/with-spark-schnorr/src/turnkeySigner.ts
```diff
@@ -0,0 +1,212 @@
...
+  SubtractSplitAndEncryptResult,
+  VerifiableSecretShare,
+  SigningNonce,
+} from "@buildonspark/spark-sdk";
+import type { Transaction } from "@scure/btc-signer";
+import type { Turnkey as TurnkeyServerSDK } from "@turnkey/sdk-server";
+
+function hex(bytes: Uint8Array): string {
+  return Buffer.from(bytes).toString("hex");
+}
```
janjakubnanista, line 39: Shame [`toHex`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/toHex#browser_compatibility) is only supported on node 25

examples/with-spark-schnorr/src/turnkeySigner.ts
```diff
@@ -0,0 +1,212 @@
...
+    identityPublicKeyHex: string,
+  ) {
+    this.client = client;
+    this.identityKeyAddress = identityKeyAddress;
+    this.identityPublicKeyHex = identityPublicKeyHex;
+  }
+
+  async getIdentityPublicKey(): Promise<Uint8Array> {
+    return Buffer.from(this.identityPublicKeyHex, "hex");
+  }
```
janjakubnanista: No `async` required here btw

examples/with-spark-schnorr/src/index.ts
andrewkmin, line 53: do we need the cast `as const`?

examples/with-spark-schnorr/src/index.ts
andrewkmin: +1; can do without the `!` because the env var requirements check earlier enforce that these will be populated

examples/with-spark-schnorr/src/index.ts
```diff
@@ -0,0 +1,186 @@
...
+  // Final balance
+  const finalBalance = await wallet.getBalance();
+  console.log(`\n── Final balances ─────────────────────────────────────────`);
+  for (const [tokenId, info] of finalBalance.tokenBalances) {
+    console.log(
+      `   ${info.tokenMetadata?.tokenTicker ?? tokenId}: ${info.ownedBalance}`,
+    );
+  }
+
+  console.log(`\n✅ All Phase 1 operations succeeded with Turnkey signing!\n`);
```
andrewkmin: Do we define what Phase 1 entails? (given it's a bit of an internal-only concept)

andrewkmin: Down to just nix it if we want; don't feel too strongly

examples/with-spark-schnorr/src/turnkeySigner.ts
```diff
@@ -0,0 +1,212 @@
...
+ * That means the critical methods for authentication are:
+ *   - getIdentityPublicKey()
+ *   - signMessageWithIdentityKey()  ← ECDSA, DER-encoded output
+ *
+ * Token operations additionally need:
+ *   - signSchnorrWithIdentityKey()  ← Schnorr, 64-byte compact output
+ *
+ * Everything else (FROST, ECIES, HD key derivation) requires key material
+ * that lives inside Turnkey and cannot be directly expressed through Turnkey's
+ * current API surface. Those methods throw a clear "not implemented" error so
```
andrewkmin: perhaps we can add a note here that this is coming soon ™️

examples/with-spark-schnorr/src/turnkeySigner.ts
andrewkmin: oh wow it's finally gonna get baked in??? that's nice!

examples/with-spark-schnorr/src/turnkeySigner.ts
```diff
@@ -0,0 +1,212 @@
...
+  throw new Error(
+    `TurnkeySparkSigner.${method} is not implemented. ` +
+      `This method requires HD key derivation or FROST/ECIES operations ` +
+      `that are not yet expressible via Turnkey's signRawPayload API.`,
+  );
+}
+
+export class TurnkeySparkSigner implements SparkSigner {
+  private readonly client: TurnkeyServerSDK;
+  /** The Turnkey key ID / address used for the Spark identity key */
```
andrewkmin: given we call the variable `identityKeyAddress`, I'm down to just stick with `address` here

```suggestion
  /** The Turnkey address used for the Spark identity key */
```

examples/with-spark-schnorr/src/turnkeySigner.ts
```diff
@@ -0,0 +1,212 @@
...
+    _seed: Uint8Array | string,
+    _accountNumber?: number,
+  ): Promise<string> {
+    // No key derivation needed — Turnkey holds the keys. Return the identity
+    // public key hex as the default signer does, since callers use this return
+    // value to identify the wallet.
+    return this.identityPublicKeyHex;
+  }
+
+  // --- Not implemented: require key material Turnkey can't expose yet ---
```
andrewkmin:
```suggestion
  // --- Not implemented. Coming soon ---
```

examples/with-spark-schnorr/src/turnkeySigner.ts
```diff
@@ -0,0 +1,212 @@
...
+  async getDepositSigningKey(): Promise<Uint8Array> {
+    return notImplemented("getDepositSigningKey");
+  }
+  async getStaticDepositSigningKey(_idx: number): Promise<Uint8Array> {
+    return notImplemented("getStaticDepositSigningKey");
+  }
+  async getStaticDepositSecretKey(_idx: number): Promise<Uint8Array> {
+    return notImplemented("getStaticDepositSecretKey");
+  }
+  async generateMnemonic(): Promise<string> {
```
andrewkmin, line 150: this one's probably... a won't implement?

examples/with-spark-schnorr/.env.local.example
```diff
@@ -0,0 +1,40 @@
...
+# Optional: Turnkey credentials for signature comparison
+# Set these if you want to compare native Spark SDK Schnorr signatures
+# against Turnkey's signRawPayload output for the same key.
+# -------------------------------------------------------------------
+BASE_URL=
+ORGANIZATION_ID=
+API_PUBLIC_KEY=
+API_PRIVATE_KEY=
+
+# The Turnkey address / key identifier that holds the same private key
```
andrewkmin:
```suggestion
# The Turnkey address that holds the same private key
```

down to just implicitly recommend the HD wallet + address path

examples/with-spark-schnorr/README.md
```diff
@@ -0,0 +1,63 @@
+# Example: `with-spark-schnorr`
+
+Tests native Schnorr signing via the [Spark SDK](https://spark.money) (`DefaultSparkSigner`) and optionally compares the output against a Turnkey-backed signer — useful for verifying that a custom `SparkSigner` implementation produces identical signatures.
```
andrewkmin: > useful for verifying that a custom `SparkSigner` implementation produces identical signatures.

I believe this is no longer precisely the case given the Spark SDK uses a random nonce. But we should def mention how this example provides utilities to _verify_ that a message was signed by a Turnkey address

examples/with-spark-schnorr/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+```bash
+pnpm start
+```
+
+---
+
+## Signature comparison with Turnkey
+
+If you have a `SparkSigner` backed by Turnkey (signing via `signRawPayload`) and want to verify it produces the same Schnorr signatures as the native SDK, set the Turnkey env vars and `TURNKEY_IDENTITY_ADDRESS`. The script signs the same payload with both signers and prints whether the signatures match.
```
andrewkmin: > The script signs the same payload with both signers and prints whether the signatures match.

this may no longer be the case?

examples/with-spark-schnorr/.env.local.example
omkarshanbhag: It looks like I mixed up my branches!!! (I had a separate branch where I was doing a bunch of general tests on spark, and then one that was to be the example that gets pushed)

examples/with-spark-schnorr/src/index.ts
omkarshanbhag: Right -- makes sense

examples/with-spark-schnorr/README.md
omkarshanbhag: accidentally had the wrong readme up!! (have two branches)

examples/with-spark-schnorr/src/turnkeySigner.ts
omkarshanbhag: I think based on the interface defined by spark this returns a promise -- which is why I've kept it.

omkarshanbhag: It seems I can use `Promise.resolve` though

examples/with-spark-schnorr/src/index.ts
omkarshanbhag: I think the `!` are needed for the compiler cause even though they're validated they're technically treated as `string | undefined`

andrewkmin (approved):

examples/with-spark-schnorr/src/index.ts
janjakubnanista: `as const` prevents TypeScript from broadening the type to `string[]` - if you use `as const`, you will get a strongly-typed tuple

PR #1280, opened 2026-04-17
Landon/eng 4017

hadrelandon:
## Summary & Motivation

added OAuth CSRF state validation to web and react native. state is now stored before opening the browser and validated on callback. the state parameter was being generated and sent in the url but never stored or checked, meaning a forged callback could be accepted.

## How I Tested These Changes

locally

## Did you add a changeset?

yes


.changeset/big-camels-try.md
```diff
@@ -0,0 +1,6 @@
+---
+"@turnkey/react-native-wallet-kit": patch
+"@turnkey/react-wallet-kit": patch
+---
+
+Added `storeOAuthState`, `consumeOAuthState`, and `hasOAuthState` utilities and wired them into `buildOAuthUrl` and `parseInAppBrowserResult` to add OAuth CSRF state validation.
```
moeodeh3: we should only include public changes

these are all internal functions, so there’s no need to include them in the changeset

packages/react-native-wallet-kit/src/utils/oauth.ts
```diff
@@ -361,6 +361,42 @@ export async function hasPKCEVerifier(
...
+    throw new Error(`Missing stored OAuth state for validation.`);
+  }
+
+  if (!returnedState || stored !== returnedState) {
+    throw new Error(
+      `Invalid OAuth state returned from provider. Expected "${stored}", got "${returnedState}".`,
+    );
+  }
+
+  await AsyncStorage.removeItem("oauth_state");
```
moeodeh3: hmm I think the remove should be in a `finally` statement? Right now if the state doesn't match then it will never get removed from local storage!

packages/react-native-wallet-kit/src/utils/oauth.ts
```diff
@@ -361,6 +361,42 @@ export async function hasPKCEVerifier(
...
+}
+
+/**
+ * Checks if a stored OAuth state exists
+ * @returns true if OAuth state exists
+ */
+export async function hasOAuthState(): Promise<boolean> {
```
moeodeh3: hmmm I think we should know if we should validate the state or not? I don't think we need this?

packages/react-native-wallet-kit/src/utils/oauth.ts
```diff
@@ -486,6 +526,11 @@ export function parseInAppBrowserResult(
   const authCode = urlParams.get("code");
   const stateParam = urlParams.get("state");
 
+  // Validate state parameter
+  if (await hasOAuthState()) {
```
moeodeh3: I think we can remove `hasOAuthState` because I believe all of them should always have it

(atleast we return it for all of them)

packages/react-native-wallet-kit/src/utils/oauth.ts
```diff
@@ -486,6 +526,11 @@ export function parseInAppBrowserResult(
...
+    await consumeOAuthState(stateParam ?? "");
```
moeodeh3: we can have a `if (!stateParam)` and error, then we don't need this fallback anymore

packages/react-wallet-kit/src/utils/oauth/storage.ts
```diff
@@ -98,11 +99,54 @@ export function hasPKCEVerifier(provider: PKCEProvider): boolean {
   return localStorage.getItem(getPKCEVerifierKey(provider)) !== null;
 }
 
+/**
+ * Stores the OAuth state string in local storage for later validation
+ * @param state - The OAuth state string to store
+ */
+export function storeOAuthState(state: string) {
+  if (typeof localStorage === "undefined") return;
```
moeodeh3: what is this 🤖

packages/react-wallet-kit/src/utils/oauth/storage.ts
```diff
@@ -98,11 +99,54 @@ export function hasPKCEVerifier(provider: PKCEProvider): boolean {
...
+
+  localStorage.removeItem(OAUTH_STATE_KEY);
+}
+
+/**
+ * Checks if an OAuth state string exists in local storage
+ * @return true if an OAuth state string is stored, false otherwise
+ */
+export function hasOAuthState(): boolean {
+  if (typeof localStorage === "undefined") return false;
```
moeodeh3: same thing here 🤖 

we can remove this

packages/react-wallet-kit/src/utils/oauth/url.ts
```diff
@@ -179,6 +180,11 @@ export function parseOAuthResponse(
     }
   }
 
+  // Validate state if we have a stored state to compare against
+  if (stateString && hasOAuthState()) {
```
moeodeh3: like RN, we shouldn’t optionally enforce state

it should be enforced consistently

moeodeh3 (commented):
left some comments, also realized that in RN's `handleGoogleOauth` we totally ignore the `additionalState` param

lets fix that :)

.changeset/tricky-hoops-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+"@turnkey/sdk-types": minor
+---
+
+Added a new error code `INVALID_OAUTH_STATE` in `TurnkeyErrorCodes`
```
moeodeh3: lets also have a changeset that says we fixed state validation for oauth

moeodeh3 (commented):
looks good, let’s:

- add a changeset
- make a breaking change and remove `additionalState` from `@turnkey/react-native-wallet-kit`

the OAuth `state` parameter serves two purposes:

- CSRF protection (handled by our randomly generated state)
- maintaining context across redirects (this is what `additionalState` is for)

on web, a full redirect destroys the JS context, so encoding data into `state` helps recover it. But in React Native, `InAppBrowser.openAuth()` preserves the async context, so local variables, and refs all survive!  TLDR:  `additionalState` isn’t needed for RN

.changeset/witty-jeans-open.md
```diff
@@ -0,0 +1,7 @@
+---
+"@turnkey/react-native-wallet-kit": major
+---
+
+Removed `additionalState` from all OAuth handler params (`HandleGoogleOauthParams`, `HandleAppleOauthParams`, `HandleFacebookOauthParams`, `HandleDiscordOauthParams`, `HandleXOauthParams`) and from `buildOAuthState` / `buildOAuthUrl` utils.
+
+On web, full-page redirects destroy JS context so encoding data into the OAuth state parameter was necessary to recover it. In React Native, `InAppBrowser.openAuth()` preserves the async context — local variables and refs survive the redirect — making `additionalState` unnecessary.
```
moeodeh3: this looks familiar 🤔 , anyways we don't need this description

moeodeh3 (approved):

PR #1286, opened 2026-04-24
add stamper overrides

moeodeh3:
## Summary & Motivation

this pr is a continuation of [1158](https://github.com/tkhq/sdk/pull/1158)

the problem we’re trying to solve here is that we have several internal stampers that are configured at provider initialization time, but in some cases we need to be able to update them at runtime

This PR introduces:
- `overrideApiKeyStamper()`
- `overridePasskeyStamper()`
to address that issue

**Question: why didn’t we also include a walletStamper override?**
the reason is that this gets significantly more complex - especially with WalletConnect -and isn’t currently possible to do without re-initializing the provider, which is fairly expensive. I’m also concerned about potential race conditions here. For now, I’m intentionally excluding this from the pr until a customer explicitly asks for it

as a bonus, this pr also adds:
- `deleteApiKeyPair()`
this is useful for teams handling auth themselves and using ephemeral keys (e.g. one-tap passkeys)

## Did you add a changeset?


.changeset/many-foxes-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+"@turnkey/sdk-types": minor
+---
+
+Add `DELETE_API_KEY_PAIR_ERROR` to `TurnkeyErrorCodes`
```
moeodeh3, line 5: missing some error codes:

```
  INITIALIZE_API_KEY_STAMPER_ERROR = "INITIALIZE_API_KEY_STAMPER_ERROR",
  INITIALIZE_PASSKEY_STAMPER_ERROR = "INITIALIZE_PASSKEY_STAMPER_ERROR",
  INITIALIZE_WALLET_MANAGER_ERROR = "INITIALIZE_WALLET_MANAGER_ERROR",
```

packages/core/src/__clients__/core.ts
```diff
@@ -177,6 +178,14 @@ export type TurnkeyClientMethods = Omit<
   "init" | "config" | "httpClient" | "constructor"
 >;
 
+const AUTHENTICATOR_TRANSPORT_MAP: Record<string, AuthenticatorTransport> = {
```
moeodeh3: nit: we don't normally define constants in `core.ts` since this file is HUGE

we shouldn't make this file larger than it needs to be, think this would be better suited in `utils.ts`

packages/core/src/__clients__/core.ts
```diff
@@ -177,6 +178,14 @@ export type TurnkeyClientMethods = Omit<
...
+  AUTHENTICATOR_TRANSPORT_INTERNAL: "internal",
+  AUTHENTICATOR_TRANSPORT_HYBRID: "hybrid",
+  AUTHENTICATOR_TRANSPORT_USB: "usb",
+  AUTHENTICATOR_TRANSPORT_NFC: "nfc",
+  AUTHENTICATOR_TRANSPORT_BLE: "ble",
+};
+
 const ERC20_TRANSFER_ABI = [
```
moeodeh3: while you're at it, can you move this out too 🙏

packages/core/src/__clients__/core.ts
```diff
@@ -3254,7 +3263,28 @@ export class TurnkeyClient {
           );
         }
 
-        return userResponse.user;
+        const { user } = userResponse;
+
+        if (this.passkeyStamper) {
```
moeodeh3: hmm I don’t think this should live in `fetchUsers` because it’s called in a lot of places!

right now it runs:
- when we authenticate (which is what you want)
- when we fetch wallets
- when we add oauthProviders
- when getting or creating a DA user
- when the user intentionally calls it

so TLDR: too many places. We don’t want to keep overriding our config everywhere, this should only happen once during session creation 

maybe moving this into `handlePostAuth()` in both react and react-native (right after we fetch the user) is better?

packages/core/src/__clients__/core.ts
```diff
@@ -3254,7 +3263,28 @@ export class TurnkeyClient {
...
+          const allowCredentials: PublicKeyCredentialDescriptor[] =
+            user.authenticators.map((a) => {
+              const b64 = base64UrlToBase64(a.credentialId);
+              const binary = atob(b64);
+              const id = Uint8Array.from(binary, (c) => c.charCodeAt(0));
+              return {
+                id,
+                type: "public-key" as const,
```
moeodeh3: nit: don't need `as const`

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
```diff
@@ -662,7 +667,26 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
       setAllSessions(allSessions);
 
       await maybeRefreshWallets();
-      await maybeRefreshUser();
+      const user = await maybeRefreshUser();
+
+      if (user && client?.config?.passkeyConfig) {
```
moeodeh3: hmm, shouldn't it be:

```
if (user && this.passkeyStamper) {
```

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
```diff
@@ -662,7 +667,26 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
       setAllSessions(allSessions);
 
       await maybeRefreshWallets();
-      await maybeRefreshUser();
+      const user = await maybeRefreshUser();
```
moeodeh3, line 685: great, so when people log in we override here!

but what about users re-initializing the provider who already have sessions 🤔 ? `handlePostAuth` won’t run in that case. I think we also need to apply this logic in `initializeSessions`!

packages/react-wallet-kit/src/providers/client/Provider.tsx
```diff
@@ -1283,7 +1288,29 @@ export const ClientProvider: React.FC<ClientProviderProps> = ({
       setSession(session);
       setAllSessions(allSessions);
 
-      await Promise.all([maybeRefreshWallets(), maybeRefreshUser()]);
+      const [, user] = await Promise.all([
+        maybeRefreshWallets(),
+        maybeRefreshUser(),
+      ]);
+
+      if (user && client?.config?.passkeyConfig) {
```
moeodeh3: same thing here: `this.passkeyStamper`?

packages/react-wallet-kit/src/providers/client/Provider.tsx
```diff
@@ -1283,7 +1288,29 @@ export const ClientProvider: React.FC<ClientProviderProps> = ({
       setSession(session);
       setAllSessions(allSessions);
 
-      await Promise.all([maybeRefreshWallets(), maybeRefreshUser()]);
+      const [, user] = await Promise.all([
```
moeodeh3, line 1301: we also need this inside `initializeSessions`

moeodeh3 (commented):
small comments but looks good

nit: can we also remove `storeOverride` from `CreateApiKeyPairParams` since we no longer need it after these changes! (and can we create a changeset for this thats marked as major 😄 )

.changeset/thick-emus-stay.md
```diff
@@ -0,0 +1,6 @@
+---
+"@turnkey/react-native-wallet-kit": patch
+"@turnkey/react-wallet-kit": patch
+---
+
+scope passkey allowCredentials to the logged in users authentication after login. refreshUser and maybeRefreshUser now return v1User | undefined
```
moeodeh3: think we can make this clearer

also, `maybeRefreshUser` isn’t exported to sdk users, so there’s no need to mention it in the changeset since it’s purely internal!

packages/core/src/__clients__/core.ts
```diff
@@ -174,25 +178,12 @@ export type TurnkeyClientMethods = Omit<
...
-  },
-] as const;
-
 export class TurnkeyClient {
   config: TurnkeySDKClientConfig;
   httpClient!: TurnkeySDKClientBase;
 
   private apiKeyStamper?: CrossPlatformApiKeyStamper | undefined;
-  private passkeyStamper?: CrossPlatformPasskeyStamper | undefined;
+  passkeyStamper?: CrossPlatformPasskeyStamper | undefined;
```
moeodeh3: hmm, not sure we should make the `passkeyStamper` public…

and if we do, then we should probably make all the stampers public for consistency

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
```diff
@@ -604,8 +609,30 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
           return;
         }
         setSession(allLocalStorageSessions[activeSessionKey]);
-        await maybeRefreshUser();
+        const user = await maybeRefreshUser();
         await maybeRefreshWallets();
+
+        if (user && client?.passkeyStamper) {
```
moeodeh3: okay I see why you made it public now

okay so right now our current state is that:
- `@turnkey/core` only initializes the `passkeyStamper` if a config is passed to it
- `@turnkey/react-native-wallet-kit` and `@turnkey/react-wallet-kit` always pass in configs (because of defaults) even is sdk users didn't provide a config

so actually what we should do here (which expands the scope a bit):
- we always initialize the` passkeyStamper` in  `@turnkey/react-wallet-kit`. Then we can always call `overridePasskeyStamper`.
- we only initialize the `passkeyStamper` in `@turnkey/react-native-wallet-kit`. We only call `overridePasskeyStamper` if a config exists

then we shouldnt need to make the `passkeyStamper` public

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
```diff
@@ -604,8 +609,30 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
...
+          const allowCredentials: PublicKeyCredentialDescriptor[] =
+            user.authenticators.map((a) => {
```
moeodeh3: hmm maybe worth defining a function in `@turnkey/core`, something like:

```
function buildAllowCredentialsFromAuthenticators(
  authenticators: v1Authenticator[],
): PublicKeyCredentialDescriptor[] {
  return authenticators.map((a) => {
    const b64 = base64UrlToBase64(a.credentialId);
    const binary = atob(b64);
    const id = Uint8Array.from(binary, (c: string) => c.charCodeAt(0));
    return {
      id,
      type: "public-key",
      transports: a.transports
        .map((t) => AUTHENTICATOR_TRANSPORT_MAP[t])
        .filter(Boolean) as AuthenticatorTransport[],
    };
  });
}
```

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
```diff
@@ -376,18 +380,30 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
...
 
-        // Define passkey and wallet config here. If we don't pass it into the client, Mr. Client will assume that we don't want to use passkeys/wallets and not create the stamper!
-        passkeyConfig: {
-          rpId: masterConfig.passkeyConfig?.rpId,
-          timeout: masterConfig.passkeyConfig?.timeout || 60000, // 60 seconds
-          userVerification:
-            masterConfig.passkeyConfig?.userVerification || "preferred",
-          allowCredentials: masterConfig.passkeyConfig?.allowCredentials || [],
-        },
+        ...(masterConfig.passkeyConfig
```
moeodeh3, line 383: 👑

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
```diff
@@ -376,18 +380,30 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
...
+                  masterConfig.passkeyConfig.allowCredentials ?? [],
+              },
+            }
+          : {}),
+
       });
 
       await turnkeyClient.init();
       setClient(turnkeyClient);
+      console.log("passkeyConfig:", turnkeyClient.config.passkeyConfig);
```
moeodeh3: lets remove this console log

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
```diff
@@ -376,18 +380,30 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
...
+      try {
+        await turnkeyClient.overridePasskeyStamper({ config: { rpId: "test" } });
+        console.log("[Turnkey] passkey stamper IS initialized");
```
moeodeh3: :(

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
```diff
@@ -376,18 +380,30 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
...
+      } catch (e: any) {
+        console.log("[Turnkey] passkey stamper NOT initialized:", e.message);
```
moeodeh3: 😠

packages/react-native-wallet-kit/src/providers/Types.tsx
```diff
@@ -176,7 +176,7 @@ export interface ClientContextType
    */
   refreshUser: (params?: {
```
moeodeh3, line 177: probably needs the comment header here updated? we use this to generate our docs

packages/react-wallet-kit/package.json
```diff
@@ -50,6 +50,7 @@
     "@lottiefiles/react-lottie-player": "^3.6.0",
     "@noble/hashes": "^1.8.0",
     "@turnkey/core": "workspace:*",
+    "@turnkey/encoding": "workspace:*",
```
moeodeh3: think we can remove this now, no longer needed after our changes

moeodeh3 (commented):
LGTM but I can't approve since this is my own pr 😢

ethankonk (approved):

PR #1290, opened 2026-04-29
fixed logout, and added scopePasskeyToUser

hadrelandon:
## Summary & Motivation

Builds off of #1286 to add auth.scopePasskeyToUser config options and fix passkey credential scoping across sessions. Before allowCredentials was unconditionally set to the current users authenticators on auth, no way to opt out and it wasn't applied consistently across all session lifecycle events. This PR introduces the provider config change so when scopePasskeyToUser is set to false, allow credentials stays empty to show all credentials, when true or unset we scope to the active users authenticator. 

## How I Tested These Changes

local

## Did you add a changeset?

yes


packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
```diff
@@ -732,33 +732,38 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
...
-    }
-  };
+  const handlePostLogout = useCallback(
+    async (sessionKey?: string) => {
+      try {
+        clearSessionTimeouts(
+          sessionKey ? [sessionKey, `${sessionKey}-warning`] : undefined,
+        );
+        setAllSessions((prev) => {
+          if (!prev) return prev;
```
figitaki, line 743: What is this checking? Are we trying to pass through undefined values?

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
```diff
@@ -732,33 +732,38 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
...
+        });
+        setSession(undefined);
+        setUser(undefined);
+        setWallets([]);
+        if (client)
+          await resetPasskeyScope(client, client.config.passkeyConfig);
+      } catch (error) {
+        callbacks?.onError?.(
+          new TurnkeyError(
+            `Failed to initialize sessions`,
```
ethankonk: This looks like an existing error but might as well fix it now, error messaging is wrong here, should prob be something like "Failed to clear session" or something logout related

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
```diff
@@ -2199,8 +2217,17 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
       setSession(session);
       setAllSessions(allSessions);
 
-      await maybeRefreshWallets();
-      await maybeRefreshUser();
+      const [, nextUser] = await Promise.all([
```
figitaki: If you reverse the order of the promise all array this could be 
```suggestion
      const [nextUser] = await Promise.all([
```

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
```diff
@@ -2243,13 +2274,16 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
       );
     setSession(undefined);
     setAllSessions(undefined);
+    setUser(undefined);
+    setWallets([]);
     clearSessionTimeouts();
-    return await withTurnkeyErrorHandling(
```
figitaki, line 2247: Was dropping the return statement intentional?

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
```diff
@@ -2361,8 +2395,17 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
       setSession(session);
       await withTurnkeyErrorHandling(
         async () => {
-          await maybeRefreshWallets();
-          await maybeRefreshUser();
+          const [, nextUser] = await Promise.all([
```
figitaki: Same here

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
```diff
@@ -2199,8 +2217,17 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
       setSession(session);
       setAllSessions(allSessions);
 
-      await maybeRefreshWallets();
-      await maybeRefreshUser();
+      const [, nextUser] = await Promise.all([
+        maybeRefreshWallets(),
+        maybeRefreshUser(),
+      ]);
```
ethankonk: We should be careful about sending synchronous requests when using passkey stamping. Usually truly synchronous requests actually fail here if using `StamperType.Passkey`, should test to make sure this still works

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
```diff
@@ -2361,8 +2395,17 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
       setSession(session);
       await withTurnkeyErrorHandling(
         async () => {
-          await maybeRefreshWallets();
-          await maybeRefreshUser();
+          const [, nextUser] = await Promise.all([
+            maybeRefreshWallets(),
+            maybeRefreshUser(),
+          ]);
```
ethankonk: Same warning here for synchronous requests + passkeys. Double check this works still

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
hadrelandon: checking if we have a sessionKey, if we do we clear that and its warning. if we don't we pass nothing and clear all timeouts. i can make it clearer with a variable or something.

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
hadrelandon: yes, i added additional cleanup after withTurnkeyErrorHandling so dont want to return early. Specifically here resetPasskeyScope

ethankonk (approved):

PR #1298, opened 2026-05-08
Add with-vaultsfyi example

pawelpolak2:
## Summary & Motivation
Adds a with-vaultsfyi example showing how to use Turnkey wallets with the vaults.fyi SDK to discover, deposit into, and withdraw from DeFi yield vaults, including scoped policies for non-root users. Companion to the vaults.fyi cookbook guide.

## How I Tested These Changes
Manually tested all of the scripts.

## Did you add a changeset?
No, it's example code.


MasterHW:
merged in #1436

PR #1304, opened 2026-05-14
Am/update/with x

am-hernandez:
## Summary & Motivation

Migrates the `with-x` example from `@turnkey/sdk-react` to `@turnkey/react-wallet-kit`. Session storage is now handled by the wallet kit's `storeSession`, while custom server-side routes remain in place to perform the OAuth 2.0 exchange with X and issue a Turnkey session token.

Key changes:

- Remove `@turnkey/sdk-react`, `@turnkey/encoding`, `axios`, and add `@turnkey/react-wallet-kit` for client-side auth and session management
- Retain custom server routes (`/auth/x`, `/auth/x/redirect`, `/auth/turnkey/x`) and the `credential-upload` script for the X OAuth 2.0 flow
- Add `nonce` (derived as `sha256(publicKey)`) to the `oauth2Authenticate` call, now required by the Turnkey API
- Extract reusable components (`LoginWithXButton`, `LogoutButton`, `Providers`, `CopyButton`, `SolanaIcon`, `Loading`) for clarity
- Rewrite the README to reflect the new setup flow: creating a Turnkey API keypair, configuring an X developer app, and uploading the X client secret via `credential-upload`

Closes [CUS-183](https://linear.app/turnkey/issue/CUS-183/update-the-with-x-example)

## How I Tested These Changes

Ran the example end-to-end locally:

1. Created a Turnkey API keypair, set up an X developer app, and uploaded the X client secret via `pnpm run credential-upload`
2. Started the app with `pnpm run dev`
3. Completed the X OAuth 2.0 login flow and verified the dashboard displayed the correct user ID and Solana wallet address
4. Verified logout clears session state correctly and the login page is shown again

## Did you add a changeset?

No changeset added. No package updates included in these changes.


t-vila:
@am-hernandez apologies, I should have been clearer here - can we leave it use a custom backend / existing routes and not via Auth Proxy ?
Reason is, we have the auth proxy X login in the ewk demo https://github.com/tkhq/sdk/tree/main/examples/react-wallet-kit and it would be good for the users to also see the custom backend flavor

examples/with-x/README.md
```diff
@@ -1,11 +1,11 @@
...
-This example shows a complete the Oauth 2.0 flow with X. It contains a NextJS app with:
+This example shows a complete OAuth 2.0 login flow with X (Twitter) using [`@turnkey/react-wallet-kit`](https://www.npmjs.com/package/@turnkey/react-wallet-kit). It contains a Next.js app with:
 
-- a frontend application
-- a backend application
+- A login page that initiates the X OAuth 2.0 flow
+- A dashboard page that displays the authenticated user's ID and Solana wallet address
 
-This example contains an example auth page as well as a stub API endpoint for "your business" (where the user is resolved into an organization ID). It relies on the `indexedDbClient` to stamp the requests. For more information on oauth, [check out our documentation](https://docs.turnkey.com/authentication/social-logins).
+Authentication and session management are handled entirely by `@turnkey/react-wallet-kit` via the Turnkey auth proxy — no custom server routes required. For more information on OAuth, [check out our documentation](https://docs.turnkey.com/authentication/social-logins).
```
t-vila: I think we do use server routes right ?

examples/with-x/README.md
```diff
@@ -24,73 +24,56 @@ $ cd examples/with-x/
 
 ### 2/ Setting up Turnkey
 
-The first step is to set up your Turnkey organization and account. By following the [Quickstart](https://docs.turnkey.com/getting-started/quickstart) guide, you should have:
+If you don't have a Turnkey account yet, follow the [Quickstart](https://docs.turnkey.com/getting-started/quickstart) guide to create an organization.
 
-- A public/private API key pair for Turnkey
-- An organization ID
+You'll also need an **auth proxy config** — create one in the Turnkey dashboard under [**Auth Proxy**](https://docs.turnkey.com/reference/auth-proxy#auth-proxy). The auth proxy handles the OAuth token exchange server-side so no API keys are needed in the app.
```
t-vila: I'm guessing that the whole readme needs an update, it still has the auth proxy leftovers

examples/with-x/README.md
am-hernandez: ah yes, I needed to update the readme after the fact and missed this. Thank you!

examples/with-x/README.md
am-hernandez: yes, same as above!

examples/with-x/src/app/auth/x/route.tsx
```diff
@@ -1,17 +1,41 @@
...
+    code_challenge_method: "S256",
   });
 
-  return NextResponse.redirect(
+  const response = NextResponse.redirect(
     `https://x.com/i/oauth2/authorize?${params.toString()}`,
   );
+  const cookieOptions = {
+    httpOnly: true,
+    sameSite: "lax" as const,
```
t-vila, line 34: not, but maybe we should mention `secure: true` for prod so people won't miss that out ?
`secure: process.env.NODE_ENV === "production", // use secure: true in production`

examples/with-x/src/app/auth/x/redirect/page.tsx
```diff
@@ -1,68 +1,54 @@
...
-          auth_code: auth_code,
-          state: state,
-          public_key: pubKey,
+        const publicKey = await createApiKeyPair();
+
+        const res = await fetch("/auth/turnkey/x", {
+          method: "POST",
+          headers: { "Content-Type": "application/json" },
+          body: JSON.stringify({ auth_code, state, public_key: publicKey }),
         });
```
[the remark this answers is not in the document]
t-vila: The server already validates state in auth/x/route.tsx by comparing body.state against the pkce_state httpOnly cookie. So this client-side check is redundant — an attacker can't spoof the cookie, and a legitimate user will always have the right state in the URL since X echoes it back.

It doesn't hurt functionally, but it introduces a race: TurnkeyProvider and the redirect page's useEffect both react to the same ?code=&state= params on mount. The Provider fires first, looks for oauth_state in localStorage (its own key, not x_oauth_state), throws, then sets ClientState.Ready via .finally(), after which the useEffect runs and succeeds. Hence this console error:

Uncaught (in promise) TurnkeyError: Missing OAuth state in storage

examples/with-x/src/app/auth/turnkey/x/route.ts
```diff
@@ -73,15 +78,12 @@ export async function POST(req: Request) {
...
+        codeVerifier,
         bearerTokenTargetPublicKey: keypair.publicKeyUncompressed, // NOTE: This only needs to be provided if you would like the encrypted bearer token to be returned via the `enctypedBearerToken` claim of the OIDC ID Token
         nonce: bytesToHex(sha256(sessionPublicKey)),
       });
-
-    // get the encryptedBearerToken
-    getEncryptedBearerTokenFromOidcToken(oauth2AuthenticateResponse.oidcToken);
-
     // you can now decrypt and store the bearer token as shown below (code commented out for security reasons)
+    // const encryptedBearerToken = getEncryptedBearerTokenFromOidcToken(oauth2AuthenticateResponse.oidcToken);
```
t-vila, line 87: we should probably add the commented import too 
// import { decryptCredentialBundle } from "@turnkey/crypto";

t-vila (approved):

PR #1306, opened 2026-05-15
Remove deprecated @tsconfig/node16-strictest and fix viem tsconfig.json

janjakubnanista:
## Summary & Motivation

Screenshot 2026-05-15 at 10 31 17 AM

The deprecated package no longer works with TypeScript 5, and is no longer visible in the NPM UI. On top of that, we are no longer on Node 16.

After the update, an `override` keyword needed to be added to `TurnkeyError` constructor.

The `viem` package `tsconfig.json` was also made consistent with the rest of the packages (it didn't reference the base config),  and `allowJs` needed to be added to `jest-config/tsconfig.json`.

## How I Tested These Changes

Build + typecheck

## Did you add a changeset?

Does not require one


packages/api-key-stamper/src/purejs.ts
ethankonk, line 1: I think this PR is still WIP but just heads up that we should add a changeset for package changes. I don't think this is an API level change but worth adding a little blurb in a changeset on the switch!

packages/http/src/shared.ts
```diff
@@ -13,7 +13,6 @@ export class TurnkeyActivityError extends Error {
   activityId: TActivityId | undefined;
   activityStatus: TActivityStatus | undefined;
   activityType: TActivityType | undefined;
-  cause: Error | undefined;
```
ethankonk, line 16: How come the error causes are being removed? Is it just unused?

packages/react-native-passkey-stamper/src/util.ts
```diff
@@ -1,12 +1,11 @@
 // Buffer is available in Node and react-native contexts but needs to be imported
 import { Buffer } from "buffer";
-import { createHash } from "sha256-uint8array";
```
ethankonk, line 3: Same here, worth a changeset note

PR #1308, opened 2026-05-16
returning credentialId in loginwithpasskey

hadrelandon:
## Summary & Motivation
loginWithPasskey was returning an empty string for credentialId. This extracts the actual credential ID from the passkey stamp header and returns it.

## How I Tested These Changes
locally

## Did you add a changeset?

yes

### DIFF
--- a/.changeset/legal-dodos-lose.md
+++ b/.changeset/legal-dodos-lose.md
@@ -1,0 +1,5 @@
+---
+"@turnkey/core": patch
+---
+
+return actual credentialId from loginWithPasskey instead of empty string
--- a/.changeset/legal-dodos-lose.md
+++ b/.changeset/legal-dodos-lose.md
@@ -2,4 +2,4 @@
 "@turnkey/core": patch
 ---
 
-return actual credentialId from loginWithPasskey instead of empty string
+return actual `credentialId` from `loginWithPasskey()` instead of empty string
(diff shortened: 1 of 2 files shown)


.changeset/legal-dodos-lose.md
```diff
@@ -0,0 +1,5 @@
+---
+"@turnkey/core": patch
+---
+
+return actual credentialId from loginWithPasskey instead of empty string
```
moeodeh3: nit: use ` around `credentialId` and `loginWithPasskey()`

packages/core/src/__clients__/core.ts
```diff
@@ -554,28 +554,44 @@ export class TurnkeyClient {
...
         );
 
+        if (!passkeyLoginRequest) {
+          throw new TurnkeyError(
+            "Failed to generate login payload for passkey login.",
+            TurnkeyErrorCodes.INTERNAL_ERROR,
+          );
+        }
+
+        const credentialId = (
```
moeodeh3, line 577: another nit: probably worth adding a comment explaining what we’re doing here. Basically, when we stamp with a passkey, the `credentialId` lives inside the `stampHeaderValue` json, and we’re just extracting it here

moeodeh3 (dismissed):
LGTM

moeodeh3 (approved):

PR #1312, opened 2026-05-18
Chore: Clean up typecheck

janjakubnanista:
## Summary & Motivation

In preparation for [updating our `tsconfig`s to non-deprecated version](https://github.com/tkhq/sdk/pull/1306) (#1306), I added a small cleanup:

- `tsconfig.typecheck.json` files have been removed in favor of running `tsc --noEmit`. In many cases these tsconfigs were out of sync with the main ones (and prsented an opportunity to make them out of sync)

## How I Tested These Changes

Locally + CI

## Did you add a changeset?

Dev changes, changeset not needed


ethankonk (commented):
Two small comments but otherwise looks good:

1. Deleting the `tsconfig.typecheck.json` for `core`, `crypto`, `http`, and `wallet-stamper` removes type checking from test files. Not entirely sure if this was the case previously since I'm not sure if the typecheck config file took priority over the base tsconfig but tested this case manually and did see it pass typecheck. Isn't a huge issue tho but worth double checking if that's what we want..

2. Could use some clarification on what `^.build` does. Was curious about this change and read through the turborepo Transit Nodes docs, since none of the packages have a .build script, my read is that it propagates input hashes for cache invalidation but doesn't actually execute anything upstream. Just want to make sure that's the intent.

janjakubnanista:
@ethankonk lemme drop the `.build` - it was a way around the fact that `build` depends on `clean` which pretty much invalidates turbo cache. I'll restrict this PR to the typecheck and will add the missing test files - I think the problem was they were missing from `tsconfig.json` in the first place

ethankonk (dismissed):

ethankonk (approved):

PR #1313, opened 2026-05-18
Chore: Cache turbo in CI

janjakubnanista:
## Summary & Motivation

We were not caching turbo builds in our CI

## How I Tested These Changes

CI

## Did you add a changeset?

Not needed

### DIFF
--- a/.github/actions/js-setup/action.yml
+++ b/.github/actions/js-setup/action.yml
@@ -49,6 +49,24 @@ runs:
       run: pnpm install -r --frozen-lockfile
       shell: bash
 
+    # We build architecture specific caches for turbo builds as there can be differences in the build artifacts between architectures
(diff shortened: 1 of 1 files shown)


ethankonk (commented):
This would be an awesome change but I don't think we're using turbo to build our packages at the moment.

We seem to use `pnpm run build-all` in CI which runs the build cmd for every individual package rather than our repo's turbo wrapper `pnpm run build`.

You can see this in the root `package.json`

```json
"build-all": "pnpm run --r --filter \"./packages/**\" build",
```

vs

```json
"build": "turbo --filter \"./packages/**\" build",
```

I'm sure we can just switch over to turbo to make CI runs much quicker but unsure if there are any downsides to doing so

PR #1314, opened 2026-05-18
Add SDK helper for Turnkey webhook signature verification

taylorjdawson:
## Summary
- Add a pure @turnkey/crypto helper for verifying Turnkey webhook signatures
- Support exact raw body verification against caller-provided Ed25519 verification keys
- Cover valid signatures, tampered payloads/headers/signatures, missing keys, stale timestamps, and unsupported protocol values
- Add crypto package docs for raw-body usage

## Notes
- This intentionally does not add a server-side key discovery/fetching helper yet; discovery endpoint shape is still undecided and should be handled in a follow-up.

## Tests
- pnpm --filter @turnkey/crypto test -- webhooks-test.ts
- pnpm --filter @turnkey/crypto typecheck
- pnpm --filter @turnkey/crypto test

### DIFF
--- a/packages/crypto/README.md
+++ b/packages/crypto/README.md
@@ -33,3 +33,39 @@ const decryptedData = hpkeDecrypt({
 // Convert decrypted data back to string
 const decryptedText = new TextDecoder().decode(decryptedData);
 ```
+
+## Verifying Turnkey webhooks
+
+Use `@turnkey/crypto` directly when you want to manage verification-key fetching and caching yourself. Verification must use the exact raw request body bytes that Turnkey sent, the Turnkey signature headers, Turnkey webhook verification keys, and an explicit `maxTimestampAgeMs` replay window.
+
+Turnkey sends these signature headers with the body: `x-turnkey-timestamp`, `x-turnkey-event-id`, `x-turnkey-signature-key-id`, `x-turnkey-signature-algorithm`, `x-turnkey-signature-version`, and `x-turnkey-signature`. Pass the complete headers object through as received.
+
+`x-turnkey-event-id` is stable across retry attempts for the same webhook event. Use it as the deduplication or idempotency key after signature verification succeeds.
+
+```ts
+import { verifyTurnkeyWebhookSignature } from "@turnkey/crypto";
+
+const body = req.body; // Buffer from express.raw(), not parsed JSON
+const verificationKeys = [
+  {
+    keyId: process.env.TURNKEY_WEBHOOK_KEY_ID!,
+    publicKey: process.env.TURNKEY_WEBHOOK_PUBLIC_KEY!, // Hex-encoded Ed25519 public key
+    algorithm: "ed25519",
+  },
+];
+
+const result = verifyTurnkeyWebhookSignature({
+  headers: req.headers,
+  body,
+  verificationKeys,
+  maxTimestampAgeMs: 5 * 60 * 1000,
+});
+
+if (!result.ok) {
+  throw new Error(`Invalid Turnkey webhook: ${result.reason}`);
+}
+
+const event = JSON.parse(body.toString("utf8"));
+```
+
+Do not verify a parsed and re-stringified JSON object. Even harmless-looking changes to whitespace or key ordering will change the signed payload.
--- a/packages/crypto/src/__tests__/webhooks-test.ts
+++ b/packages/crypto/src/__tests__/webhooks-test.ts
@@ -1,0 +1,325 @@
+import { test, expect, describe } from "@jest/globals";
+import { ed25519 } from "@noble/curves/ed25519";
+import { uint8ArrayToHexString } from "@turnkey/encoding";
+import {
+  TurnkeyWebhookVerificationFailureReasons,
+  verifyTurnkeyWebhookSignature,
+  type TurnkeyWebhookVerificationKey,
+} from "../";
+import { buildSignedInput } from "../webhooks";
+
+const PRIVATE_KEY = new Uint8Array([
+  0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0a, 0x0b, 0x0c,
+  0x0d, 0x0e, 0x0f, 0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16, 0x17, 0x18, 0x19,
+  0x1a, 0x1b, 0x1c, 0x1d, 0x1e, 0x1f,
+]);
+
+const KEY_ID = "webhook-key-1";
+const EVENT_ID = "evt_123";
+const TIMESTAMP_MS = 1_700_000_000_000;
+const NOW_MS = TIMESTAMP_MS + 1_000;
+const MAX_TIMESTAMP_AGE_MS = 5 * 60 * 1_000;
+const BODY = '{"type":"activity.created","data":{"id":"act_123"}}';
+
+describe("verifyTurnkeyWebhookSignature", () => {
+  test("verifies a valid signature with plain record headers", () => {
+    const fixture = createFixture({ body: BODY });
+
+    const result = verifyTurnkeyWebhookSignature({
+      headers: {
+        "x-turnkey-timestamp": String(TIMESTAMP_MS),
+        "X-Turnkey-Event-Id": EVENT_ID,
+        "x-turnkey-signature-key-id": KEY_ID,
+        "X-Turnkey-Signature-Algorithm": "ed25519",
+        "x-turnkey-signature-version": "v1",
+        "X-Turnkey-Signature": fixture.signature,
+      },
+      body: BODY,
+      verificationKeys: fixture.verificationKeys,
+      maxTimestampAgeMs: MAX_TIMESTAMP_AGE_MS,
+      nowMs: NOW_MS,
+    });
+
+    expect(result).toEqual({
+      ok: true,
+      eventId: EVENT_ID,
+      keyId: KEY_ID,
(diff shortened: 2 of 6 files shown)


packages/crypto/src/webhooks.ts
```diff
@@ -0,0 +1,266 @@
+import { ed25519 } from "@noble/curves/ed25519";
+import { uint8ArrayFromHexString } from "@turnkey/encoding";
+
+export type TurnkeyWebhookVerificationKey = {
+  keyId: string;
+  publicKey: string;
+  algorithm?: "ed25519";
+};
+
+export type TurnkeyWebhookHeaders =
```
moeodeh3, line 10: a lot of these types would probably fit better in `@turnkey/sdk-types` rather than here

anything specific to this package (e.g. `VerifyTurnkeyWebhookSignatureParams`) can stay here, but anything Turnkey-wide and likely to be reused across packages should live in `@turnkey/sdk-types` so we avoid duplicating types across every package!

packages/crypto/src/webhooks.ts
```diff
@@ -0,0 +1,266 @@
...
+
+function hexToBytes(
+  input: string,
+  expectedLength: number,
+  reason: string,
+): { ok: true; value: Uint8Array } | { ok: false; reason: string } {
+  try {
+    return {
+      ok: true,
+      value: uint8ArrayFromHexString(input, expectedLength),
```
ethankonk: I believe this helper 0 pads inputs that are shorter than the `expectedLength` meaning if say a 2 character  "signature" was sent through on accident, it would zip right through length validation and into the signature verification which would throw a cryptographic error rather than a message like "this signature is too short" or something along those lines which could be confusing.

Easy fix would be to do an explicit length verification before calling the helper

packages/crypto/src/webhooks.ts
```diff
@@ -0,0 +1,266 @@
...
+  | {
+      ok: true;
+      eventId: string;
+      keyId: string;
+      timestampMs: number;
+    }
+  | {
+      ok: false;
+      reason: string;
+    };
```
ethankonk: I feel like this would look nicer as two distinct `Success` and `Failure` types rather than inlined types like they are right now. Maybe something like:

```ts
export type TurnkeyWebhookVerifiactionSuccess = {
  ok: true;
  eventId: string;
  keyId: string;
  timestampMs: number;
}

export type TurnkeyWebhookVerifiactionFailure = {
  ok: false;
  reason: string;
}

export type TurnkeyWebhookVerifiactionResult =  TurnkeyWebhookVerifiactionSuccess | TurnkeyWebhookVerifiactionFailure
```

packages/crypto/src/webhooks.ts
```diff
@@ -0,0 +1,266 @@
...
+const HEADER_ALGORITHM = "x-turnkey-signature-algorithm";
+const HEADER_VERSION = "x-turnkey-signature-version";
+const HEADER_SIGNATURE = "x-turnkey-signature";
+
+const SUPPORTED_VERSION = "v1";
+const SUPPORTED_ALGORITHM = "ed25519";
+const PUBLIC_KEY_BYTE_LENGTH = 32;
+const ED25519_SIGNATURE_BYTE_LENGTH = 64;
+
+export function verifyTurnkeyWebhookSignature({
```
ethankonk, line 79: Methods surfaced to SDK users should include some JS doc comments for easy to access documentation goodness!

Simply adding some light descriptions for the input params, throw cases, and maybe return values would be perfect

packages/crypto/README.md
ethankonk, line 1: Played around with webhooks before these changes and found the dedupe key was super helpful for retry dedupage. Would be nice to add a bit of documentation on using `X-Turnkey-Event-Id` as the stable handle and making it clear that it exists and is a great way to dedupe retries!

packages/crypto/src/__tests__/webhooks-test.ts
```diff
@@ -0,0 +1,279 @@
...
+
+function bodyToBytes(body: string | Uint8Array | ArrayBuffer): Uint8Array {
+  if (typeof body === "string") {
+    return new TextEncoder().encode(body);
+  }
+  if (body instanceof Uint8Array) {
+    return body;
+  }
+  return new Uint8Array(body);
+}
```
ethankonk: I think we should avoid re-implementing functions for test cases to avoid drift in production code vs what we test on. Don't think it would be too bad to specify methods we want to export in the base `index.ts` for webhook related stuff to get closer test coverage (using the same helpers being shipped to production code in our tests).

ethankonk (commented):
Looking good but not sure this is ready to be released just yet, there is still the issue of _how_ SDK devs are supposed to get the public key to verify against (unless thats already solved!)

packages/crypto/src/webhooks.ts
```diff
@@ -0,0 +1,266 @@
...
+
+export type TurnkeyWebhookVerificationKey = {
+  keyId: string;
+  publicKey: string;
+  algorithm?: "ed25519";
+};
+
+export type TurnkeyWebhookHeaders =
+  | Headers
+  | Record<string, string | string[] | undefined>;
```
moeodeh3, line 12: why do we have this? it just adds complexity tbh. If we accept a record string then we have to worry about case insensitivity too I think?

I feel like we should just do

```
export type VerifyTurnkeyWebhookSignatureParams = {
  headers: Headers;
  body: TurnkeyWebhookBody;
  verificationKeys: TurnkeyWebhookVerificationKey[];
  maxTimestampAgeMs: number;
  nowMs?: number;
};
```

then we can just do
```
function getHeader(
  headers: Headers,
  name: string,
): { ok: true; value: string } | { ok: false; reason: string } {
  const value = headers.get(name);
  if (!value) {
    return invalid(`missing_header:${name}`);
  }
  return { ok: true, value };
}
```

packages/crypto/src/webhooks.ts
moeodeh3: actually idk if we need to move any of these now that I think of it

packages/crypto/src/webhooks.ts
```diff
@@ -0,0 +1,266 @@
...
+  }
+
+  return hexToBytes(
+    key.publicKey,
+    PUBLIC_KEY_BYTE_LENGTH,
+    "invalid_verification_key",
+  );
+}
+
+function hexToBytes(
```
moeodeh3, line 290: shouldn't need this once you start throwing errors

packages/crypto/src/webhooks.ts
```diff
@@ -0,0 +1,266 @@
...
+    return {
+      ok: true,
+      value: uint8ArrayFromHexString(input, expectedLength),
+    };
+  } catch {
+    return invalid(reason);
+  }
+}
+
+function invalid(reason: string): { ok: false; reason: string } {
```
moeodeh3: dont need

packages/crypto/src/webhooks.ts
```diff
@@ -0,0 +1,266 @@
...
+  keyId: string;
+  publicKey: string;
+  algorithm?: "ed25519";
+};
+
+export type TurnkeyWebhookHeaders =
+  | Headers
+  | Record<string, string | string[] | undefined>;
+
+export type TurnkeyWebhookBody = string | Uint8Array | ArrayBuffer;
```
moeodeh3, line 14: why not just make this a string?

packages/crypto/src/webhooks.ts
taylorjdawson: At least for now, no other packages besides crypto uses these types so it makes sense to leave them

packages/crypto/src/webhooks.ts
taylorjdawson: while it does add complexity in our code it does reduce the burden on the developer when using this function in different environments
for instance
In Express / raw node / fastify / Next.js pages `headers` is defined as a plain object `Record<string, string | string[] | undefined>`

whereas in Next.js app router / fetch runtimes `headers` is a fetch header instance with `.get()` `.set()` etc.

So devs can just directly pass in the header object without needing to do any extra work

taylorjdawson:
@moeodeh3 For the validation errors, you mentioned throwing an error instead of returning `{ ok: false; reason: string }`
I am open to that if we have a strong argument. Currently, I am in favour of keeping it a structured error response rather than throwing a custom error. With this structured error response, we get stronger typing on errors (we can't enforce in the function signature the thrown errors), reserves real exceptions for true bugs or broken runtimes, and creates a better dev experience. One thing I would add if we decide to stick with the current implementation are enums for the different error states

packages/crypto/src/webhooks.ts
taylorjdawson: (see comment below)

packages/crypto/src/webhooks.ts
taylorjdawson: the reason for this is that it matches real runtimes:
```
// Express / Node with express.raw()
body: req.body
// Buffer, which is a Uint8Array subclass

// Next App Router / Workers / Fetch runtimes
body: await request.arrayBuffer()
```

it enables users to pass in the body directly without needing to do the decode the raw bytes before validation. For instance:
```
// Express / Node with express.raw()
body: req.body.toString("utf8")

// Next App Router / Workers / Fetch runtimes
body: new TextDecoder().decode(await request.arrayBuffer())
```
which IMO is better ergonomics and devX

ethankonk (dismissed):

ethankonk (dismissed):

ethankonk (approved):

PR #1319, opened 2026-05-19
Chore: Leverage turbo cache for CI build

janjakubnanista:
## Summary & Motivation

Currently we are not using `turbo` build cache at all. This PR:

- Adds `turbo` caching to the CI
- Removes the dependency of `build` on `clean` - this introduces cache breakage & graph traversal errors (packages being cleaned after being built)
- Since we now have the builds cached, there is no need to upload/download build artifacts for `test-*` jobs in `js-build.yaml` - if they depend on `build`, they can leverage the cache

[First build (cache not present)](https://github.com/tkhq/sdk/actions/runs/26180674985/job/77022579562?pr=1319)
[Second build (just a UI rerun of the first one, cache present)](https://github.com/tkhq/sdk/actions/runs/26180674985/job/77025218916)

## How I Tested These Changes

Locally + CI

## Did you add a changeset?

Dev changes only


ethankonk (approved):

moeodeh3 (approved):
LGTM, only thing I’ll mention is that in a follow-up pr we should also use Turbo cache in `version-and-publish.yaml`

right now it uses `pnpm build-all` rather than `pnpm build` and it takes forever....

PR #1321, opened 2026-05-19
Chore: Leverage turbo cache in version workflow

janjakubnanista:
## Summary & Motivation

Bringing the cached builds to the release workflow:

- Dropping the artifact upload/download
- Instead of checking out the `main` branch, we limit the whole workflow to `main` only

## How I Tested These Changes

Before doing anything, I disabled the `publish` and `prepare-release` jobs. I then set the target branch to this one.

1. [Test that the workflow fails if not run on `main`. This replicates the old behavior of always checking out the default branch](https://github.com/tkhq/sdk/actions/runs/26116083960/job/76805601383)
2. Then I modified the target branch and commented out the release steps
3. [Then I ran the workflow](https://github.com/tkhq/sdk/actions/runs/26189088716)

## Did you add a changeset?

Dev changes only


.github/workflows/version-and-publish.yml
```diff
@@ -7,8 +7,7 @@ on:
         description: "Version tag (leave empty to auto-generate v{year}.{month}.{counter})"
         required: false
         type: string
-  pull_request:
-    types: [closed]
+  push:
```
ethankonk: This change might mess with the flow we're trying to achieve with this workflow (or at least partially)

- This release workflow is meant to be used in two ways
  a. A "release branch" is created (the release branch is named according to the version we are release, e.g. `release/v2026.5.1`) the version found in the release branch's name is extracted and used for the tag (in this case, we would tag the release as `v2026.5.1`
  b. The flow used to test this PR using the manual invocation, where the version is explicitly inputted

- I don't think `GITHUB_HEAD_REF` is set on `push` events, which would mess up flow A

ethankonk: You can see where we need `GITHUB_HEAD_REF` [here](https://github.com/tkhq/sdk/blob/main/.github/workflows/version-and-publish.yml#L59)

janjakubnanista: Yeah that's what i was unsure about. How come we always checked out main then?

ethankonk: I believe thats because the workflow is meant to run on branch merge, so the changes would already be in main, we just use the latest ref to get the metadata stuff

ethankonk (approved):

PR #1324, opened 2026-05-20
Bring API types in

janjakubnanista:
## Summary & Motivation

Brings API types from current `main` in.

Unfortunately, a tiny yet breaking change has been introduced in [this mono commit](https://github.com/tkhq/mono/commit/b8623c4e86e061d23137769b04b7ff17b67f3f56) and [this mono commit](https://github.com/tkhq/mono/commit/5b629656dbec48e2b51777db9a78bdcf9b979521) - this means a major version bump to `sdk-types` package

## How I Tested These Changes

CI

## Did you add a changeset?

Indeed


.changeset/green-baboons-deny.md
```diff
@@ -0,0 +1,51 @@
...
+  - `solana:mainnet`
+  - `solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp`
+  - `solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d`
+  - `solana:devnet`
+  - `solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1`
+  - `solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1wcaWoxPkrZBG`
+- Introduces new SDK methods:
+  - TVC (these methods will only work if you are part of the [TVC beta](https://www.turnkey.com/turnkey-verifiable-cloud#waitlist))
+    - `getTvcApp`
+    - `stampGetTvcApp`
```
andrewkmin: usually the `stamp*` support is implied and we don't broadcast/mention it explicitly. I'd request either removing them altogether, or just organizing them such that you have all "normal / full" requests, and then all stamp* requests after. like

```
get...
get...
...
stamp...
stamp...
```

.changeset/green-baboons-deny.md
```diff
@@ -0,0 +1,51 @@
+---
+"@turnkey/sdk-browser": major
+"@turnkey/sdk-server": major
+"@turnkey/sdk-types": major
+"@turnkey/core": major
+"@turnkey/http": major
+---
+
+- Makes `nonce` parameter in `oauth2Authenticate` method parameters required.
+- Solana transactions are now limited to the following `caip2`:
```
andrewkmin: to be honest I'm not fully sure what this means. I think it would help to add a bit more context or perhaps narrow this comment down to indicate that this is applicable only for those using Turnkey's Transaction Management

.changeset/green-baboons-deny.md
```diff
@@ -0,0 +1,51 @@
+---
+"@turnkey/sdk-browser": major
+"@turnkey/sdk-server": major
+"@turnkey/sdk-types": major
+"@turnkey/core": major
+"@turnkey/http": major
+---
+
+- Makes `nonce` parameter in `oauth2Authenticate` method parameters required.
```
andrewkmin: Given that this is arguably the most notable and non-routine change of the changelog (given it requires user action), I think we could add a `Breaking change` notice similar to how we've done in the past like: https://github.com/tkhq/sdk/blob/main/packages/sdk-server/CHANGELOG.md#breakingbehavioral-changes

.changeset/green-baboons-deny.md
```diff
@@ -0,0 +1,71 @@
+---
+"@turnkey/sdk-browser": major
+"@turnkey/sdk-server": major
+"@turnkey/sdk-types": major
+"@turnkey/core": major
```
moeodeh3: okay, I know this technically isn’t supposed to happen, but IMO we should not mark this as a major because of the nonce changes

I know these are technically breaking changes, but the impact/radius here is relatively small, and we just released major changes recently. I’m a bit worried about upgrade fatigue

when people see a major version bump, they become much more hesitant to update because they expect things to break, whereas for like 99% of users this really shouldn’t affect them

moeodeh3: actually, this is already enforced at the API level as well, so if people weren’t setting it their integrations would already be broken today

so that gives even more confidence here, anyone with a working Twitter/Discord implementation today must already be setting a nonce!

moeodeh3 (changes requested):

.changeset/green-baboons-deny.md
janjakubnanista: Happy to make it minor again. What worries me though is that we are deviating from strict semver

examples/with-x/src/app/auth/turnkey/x/route.ts
```diff
@@ -71,6 +73,7 @@ export async function POST(req: Request) {
         redirectUri: process.env.X_REDIRECT_URI!,
         codeVerifier: "base64_encoded_sha256(code_verifier)", // in production this value should be a random value and the codeChallenge will be the base64_encoded_sha256(code_verifier)
         bearerTokenTargetPublicKey: keypair.publicKeyUncompressed, // NOTE: This only needs to be provided if you would like the encrypted bearer token to be returned via the `enctypedBearerToken` claim of the OIDC ID Token
+        nonce: bytesToHex(sha256(keypair.publicKey)),
```
moeodeh3: ahh this won’t work! The nonce has to be the SHA256 of the session keypair that we later call `oauth_login` with

so instead we should do:

```ts "
    const sessionPublicKey = body.public_key;
    const nonce = bytesToHex(sha256(sessionPublicKey));
```

then in `oauth_login` we can rename things to use `sessionPublicKey` for readability 🙂

examples/with-x/package.json
```diff
@@ -11,6 +11,8 @@
     "credential-upload": "npx tsx credential-upload.tsx"
   },
   "dependencies": {
+    "@noble/curves": "1.9.7",
+    "@noble/hashes": "^1.8.0",
```
moeodeh3, line 15: then we shouldn't need either of these

janjakubnanista: you still need to hash the publickey right?

moeodeh3: yea, whoops!

moeodeh3 (approved):
LGTM

andrewkmin (approved):
🚢

PR #1326, opened 2026-05-20
Fix with-x example

janjakubnanista:
## Summary & Motivation

[As it transpired in another PR](https://github.com/tkhq/sdk/actions/runs/26175985864/job/77006082387?pr=1319#step:4:646), `with-x` example was a bit borken (the error in that build only shows the fact that the example was created on a case-insensitive fs but there were couple more type problems once that has been resolved):

- `react` types needed a bump
- `typecheck` script was missing 
- unused `asChild` prop was removed from a `Button` component since it was also causing some type issues. The no longer used radix `Slot` component was removed as well

## How I Tested These Changes

Locally + CI

## Did you add a changeset?

Dev changes only


moeodeh3 (dismissed):
🚢

moeodeh3 (approved):

PR #1327, opened 2026-05-20
Fix codegen

janjakubnanista:
## Summary & Motivation

During the latest migration of SDK types, [this error entered the stage](https://github.com/tkhq/sdk/actions/runs/26192412315/job/77063842840?pr=1324#step:6:6825). It looks like type definitions were only collected from the "main" swagger file and the definitions from auth proxy were not considered

## How I Tested These Changes

Locally

## Did you add a changeset?

These changes will have their changeset in the PR that publishes the new Spark types #1324


andrewkmin (approved):
lgtm in isolation, but we may want to add the same change to the other codegen scripts as well

PR #1328, opened 2026-05-20
Fix package manager field & remove unused package locks

janjakubnanista:
## Summary & Motivation

`packageManager` field was supplied in places where it does not belong, causing `pnpm` to be [redownloaded when tasks were run in these packages](https://github.com/tkhq/sdk/actions/runs/26195160000/job/77072803143?pr=1327#step:6:716).

On top of that `package-lock.json` files were left over from, I guess, a legacy NPM-based setup.

Although socket below reports dependency changes, none of them are real since these locks have not had effect on `pnpm i` - [no new packages have been downloaded](https://github.com/tkhq/sdk/actions/runs/26235501776/job/77211160490?pr=1328#step:3:108)

## How I Tested These Changes

[CI](https://github.com/tkhq/sdk/actions/runs/26195503268/job/77073834132?pr=1328#step:6:961)

## Did you add a changeset?

Dev changes only


examples/with-react-native-wallet-kit/package-lock.json
```diff
@@ -1,15404 +0,0 @@
-{
-  "name": "with-react-native-wallet-kit",
```
moeodeh3, line 2: okay this one is a bit weird, but we actually have to use `npm` for this example because we haven’t figured out how to get React Native native modules and `pnpm` working together yet

we know it’s possible, but we haven’t had time to fully solve it. We had an attempt here: https://github.com/tkhq/sdk/pull/1205, but it ran into its own issues and we ended up dropping it because we didn’t want to spend too much time on it

anyways TLDR: having a `package-lock.json` here is actually fine

janjakubnanista: Okay lemme take a look, maybe we can get this to work

janjakubnanista: I made the example kinda work but got stuck on asset loading - for some reason the assets were not being resolved into buffers but were passed around as filenames.

Restored the lock in the meantime @moeodeh3

moeodeh3 (approved):

PR #1339, opened 2026-05-27
Spark support: Update Spark signer with the new API methods

janjakubnanista:
## Summary & Motivation

This PR takes the improvements made by @andrewkmin in #1284 and isolates the changes made to the signer only.

Several changes were made to the original implementation:
- Prefetching accounts was a hanging promise (i.e. a memory leak). Prefetching has been removed from this PR since it's already large enough and will be added as a separate piece later on
- The key management got its own (tested) class/classes that combines a `Map` and server-side account fetching into a read/write safe object

## How I Tested These Changes

Added unit tests for the `TurnkeyValueManager`, that's that

## Did you add a changeset?

This package is not being published yet.

### DIFF
--- a/examples/with-spark/src/index.ts
+++ b/examples/with-spark/src/index.ts
@@ -47,7 +47,8 @@ async function main() {
     "API_PUBLIC_KEY",
     "API_PRIVATE_KEY",
     "ORGANIZATION_ID",
-    "TURNKEY_IDENTITY_ADDRESS",
+    "TURNKEY_SPARK_ADDRESS",
+    "TURNKEY_ECDSA_ADDRESS",
     "IDENTITY_PUBLIC_KEY_HEX",
     "RECEIVER_SPARK_ADDRESS",
   ] as const;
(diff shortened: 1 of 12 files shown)


examples/with-spark/src/index.ts
```diff
@@ -47,7 +47,8 @@ async function main() {
     "API_PUBLIC_KEY",
     "API_PRIVATE_KEY",
     "ORGANIZATION_ID",
-    "TURNKEY_IDENTITY_ADDRESS",
+    "TURNKEY_SPARK_ADDRESS",
```
andrewkmin, line 50: nonblocking, esp as I'm aware this is my own doing 😛 : in a followup or later commit, I think we could consider renaming some of these vars to make it clearer what their purpose is. ie it might be confusing between TURNKEY_SPARK_ADDRESS, TURNKEY_ECDSA_ADDRESS, and IDENTITY_PUBLIC_KEY_HEX

andrewkmin (approved):

examples/with-spark/src/index.ts
janjakubnanista: Yeah agreed - I only did the minimum here not to make this PR larger

PR #1351, opened 2026-05-27
Chore: drop unused turbo.json & tsconfig.typecheck.json files

janjakubnanista:
## Summary & Motivation

These are either no longer used (`tsconfig.typecheck.json`) or do not have a meaningful purpose (`turbo.json` overrides)

## How I Tested These Changes

CI

## Did you add a changeset?

Dev changes only


moeodeh3 (approved):
nice 🔥

PR #1354, opened 2026-05-28
Chore: use base tsconfig for examples

janjakubnanista:
## Summary & Motivation

Lots of the examples were not using `tsconfig.base.json`. When plugged in, there was a bunch of type errors that needed fixing (mostly adding `!` since I assume the code is working and it just needed a bit of a forceful persuasion to compile; adding `type` to type-only imports, removing unused variables etc).

Of course we would ideally not have any `!` in the code but the scope of this problem is massive, this is just a starting point.

## How I Tested These Changes

Locally + CI

## Did you add a changeset?

Dev changes only

### DIFF
--- a/examples/react-wallet-kit/src/components/demo/AuthButtons/AuthenticatorButton.tsx
+++ b/examples/react-wallet-kit/src/components/demo/AuthButtons/AuthenticatorButton.tsx
@@ -1,7 +1,6 @@
 import { KeySVG } from "@/components/Svg";
 import { AuthToggleButton } from "./index";
 import { useTurnkey } from "@turnkey/react-wallet-kit";
-import { toast } from "react-toastify";
 
 export default function AuthenticatorButton({
   canRemoveAuthMethod,
@@ -10,9 +9,6 @@ export default function AuthenticatorButton({
 }) {
   const { user, handleAddPasskey, handleRemovePasskey } = useTurnkey();
   const authenticator = user?.authenticators?.[0];
-
-  const notify = () =>
-    toast.error("Error: Cannot remove your last authenticator.");
 
   return (
     <AuthToggleButton
--- a/examples/with-x/src/app/auth/turnkey/x/route.ts
+++ b/examples/with-x/src/app/auth/turnkey/x/route.ts
@@ -100,7 +100,7 @@ export async function POST(req: Request) {
       filterValue: oauth2AuthenticateResponse.oidcToken,
     });
 
-    let subOrgId;
+    let subOrgId: string;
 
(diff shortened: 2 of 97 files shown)


examples/react-wallet-kit/src/components/demo/AuthButtons/AuthenticatorButton.tsx
```diff
@@ -11,9 +10,6 @@ export default function AuthenticatorButton({
   const { user, handleAddPasskey, handleRemovePasskey } = useTurnkey();
   const authenticator = user?.authenticators?.[0];
 
-  const notify = () =>
-    toast.error("Error: Cannot remove your last authenticator.");
-
```
ethankonk, line 16: wondering why this toast notif was removed from this example?

examples/with-x/src/app/auth/turnkey/x/route.ts
```diff
@@ -164,20 +164,20 @@ export async function POST(req: Request) {
...
-  const claims = JSON.parse(json) as Record<string, unknown>;
+//   const json = Buffer.from(padded, "base64").toString("utf8");
+//   const claims = JSON.parse(json) as Record<string, unknown>;
 
-  // Your example token actually has "encrypted_bearer_token".
-  return claims["encrypted_bearer_token"] as string | undefined;
-}
+//   // Your example token actually has "encrypted_bearer_token".
+//   return claims["encrypted_bearer_token"] as string | undefined;
+// }
```
ethankonk, line 183: why was this commented out?

ethankonk (commented):
Two nits but otherwise looks amazing!

examples/with-x/src/app/auth/turnkey/x/route.ts
janjakubnanista: because the code that calls it is commented out (but we want to preserve it for reference)

examples/react-wallet-kit/src/components/demo/AuthButtons/AuthenticatorButton.tsx
janjakubnanista: no idea, tsc was just complaining about it being unused

ethankonk: hmm thats weird... maybe we just added it and never hooked it up 😅

examples/with-x/src/app/auth/turnkey/x/route.ts
ethankonk: ahhh okok, makes sense

ethankonk (approved):

PR #1360, opened 2026-06-01
Chore: Create reusable test workflows

janjakubnanista:
## Summary & Motivation

The current setup had a large blocks of workflow code duplicated across multiple files (`js-build`, `version-and-publish`).

This PR:
- Isolates the `build` step into a reusable workflow that can be called from both the build & version workflows
- Creates a "blank" `reusable-test.yml` workflow that has all the code that runs the tests. Two flavours of this workflow are then introduced:
  - `reusable-test-prod` that calls `reusable-test` with prod secrets
  - `reusable-test-pre-prod` that calls `reusable-test` with pre-prod secrets
- `anvil` is no longer started in the same shell as the `pnpm test` command - instead, it is started as a `service`
- `globalPassthroughEnv` is adjusted so that `turbo` can be used to run `test` scripts. Without this, `turbo` would not allow any non-declared environment variables into the tests

## How I Tested These Changes

CI

## Did you add a changeset?

Dev changes only


.github/workflows/reusable-test-pre-prod.yml
```diff
@@ -0,0 +1,25 @@
+name: Test (pre-prod)
+
+on:
+  workflow_call:
+
+env:
+  NODE_ENV: production
```
ethankonk, line 7: smol question, the `NODE_ENV` for the previous tests used to be `development`, what's the reason for the switch here?

.github/workflows/reusable-test-prod.yml
```diff
@@ -0,0 +1,25 @@
+name: Test (prod)
+
+on:
+  workflow_call:
+
+env:
+  NODE_ENV: production
```
ethankonk, line 7: same here

.github/workflows/reusable-test.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Setup JS
+        uses: ./.github/actions/js-setup
+
+      - name: Build
+        run: pnpm build --filter='./packages/*'
+
+      - name: Test
+        run: pnpm test --concurrency=1
+        env:
+          NODE_ENV: development
```
ethankonk, line 52: oh wait now I'm confused, do the previous `NODE_ENV`s not matter since it's defined here at the bottom level? What would the reason be for setting `NODE_ENV` to `production` in the workflows that call this be?

.github/workflows/reusable-test-pre-prod.yml
ethankonk: Actually this might not be an issue, I see that the root workflow already sets `NODE_ENV` to `production` but then passes a different one for the test specifically which your changes mimic

ethankonk (approved):

.github/workflows/reusable-test.yml
janjakubnanista: `production` is the default for the build step (and anything else for that matter). It was set because without it, one of the examples fails to build (`next.js` static page data collection fails).

`development` is set for the tests since somewhere we force `https` when this is set to `production` (and `anvil` does not run on `https`) - grep for `DEV_ENVS.includes(process.env.NODE_ENV ?? "")`

ethankonk: ah okok, sounds good!

PR #1361, opened 2026-06-02
Mehdi/eng 4440

mssabr01:
## Summary & Motivation

Bumping deps to resolve vulnerabilities

## How I Tested These Changes

Locally, `pnpm test-all`. There are 5 failing test that also fail on main branch

## Did you add a changeset?

No packages updated


ethankonk (commented):
Quick question, what is the purpose of moving the overrides to the pnpm-workspace?

mssabr01:
> Quick question, what is the purpose of moving the overrides to the pnpm-workspace?

I see this when doing things with pnpm: `[WARN] The "pnpm" field in package.json is no longer read by pnpm.`

It looks like pnpm officially stops caring about pnpm in package.json in 11 and we are on 10.6, but this warning is giving me some concern

ethankonk (dismissed):

ethankonk (approved):

PR #1363, opened 2026-06-02
Chore: bump node & actions

janjakubnanista:
## Summary & Motivation

Node 20 is nearing EOL, let's see if 24 works. Also it comes with `npm@11.6.1`. We also had an outdated `checkout` action that would also stop working soon - on September 16th, 2026 to be precise.

## How I Tested These Changes

CI

## Did you add a changeset?

Dev changes only


.nvmrc
ethankonk, line 1: quick question on this, will this affect people using the SDKs or is this just a dev-only node bump?

ethankonk (approved):

.nvmrc
janjakubnanista: Just a dev bump - we don't specify `engines` anywhere in our package.jsons. This is for us (to get a nicer publishing plus well github will deprecate node 20 soon-ish)

PR #1366, opened 2026-06-02
Chore: Slice up the publish workflow

janjakubnanista:
## Summary & Motivation

Separates the two possible triggers for the publish workflows into:

- `publish-manual.yml` triggered by a button click
- `publish-release-branch.yml` triggered by merging a PR to `main` from a `release/*` branch. An additional `if` was added to make sure this only runs on `release/*` branches once they have been `merged`, not `closed`. Since the value of `github.ref` on a `merged` PR is the merge commit (on `main` branch), we don't need to checkout `${{ github.event.repository.default_branch }}` as was previously the case in the deleted workflow. See the note [here](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request)

The inlined bash code has been moved to script files for easy testability:

- `get-version.sh` gets a version tag from a release branch name
- `check-version.sh` checks that a version conforms to the format `vXXXX.X{X}.X`
- `generate-version.sh` auto generates a version tag

Since we have node 24 in, the extra `npm` installation is also no longer necessary. This PR then removes the need for #1321 since all the caching will work nicely.

## How I Tested These Changes

CI

## Did you add a changeset?

Dev changes only


.github/workflows/publish-release-branch.yml
```diff
@@ -0,0 +1,40 @@
...
+permissions:
+  contents: write
+  packages: write
+  id-token: write
+
+jobs:
+  version:
+    runs-on: ubuntu-latest
+    # Only run this job if the PR was merged and the branch name starts with "releases/"
+    if: github.event.pull_request.merged == true && startsWith(github.head_ref, 'releases/')
```
ethankonk: Release branch prefix is `release` singular

.github/workflows/reusable-publish.yml
```diff
@@ -0,0 +1,106 @@
+name: Test, Build, Publish
+
+on:
+  workflow_call:
+    inputs:
+      version:
+        description: "Version tag (e.g. v2024.06.0)"
```
ethankonk: nit: we usually do single digit month releases as `v2024.6.0` rather than `v2024.06.0` but that's a small technicality, if it's a lot of work to get it this way you can ignore this comment!

.github/workflows/reusable-publish.yml
```diff
@@ -0,0 +1,106 @@
...
+        run: |
+          echo "Publishing packages to NPM..."
+          pnpm publish -r --no-git-checks --report-summary || {
+            echo "❌ Publish failed. Check logs above.";
+            exit 1;
+          }
+          echo "npm publish summary:"
+          cat pnpm-publish-summary.json || echo "No NPM publish summary generated"
+
+  prepare-release:
```
ethankonk: This was an issue in our previous `version-and-publish` workflow but since we're reworking it, could we rename this job to `publish-gh-release` instead? Would be a more accurate title

.github/workflows/reusable-publish.yml
```diff
@@ -0,0 +1,106 @@
...
+    runs-on: ubuntu-latest
+    needs: [test-pre-prod, test-prod]
```
ethankonk: We also only want to run this job after the NPM publish succeeds. This job just creates the gh release found [here](https://github.com/tkhq/sdk/releases) so if the publish job fails, we wouldn't want the tag/release to be created juuuust yet.

.github/workflows/reusable-publish.yml
ethankonk: Ah, this description and `check-version` comment might just be out of date since I see the version check checks for `vXXXX.X.X`, if so you can downgrade this to a super nit 😆

.github/workflows/publish-release-branch.yml
janjakubnanista: Fixed

.github/workflows/reusable-publish.yml
janjakubnanista: Fixed

.github/workflows/reusable-publish.yml
janjakubnanista: Aaand fixed

.github/workflows/reusable-publish.yml
janjakubnanista: And fixed as well

ethankonk (approved):

PR #1377, opened 2026-06-04
Spark support: Add batch refund tx signing capability to signer

janjakubnanista:
## Summary & Motivation

Adds strongly typed functionality for signing refund transactions for claims, transfers, lightning interactions & coop exits to `TurnkeySparkSigner`

## How I Tested These Changes

Using the example (changes not part of this PR)

## Did you add a changeset?

Yea

### DIFF
--- a/packages/spark/src/signer.ts
+++ b/packages/spark/src/signer.ts
@@ -1214,11 +1214,7 @@ export class TurnkeySparkSigner implements SparkSigner {
       ]);
     };
 
-    return this.signRefundsBatchedCore(
-      input,
-      createRefundTxs,
-      createMessage,
-    );
+    return this.signRefundsBatchedCore(input, createRefundTxs, createMessage);
   }
 
   async signRefundsBatchedClaim(input: SignRefundsBatchedClaimInput) {
--- a/packages/spark/src/signer.ts
+++ b/packages/spark/src/signer.ts
@@ -59,6 +59,18 @@ import {
   type SubtractSplitAndEncryptResult,
   type VerifiableSecretShare,
   type SigningNonce,
+  Network,
+  getTxFromRawTxBytes,
+  SparkValidationError,
+  getCurrentTimelock,
+  createConnectorRefundTxs,
+  createCurrentTimelockRefundTxs,
+  getSigHashFromMultiInputTx,
+  getSigHashFromTx,
+  getNextHTLCTransactionSequence,
+  createRefundTxsForLightning,
+  getNetwork,
+  createDecrementedTimelockRefundTxs,
 } from "@buildonspark/spark-sdk";
 import {
   uint8ArrayToHexString,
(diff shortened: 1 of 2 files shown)


packages/spark/src/signer.ts
```diff
@@ -971,3 +1489,15 @@ const statechainCommitmentToSparkFrostCommitment = (
     binding: uint8ArrayToHexString(commitment.binding),
   };
 };
+
+const emptyCommitment = (): SigningCommitmentWithOptionalNonce => ({
```
andrewkmin, line 1493: small nit: is it worth moving these to a separate consts/utils?

packages/spark/src/signer.ts
```diff
@@ -971,3 +1489,15 @@ const statechainCommitmentToSparkFrostCommitment = (
...
+  commitment: {
+    hiding: new Uint8Array(33),
+    binding: new Uint8Array(33),
+  },
+});
+
+const dontModifySignFrostParams: ModifySignFrostParams = (params) => params;
```
andrewkmin, line 1500: would like a comment here or at callsite for some added context

packages/spark/src/signer.ts
```diff
@@ -81,6 +93,14 @@ import {
   TurnkeySparkSigningKeyManager,
 } from "./key-management";
 import { areBytesEqual, compressedPublicKeyHexFromString } from "./utils";
+import type {
+  ClaimPackage,
+  SigningResult,
+  TreeNode,
+  UserSignedTxSigningJob,
+} from "@buildonspark/spark-sdk/dist/proto/spark";
```
andrewkmin, line 101: is `.../dist/...` the only/best place we can pull these in from?

packages/spark/src/signer.ts
```diff
@@ -216,6 +236,87 @@ export interface TurnkeySparkSignerOptions {
...
+  tx: Transaction;
+  nodeOutput: TransactionOutput;
+}
+
+type CreateRefundTxMessage = (
+  params: CreateRefundTxMessageParams,
+  index: number,
+) => Uint8Array;
+
+type ModifySignFrostParams = (params: SignFrostParams) => SignFrostParams;
```
andrewkmin, line 274: This is probably a relic of being over-generalized in my initial PR. Down to remove this at some point if we need to / if we just need minimal modifications to the params. Otherwise, a nonblocker

andrewkmin (approved):
probably nearing the point where we start needing unit tests. but that will be a team effort! 🚢

packages/spark/src/signer.ts
janjakubnanista: Oh no this is a legit type, will add some comments

packages/spark/src/signer.ts
janjakubnanista: Yeah their package is not configured well enough but I'll try and see if I can import it from a better place in the next PR

packages/spark/src/signer.ts
janjakubnanista: I'll add a bunch of comments in the next PR

packages/spark/src/signer.ts
janjakubnanista: they are not being used anywhere else, better keep things contained

PR #1378, opened 2026-06-05
Am/examples/with stellar

am-hernandez:
## Summary & Motivation

Adds a new `examples/with-stellar` example. It's a single-page Next.js app using `@turnkey/react-wallet-kit` that demonstrates end-to-end Stellar wallet functionality on testnet:

- Sub-organization creation with a Stellar HD wallet (`ADDRESS_FORMAT_XLM`, `CURVE_ED25519`, path `m/44'/148'/0'/0'/0`) provisioned at sign-up
- Wallet info display: user ID, sub-org ID, wallet ID, and Stellar account addresses (linked to Stellar Expert)
- Arbitrary message signing using the react-wallet-kit `handleSignMessage` modal
- Native XLM payment on testnet: builds a `TransactionBuilder` envelope, computes the SHA-256 transaction hash, signs the raw 32-byte hash directly via `httpClient.signRawPayload` (bypassing the UTF-8 re-encoding in `signMessage`), reconstructs the 64-byte `ed25519` signature from `r` + `s`, attaches a `DecoratedSignature`, and submits via Horizon
- Testnet funding via Stellar Friendbot

Closes [CUS-184](https://linear.app/turnkey/issue/CUS-184/add-a-with-stellar-example)

## How I Tested These Changes

Ran `pnpm dev` inside `examples/with-stellar`, authenticated via email OTP and passkey, verified wallet and account display, signed a message, funded via Friendbot, and submitted a native XLM payment to a second testnet account. Confirmed the transaction on [Stellar Expert testnet](https://stellar.expert/explorer/testnet).

## Did you add a changeset?

No — this PR only adds a new example app and does not modify any published packages.


examples/with-stellar/tsconfig.json
```diff
@@ -0,0 +1,29 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
t-vila: target: "es5" in tsconfig.json:3 is the legacy Next.js default but unnecessary on Next 15 + React 19. Comparable new examples target es2017 or higher.

examples/with-stellar/README.md
```diff
@@ -0,0 +1,96 @@
...
+- Signing an arbitrary message with an ed25519 Stellar key
+- Building, signing, and submitting a native XLM payment on Stellar testnet
+
+The app is a single-page Next.js application. Authentication is handled by `@turnkey/react-wallet-kit` with no custom backend required. Turnkey's managed [Auth Proxy](https://docs.turnkey.com/reference/auth-proxy) handles sign-up and login securely from the browser.
+
+## Getting started
+
+### 1/ Cloning the example
+
+Make sure you have `Node.js` installed locally; we recommend using Node v18+.
```
t-vila: Next 15.5 requires Node 18.18+ - let's maybe just go with Node v20+, I think that what we started to do recently in other examples

t-vila: let's also add a .gitignore file

examples/with-stellar/src/components/SignMessage.tsx
```diff
@@ -0,0 +1,88 @@
...
+    }
+    setError(null);
+    setSignature(null);
+    setLoading(true);
+    try {
+      const { r, s } = await handleSignMessage({
+        message,
+        walletAccount: stellarAccount,
+        // ed25519 requires HASH_FUNCTION_NOT_APPLICABLE
+        encoding: "PAYLOAD_ENCODING_TEXT_UTF8" as any,
```
t-vila: not sure if we need `cast as any` here - they should be properly typed ?
generally I think we should make sure to have proper typing anywhere and expose them if missing, avoid masking them with `as any`

examples/with-stellar/src/components/SendTransaction.tsx
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  async function handleFund() {
+    if (!stellarAccount?.address) return;
+    setFunding(true);
+    setError(null);
+    try {
+      await fetch(`${FRIENDBOT_URL}/?addr=${stellarAccount.address}`);
+      // Always refresh balance — a non-ok response just means the account
+      // was already funded, which is fine.
```
t-vila: Friendbot error path silently succeeds — SendTransaction.tsx:52-66 calls fetch(...) without checking response.ok. fetch resolves on non-2xx, so a real Friendbot failure (rate-limited, service down) is swallowed and the UI just reads back the unchanged balance with no indication. The comment on 58-59 is half-right — "already funded" is fine, but other failures should surface. Suggest: read response.json(), treat only the "already funded" error code as benign; otherwise call setError.

examples/chain-integrations/with-stellar/src/components/SendTransaction.tsx
```diff
@@ -0,0 +1,233 @@
...
+            {funding ? "Funding…" : "Fund with Friendbot"}
+          </button>
+        </div>
+      )}
+
+      <div className="flex flex-col gap-3">
+        <div className="flex flex-col gap-1">
+          <label className="text-sm text-gray-600">Destination Address</label>
+          <input
+            value={destination}
```
t-vila, line 218: destination input is unvalidated — SendTransaction.tsx:181-185 accepts any string and passes it directly to Operation.payment. The Stellar SDK will throw inside Operation.payment for invalid addresses, but the error bubbles up as a raw Error message. Minor: a StrKey.isValidEd25519PublicKey(destination) check before building the tx would produce a clearer user-facing error.
Also, same nit, amount not sanitized against XLM's 7-decimal precision — input type="number" min="0.0000001" step="0.1" allows e.g. 0.12345678 which the Stellar SDK will reject (invalid amount).

examples/chain-integrations/with-stellar/src/components/SendTransaction.tsx
```diff
@@ -0,0 +1,233 @@
...
+      // Reconstruct the 64-byte ed25519 signature from r and s
+      const rHex = r.startsWith("0x") ? r.slice(2) : r;
+      const sHex = s.startsWith("0x") ? s.slice(2) : s;
+      const sigHex = rHex.padStart(64, "0") + sHex.padStart(64, "0");
+      const signatureBytes = Uint8Array.from(
+        sigHex.match(/.{2}/g)!.map((b) => parseInt(b, 16)),
+      );
+
+      // Attach signature to transaction
+      const keypair = Keypair.fromPublicKey(stellarAccount.address);
```
t-vila, line 160: Keypair.fromPublicKey(stellarAccount.address) only to extract the hint, but signatureHint() is just the last 4 bytes of the raw public key - probably worth a comment for devs to understand

examples/with-stellar/src/components/SignMessage.tsx
am-hernandez: agreed! fixing up all instances of `any`

examples/with-stellar/tsconfig.json
am-hernandez: resolved here: 1f7cc6a2564c634437e272396cc582821032b5e0

examples/with-stellar/README.md
am-hernandez: resolved here: 9b72bda0d7058e386befbc4b616098eea8e14bc2

examples/with-stellar/src/components/SignMessage.tsx
am-hernandez: resolved here: 99be14501ae1c5f0695927df5810a6f20639a3c2

examples/with-stellar/src/components/SendTransaction.tsx
am-hernandez: resolved here: 0e447dff3454ddc58b1405f4ac6c12be4638e29f

examples/chain-integrations/with-stellar/src/components/SendTransaction.tsx
am-hernandez: resolved here: 501fca2676908179dd1102076e567388adf643de

examples/chain-integrations/with-stellar/src/components/SendTransaction.tsx
am-hernandez: resolved here: 9d58d930bea18fd29376bbc20fd17150a9a9f9bb

t-vila:
@am-hernandez quick one and then we can merge - add examples/with-stellar/.eslintrc.json with `{ "extends": "next/core-web-vitals" }`
that activates the lint script and validates the existing eslint-disable directive ex https://github.com/tkhq/sdk/blob/main/examples/oauth/.eslintrc.json

examples/chain-integrations/with-stellar/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+Make sure you have `Node.js` installed locally; we recommend using Node v20+.
+
+```bash
+$ git clone https://github.com/tkhq/sdk
+$ cd sdk/
+$ corepack enable  # Install `pnpm`
+$ pnpm install -r  # Install dependencies
+$ pnpm run build-all  # Compile source code
+$ cd examples/with-stellar/
```
t-vila: nit `cd examples/chain-integrations/with-stellar/`

t-vila (dismissed):

examples/chain-integrations/with-stellar/README.md
am-hernandez: resolved here: 8f3ae972156673dcbeb4dfd0722364130efe619c

t-vila (approved):

PR #1380, opened 2026-06-06
Am/update example oauth cross platform

am-hernandez:
## Summary & Motivation

Updates to the `examples/oauth-cross-platform` web example.

**Auto-link platforms after login**

Replaces `PlatformsCard` + `VerificationModal` with a new `LinkedPlatformsCard` component that checks and silently links unregistered platform `oidcClaims` on mount. No user action required. Each platform shows a live badge: `verified`, `oidcClaim - linked`, `oidcClaim - checking...`, `oidcClaim - linking...`, or `oidcClaim - error`.

Uses `httpClient.createOauthProviders` to add more identities with the same issuer and subject but different audience.

**Delete sub-org warning in READMEs**

Adds a warning section to the root, web, and mobile READMEs clarifying that sub-org deletion can only be initiated by the authenticated user from within the app (not from the Turnkey dashboard), and that the delete button uses `deleteWithoutExport: true`.

## How I Tested These Changes

- Signed up as a new user with all three client IDs configured; all platforms show `oidcClaim - linked` automatically on dashboard load
- Signed in as an existing account that was created before secondary platforms were registered; platforms link automatically and the warning only appears if linking fails
- Verified the delete button still works and warning text is accurate

Closes [CUS-208](https://linear.app/turnkey/issue/CUS-208/oauth-cross-platform-example-auto-link-platforms-post-login-delete-sub)

## Did you add a changeset?

No package changes.


examples/oauth-cross-platform/web/src/components/LinkedPlatformsCard.tsx
```diff
@@ -0,0 +1,175 @@
...
+}: Props) {
+  const { httpClient, session } = useTurnkey();
+
+  const [linkStatus, setLinkStatus] = useState<Record<string, LinkStatus>>({});
+  const [linkErrors, setLinkErrors] = useState<Record<string, string>>({});
+
+  useEffect(() => {
+    if (!claims || !currentSubOrgId || !httpClient || !session?.userId) return;
+
+    for (const p of platforms.filter((p) => !p.verified)) {
```
truham:
```suggestion 
for (const p of platforms.filter((p) => !p.verified && !linkStatus[p.label])) { 
```
Check for linkStatus to prevent re-running when platform is in any status (checking, linking, linked, error)

t-vila (approved):

PR #1383, opened 2026-06-08
Moe o/modal context

moeodeh3:
## Summary & Motivation

branches off of: https://github.com/tkhq/sdk/pull/1162

## Did you add a changeset?


packages/core/src/utils.ts
```diff
@@ -1569,3 +1571,84 @@ export async function resetPasskeyScope(
...
+   * API shape:
+   * {
+   *   listings: {
+   *     [id: string]: WalletConnectApp
+   *   }
+   * }
+   */
+  const listings = json.listings ?? {};
+
+  return Object.values(listings) as WalletConnectApp[];
```
figitaki, line 1606: What about `zod` here instead of unsafe casting?

packages/core/src/utils.ts
```diff
@@ -1569,3 +1571,84 @@ export async function resetPasskeyScope(
...
+  projectId: string,
+  namespaces: string[],
+): Promise<WalletConnectAppEntry[]> {
+  const rawApps = await fetchWalletConnectApps(projectId);
+  const entries: WalletConnectAppEntry[] = [];
+
+  // we derive a unique chains list from the namespaces
+  const chains = new Set<Chain>();
+  for (const ns of namespaces) {
+    if (ns.startsWith("eip155:")) chains.add(Chain.Ethereum);
```
figitaki: nit: An exauhstive switch here would be nice to at least have a `default` clause showing fall through logic in the case we get an unsupported chain.

```
const chainPrefix = ns.split(":")[0];
switch (chainPrefix)
```

packages/core/src/utils.ts
ethankonk: Thats a good idea but would be better added in a follow-up PR to make a full switch rather than just patch this specific case, there are plenty of areas using `zod` would be useful, will add a ticket for this!

moe-dev (approved):

PR #1385, opened 2026-06-08
Stub of @turnkey/bitcoin package

janjakubnanista:
## Summary & Motivation

In order to run full E2E tests for Spark, we'll need to be able to send some bitcoin REGTEST transactions. This PR takes the bitcoin signer out of the example for start and puts it into a new `@turnkey/bitcoin` package.

## How I Tested These Changes

Pure copypasta

## Did you add a changeset?

None needed

### DIFF
--- a/packages/bitcoin/LICENSE
+++ b/packages/bitcoin/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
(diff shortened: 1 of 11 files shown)


packages/bitcoin/src/signer.ts
```diff
@@ -22,25 +22,21 @@ import type { Turnkey as TurnkeyServerSDK } from "@turnkey/sdk-server";
...
    * @param publicKey public key buffer. To sign P2TR outputs it needs to be the decoded address (`bitcoin.address.fromBech32(address).data`).
    *                  For P2WPKH it should be the key pair's underlying public key buffer, uncompressed.
    */
-  constructor(client: TurnkeyServerSDK, address: string, publicKey: Buffer) {
-    this.client = client;
-    this.address = address;
-    this.publicKey = publicKey;
-  }
+  constructor(
+    public readonly client: TurnkeyApiClient,
```
ethankonk: We normally pass in generic or all the http clients that way these packages can work with any of our core client (http clients from: `@turnkey/core`, `@turnkey/http`, `@turnkey/sdk-server`)

If you take a look at our viem and ethers packages you can see that in action.

packages/bitcoin/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2023 Turnkey
```
ethankonk: bump this up to 2026, not sure if it's necessary but might as well

ethankonk (dismissed):
looks good for now! we can keep iterating on it as we go

packages/bitcoin/src/signer.ts
ethankonk: actually we haven't merged the generic client PR yet but the PR for it is here: https://github.com/tkhq/sdk/pull/1203

janjakubnanista: oh I think there's a nicer way of doing that https://github.com/tkhq/sdk/pull/1203/changes#r3383058821

janjakubnanista: check the latest commit @ethankonk and let me know what you think about the generic client types there

packages/bitcoin/LICENSE
janjakubnanista: Done

packages/bitcoin/src/signer.ts
ethankonk: Oh that looks promising! Will do some testing with that quick

ethankonk: I think we've tried a variation of that at some point but ran into some odd issues, will get back to you when I can jump on this. Going to pre-approve now since I think this should work but will follow up if I find any issues!

packages/bitcoin/src/signer.ts
```diff
@@ -56,3 +59,9 @@ export class TurnkeyBitcoinSigner {
     return Buffer.from(r + s, "hex");
   }
 }
+
+export type TurnkeyApiClientBitcoinSigner<
+  T extends TurnkeyApiClientServer | TurnkeyApiClientBrowser =
+    | TurnkeyApiClientServer
+    | TurnkeyApiClientBrowser,
```
ethankonk: actually can we add `@turnkey/core` to this extension as well?

ethankonk: and `@turnkey/http` maybe? might see some issues with that one but hopefully it slots right in

janjakubnanista: The HTTP one is too low-level and has different signatures so that one would not work. I checked the core one but it does not seem to have the same interface - `signRawPayload` is missing.

janjakubnanista: Oke added

ethankonk (approved):
I like this but I still feel we should eventually find a way to include `@turnkey/http` as a compatible client (since that's the norm for every other "wrapper" package).

LGTM tho!

PR #1390, opened 2026-06-09
Spark support: Bump required SDK version

janjakubnanista:
## Summary & Motivation

We can remove some hacks that have been there because the previous version of the Spark SDK hadn't been published at the moment of merging.

This includes `Object.assign` some of our overriden services and moving patched code from wallet to a new `TurnkeyCoopExitService`.

Once the new version was in, the withdraw test started failing. Turns out there was a problem with how the leaves were paritioned and some `directFromCpfpLeaves` ended up in `directLeaves`. This is a bit strange since the new error comes from the backend and that hasn't changed.

## How I Tested These Changes

Existing tests

## Did you add a changeset?

Yes

### DIFF
--- a/packages/spark/src/signer.ts
+++ b/packages/spark/src/signer.ts
@@ -993,6 +993,11 @@ export class TurnkeySparkSigner implements SparkSigner {
       );
     }
 
+    // The jobs array will hold three items per element:
+    //
+    // - the cpfp job at index 0
+    // - the direct job at index 1
+    // - the directFromCpfp job at index 2
     const jobs: Triple<UserUnsignedTxSigningJob | undefined>[] = [];
 
     for (let i = 0; i < leaves.length; i++) {
@@ -1100,6 +1105,7 @@ export class TurnkeySparkSigner implements SparkSigner {
       jobs.push(jobsForLeaf);
     }
 
+    // Now we flatten the jobs array into SignFrostParams[] for the batch signing call
     const signFrostParamsForLeaves: SignFrostParams[] = jobs.flatMap(
       (jobsForLeaf, index) => {
         const leaf = leaves[index]!;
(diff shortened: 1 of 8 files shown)


packages/spark/src/signer.ts
```diff
@@ -1180,14 +1174,9 @@ export class TurnkeySparkSigner implements SparkSigner {
...
-      const isZeroNode = !getCurrentTimelock(nodeTx.getInput(0).sequence);
-      const effectiveDirectNodeTx = isZeroNode ? undefined : directNodeTx;
-
       return createConnectorRefundTxs({
         nodeTx,
-        ...(effectiveDirectNodeTx
-          ? { directNodeTx: effectiveDirectNodeTx }
-          : {}),
+        directNodeTx: directNodeTx!,
         sequence,
```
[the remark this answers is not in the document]
janjakubnanista: The reason behind this is the [`exactOptionalPropertyTypes`](https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes) setting

packages/spark/src/wallet.ts
```diff
@@ -27,6 +17,21 @@ export class TurnkeySparkWallet extends SparkWallet {
       this.config,
       this.connectionManager,
       this.signingService,
+      this.__getLogging__(),
+    );
+
+    this.swapService = new SwapService(
+      this.config,
+      this.transferService,
+      this.sspClient!,
```
omkarshanbhag, line 26: @janjakubnanista is the non-null assertion safe to do? Any checks needed before we do this?

packages/spark/src/wallet.ts
```diff
@@ -38,135 +43,31 @@ export class TurnkeySparkWallet extends SparkWallet {
...
+      this.transferService,
+      this.connectionManager,
+      (balance) => {
+        this.emit(SparkWalletEvent.BalanceUpdate, {
+          available: BigInt(balance.available),
+          owned: BigInt(balance.owned),
+          incoming: BigInt(balance.incoming),
+        });
+      },
+      undefined,
```
omkarshanbhag, line 58: @janjakubnanista any context here on the undefined argument???

omkarshanbhag (approved):
Approved with some comments and questions

packages/spark/src/wallet.ts
janjakubnanista: Ah yeah this is safe, this is being assigned in the parent constructor but their typescript is looser than ours

packages/spark/src/wallet.ts
janjakubnanista: Hehe it's the big comment above:

```typescript
// Disable the SDK's claim-time auto-optimize swap. The optimizer splits a
// local leaf is SWAP_PENDING. Disabling the auto-optimize keeps the claim's
// leaf AVAILABLE so the next operation proceeds. Users who want
// optimization can call wallet.optimizeLeaves() explicitly and await it.
```

PR #1397, opened 2026-06-10
Spark support: Make sure parallel test executions don't step on each other's toes

janjakubnanista:
## Summary & Motivation

Since we are reusing the same account for all tests, we might run into `TRANSFER_LOCKED` errors if we try to transfer funds that are already locked by a pending transfer.

The solution here is using the `asyncRetry`, checking the error on and retrying on `TRANSFER_LOCKED` errors. The rest of the errors trigger the `bail` callback and abort retrying.

See [this workflow run](https://github.com/tkhq/sdk/actions/runs/27298318153/job/80637353586)

## How I Tested These Changes

Locally

## Did you add a changeset?

Test changes only

### DIFF
--- a/packages/spark/src/__tests__/wallet.test.ts
+++ b/packages/spark/src/__tests__/wallet.test.ts
@@ -3,7 +3,11 @@ import {
   TurnkeyApiClient,
   v1WalletAccount,
 } from "@turnkey/sdk-server";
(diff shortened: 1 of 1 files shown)


omkarshanbhag (approved):
Nice Saftety measure!! LGTM

PR #1399, opened 2026-06-11
React Native in workspaces: Start with oauth-cross-platform/mobile [1/N]

janjakubnanista:
## Summary & Motivation

Until now we kept the `react-native` examples out of the `pnpm` workspaces. This PR:

- [Modifies `TurnkeyProvider` in `@turnkey/react-native-wallet-kit`](https://github.com/tkhq/sdk/pull/1399/changes/5dff55f0aecd6258179760b8d1cd5d753d71222e) so that it does not depend on `react-native-device-info` (incompatible with `expo`). This is a backwards-compatible change as it only adds an optional property. Missing `devDependencies` have also been added, just as a little cleanup (until now the package would rely on these implicitly)
- [Updates `expo` & Runs `npx expo-doctor` in `oauth-cross-platform/mobile`](https://github.com/tkhq/sdk/pull/1399/changes/7405e7dd4623349bd4246d605a2f46120cde9f83)
- Adds `typecheck` command and fixes some smaller issues
- [Fixes a new security vulnerability](https://github.com/tkhq/sdk/pull/1399/changes/f4393626e2fd96325bd909efc80d24106103ef58)
- [Fixes build errors that popped up once the examples were plugged in](https://github.com/tkhq/sdk/pull/1399/changes/33a09fabf5d61a91c5dbb00a90cfef0d84d8bf83)
- [Fixes jest config errors that popped up once the examples were plugged in](https://github.com/tkhq/sdk/pull/1399/changes/f78e0fec27520b9ff19842dc4c6110cfbbf1ccb3)

## How I Tested These Changes

Locally on an iOS simulator + CI for typecheck

## Did you add a changeset?

Yes, to `@turnkey/react-native-wallet-kit`


packages/react-native-wallet-kit/package.json
```diff
@@ -50,12 +50,18 @@
...
   "devDependencies": {
     "@babel/core": "7.26.9",
+    "@invertase/react-native-apple-authentication": "~2.5.1",
     "@rollup/plugin-babel": "5.3.0",
+    "@types/react": "18.2.14",
+    "react": "19.1.0",
+    "react-native": "0.81.4",
+    "react-native-gesture-handler": "~2.28.0",
+    "react-native-inappbrowser-reborn": "^3.7.1",
+    "react-native-safe-area-context": "~5.6.0",
```
ethankonk, line 64: Curious why these new dev dependencies were added, do we use them in examples only? Why add them at the package level?

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
```diff
@@ -912,9 +913,7 @@ export const TurnkeyProvider: React.FC<TurnkeyProviderProps> = ({
       const websiteName =
         Platform.OS === "web" && typeof window !== "undefined"
           ? window.location.hostname
-          : DeviceInfo.getApplicationName() ||
-            DeviceInfo.getBundleId() ||
-            "mobile-app";
```
ethankonk, line 917: I don't believe this is breaking but anyone that does rely on passkey names being named based on their device app name will need to update their config with their app name.

Not a huge deal but is there a reason why `react-native-device-info` needs to be removed? Is that the package that blocks us from using `@turnkey/react-native-wallet-kit` in our workspace?

pnpm-workspace.yaml
```diff
@@ -5,14 +5,14 @@ packages:
   - "!examples/demos/with-react-native-wallet-kit/**"
```
ethankonk, line 5: Will it be possible to remove `with-react-native-wallet-kit` from the workspace block with these changes as well? That would be amazing

ethankonk (commented):
This is looking super promising! Can't wait for this!

packages/react-native-wallet-kit/src/providers/TurnkeyProvider.tsx
janjakubnanista: `react-native-device-info` does not play nice with `expo` - you would need to create a custom build for it to run (which is out of scope for an example that does not have a build step). Expo, in a managed workflow, uses `expo-device` rather than `react-native-device-info`. So removing this dependency not only fixes this issue but makes this code compatibility a bit wider

pnpm-workspace.yaml
janjakubnanista: Yeah I am going one by one though

PR #1419, opened 2026-06-23
Remove changelog generation for SDK examples

taylorjdawson:
## Summary
- remove tracked changelogs from private SDK examples
- expand Changesets ignores to cover current example package naming patterns
- add a guard so future workspace examples cannot silently re-enter release/changelog generation
- fix the reusable build workflow to run the repo version script instead of pnpm built-in version command

## Validation
- pnpm build
- pnpm typecheck
- formatting and generated-code cleanliness checks
- Changesets/example ignore guard and release-prep path

Linear: INT-503


.changeset/config.json
```diff
@@ -9,8 +9,16 @@
...
+    "@turnkey/import-export-with-*",
     "@turnkey/kitchen-sink",
-    "@turnkey/with-*"
+    "@turnkey/otp-auth-*",
+    "@turnkey/paymaster-usdc-transfer",
+    "@turnkey/react-wallet-kit-playground",
+    "@turnkey/wallet-auth-*",
+    "@turnkey/with-*",
+    "magic-link-auth",
+    "with-*"
```
ethankonk: I wonder if theres a way we can automate this instead of having to constantly update this ignore list... Would a script that pulls all package names in the `examples` directory and throws them in here be better?

ethankonk (commented):
Actually yeah, I feel a script that automatically detects and adds packages that should be ignored would be super helpful. The regex ignore patterns should stay but for future one examples that don't follow these naming conventions, a simple pnpm script that adds them to the list automatically would be 🔥 

That way we can error out with a message like:

"Run `pnpm run ignore-packages` (or something) to add your example to the changeset ignore list"

Makes it a little more straightforward and fool proof

.changeset/config.json
taylorjdawson: yea good call, i noticed that too, 
 I ended up taking a slightly simpler route and setting the `privatePackages.version: false` in Changesets, since examples are private packages, and then we added the guard to ensure examples stay private 
let me know if you think that makes sense

ethankonk: ohh love that idea, fixes the issues we have with un-published packages attempting to be published via trusted publishers too!

ethankonk (approved):

PR #1422, opened 2026-06-24
react-wallet-kit: Add error message, expiry message and reset, remove extra otp captcha

justinformentin:
## Summary

Three issues this PR solves

### No error message
Currently, when the Cloudflare Turnstile captcha widget errors or expires, the auth buttons get disabled with no explanation to the user, they see a grayed-out form with no indication of why they cannot proceed. We added an error message.

### No expiry handling
When a captcha token expires, nothing happened. We now show a message, and reset the turnstile widget.

### Double captcha
**Current Behavior**
1. User enters email → clicks submit → handleEmailSubmit in index.tsx calls initOtp with consumeToken() → CAPTCHA 1 - (Turnstile widget in index.tsx)
2. User gets OTP code → enters it → handleContinue in OTP.tsx calls completeOtp which internally calls signup → backend requires captcha again → CAPTCHA 2 - (Turnstile widget in OTP.tsx)
**New Behavior**
User enters email → clicks submit → handleEmailSubmit calls initOtp with consumeToken() → CAPTCHA 1 - (Turnstile widget in index.tsx)
User gets OTP code → enters it → handleContinue calls completeOtp which gets a verificationToken from verifyOtp, then passes it to signup → backend sees the verificationToken + email/phone → skips captcha [mono PR #6940](https://github.com/tkhq/mono/pull/6940)

## Changes

1. Extracted <Turnstile> widget into separate component that is shared between auth/index.tsx and OTP.tsx.
2. Added expiry and error message
    a. Error (e.g. network issue, widget failure): shows "Verification failed. Please try again."
    b. Expiry (token timed out): shows "Verification expired — retrying..."
4. Added `turnstileRef.current?.reset()` in onExpire to immediately re-run the challenge. Clears when onSuccess receives new token.
5. Removed `consumeToken` from `handleContinue ` in `OTP.tsx`. With the companion [backend change](https://github.com/tkhq/mono/pull/6940), the captcha check is skipped for email/phone signups when a valid verificationToken is present. Since completeOtp calls verifyOtp internally to obtain a verificationToken before signing up, the backend never validates the captcha token passed through that path. Consuming a captcha token in handleContinue (which calls completeOtp → verifyOtp → signUpWithOtp) is not needed.

### DIFF
--- a/packages/react-wallet-kit/src/components/auth/OTP.tsx
+++ b/packages/react-wallet-kit/src/components/auth/OTP.tsx
@@ -8,8 +8,7 @@ import { OtpType, TurnkeyError, TurnkeyErrorCodes } from "@turnkey/core";
 import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
 import { faEnvelope, faPhone } from "@fortawesome/free-solid-svg-icons";
 import clsx from "clsx";
-import { Turnstile, type TurnstileInstance } from "@marsidev/react-turnstile";
-import { consumeCaptchaToken } from "../../utils/captcha";
+import { useTurnstile } from "./TurnstileWidget";
 
 interface OtpVerificationProps {
   contact: string;
@@ -32,8 +31,7 @@ export function OtpVerification(props: OtpVerificationProps) {
     sessionKey,
     onContinue = null, // Default to null if not provided
   } = props;
-  const { initOtp, completeOtp, config, getTurnstileToken, setTurnstileToken } =
-    useTurnkey();
+  const { initOtp, completeOtp } = useTurnkey();
   const { closeModal, isMobile } = useModal();
   const [submitting, setSubmitting] = useState<boolean>(false);
   const [resending, setResending] = useState<boolean>(false);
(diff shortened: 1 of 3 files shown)


packages/react-wallet-kit/src/components/auth/index.tsx
```diff
@@ -485,40 +487,57 @@ export function AuthComponent({
...
+                setTurnstileToken(null);
+                setShowTurnstileError(true);
+                setTurnstileErrorMessage(
+                  "Verification failed. Please try again.",
+                );
+              }}
+              onExpire={() => {
+                setTurnstileToken(null);
+                setAuthEnabled(false);
+                setShowTurnstileExpired(true);
```
justinformentin: Added widget reset

packages/react-wallet-kit/src/components/auth/OTP.tsx
```diff
@@ -68,7 +62,6 @@ export function OtpVerification(props: OtpVerificationProps) {
           contact,
           otpType,
           ...(sessionKey && { sessionKey }),
```
justinformentin, line 64: Token is already passed during initOtp. Signups that include a verificationToken (which completeOtp obtains from verifyOtp). This is the only place where we should remove consumeToken.

ethankonk (approved):

PR #1433, opened 2026-07-07
Add EthSendTransactionV2 support properly

amircheikh:
## Summary & Motivation

- Synced mono. 
- Added the ability for `core` and `sdk-types` to generate version specific function overrides. 
  - This is used to generate separate `httpClient.EthSendTransaction()` and `httpClient.EthSendTransactionV2()` functions so we don't need a breaking major bump due to the intent change.
  - `sdk-server` and `sdk-browser` are major bumped since they don't have this override. 
- Replaces `TKGasStation` and `TKGasDelegate` contract addresses with the updated ones
- Examples updated accordingly

## Did you add a changeset?

YES :D


packages/react-wallet-kit/src/providers/client/Provider.tsx
```diff
@@ -5715,40 +5715,56 @@ export const ClientProvider: React.FC<ClientProviderProps> = ({
...
-                  ...(value ? { value } : {}),
-                  ...(cleanedData ? { data: cleanedData } : {}),
                   ...(nonce ? { nonce } : {}),
                   ...(gasLimit ? { gasLimit } : {}),
                   ...(maxFeePerGas ? { maxFeePerGas } : {}),
                   ...(maxPriorityFeePerGas ? { maxPriorityFeePerGas } : {}),
                 };
 
+                // V2 multi-call intent vs. legacy single-call (V1) intent.
+                // TODO (breaking change): eventually, we wont generate the v1 activity at all, remove this check and update the intent.
```
DeRauk, line 5744: Is the v1 branch still here to be backwards compatible while we roll out?

packages/sdk-browser/src/__inputs__/public_api.types.ts
```diff
@@ -4,6 +4,26 @@
...
+    post: operations["PublicApiService_EarnPositions"];
+  };
+  "/public/v1/query/earn_vaults": {
+    /** Get the catalog of all wrappable yield vaults across supported chains, enriched with live TVL and APY. Annotates which vaults the organization has already enabled. */
+    post: operations["PublicApiService_EarnVaults"];
+  };
+  "/public/v1/query/earn_withdraw_status": {
+    /** Poll the status of a withdrawal by its withdraw_request_id. */
+    post: operations["PublicApiService_EarnWithdrawStatus"];
+  };
```
DeRauk, line 26: Do you know what this `paths` list is used for? We're not quite ready to expose earn apis yet.

DeRauk (dismissed):
LGTM! Although `prettier` needs run and there's some random earn/swap stuff in here.

packages/sdk-browser/src/__inputs__/public_api.types.ts
hadrelandon: these are built of the enums and activity-type enums,  doesnt seem like any thing references them and no /submit paths so this should be ok, as long as query doesnt expose anything youd want.

DeRauk: Cool, thanks Landon!

packages/react-wallet-kit/src/providers/client/Provider.tsx
hadrelandon: yea since we dont want to do a breaking change on react-wallet-kit we can support both until we want to eventually remove `ethSendTransaction` and just support V2

hadrelandon (approved):
looks good to me, the earn stuff as mentioned by @DeRauk is feature flagged and there is only a query path built off the enums so no actual submit path yet.

PR #1440, opened 2026-07-16
fix: Broken Vaults.fiy example

janjakubnanista:
## Summary & Motivation

The `@turnkey/with-vaultsfyi` example changes introduced by #1436 broke the dependencies & the build. This PR:

- Updates the `package.json` to reflect the required dependencies
- Fixes the `build` script
- Resolves `tsc` errors from `build`
- Runs prettier

## How I Tested These Changes

CI passes we're happy

## Did you add a changeset?

Lockfile changes only

PR #1441, opened 2026-07-16
Fix vaultsfyi example lockfile path

figitaki:
## Summary
- update the lockfile importer path for the vaultsfyi example after it moved under `examples/defi`
- fix the relative workspace links for `@turnkey/sdk-server` and `@turnkey/viem`

## Root cause
Commit `740b2a4e6` moved `examples/with-vaultsfyi/package.json` to `examples/defi/with-vaultsfyi/package.json`, but `pnpm-lock.yaml` still referenced the old importer path and two-level workspace links. The mono puppeteer workflow checks out `tkhq/sdk@main` and runs `pnpm install -r --frozen-lockfile`, so the stale importer caused `ERR_PNPM_OUTDATED_LOCKFILE` before mono tests started.

## Validation
- `pnpm install -r --frozen-lockfile`

### DIFF
--- a/examples/defi/with-vaultsfyi/package.json
+++ b/examples/defi/with-vaultsfyi/package.json
@@ -3,7 +3,7 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
-    "build": "pnpm -w run build-all",
+    "build": "tsc --noEmit false",
     "setup-policy": "tsx src/setupPolicy.ts",
     "discover": "tsx src/discover.ts",
     "deposit": "tsx src/deposit.ts",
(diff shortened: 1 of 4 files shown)


figitaki:
Pushed a follow-up commit to unblock SDK CI after the lockfile fix exposed the vaultsfyi example build script.

Follow-up changes:
- changed `@turnkey/with-vaultsfyi` build from recursive `pnpm -w run build-all` to a local TypeScript build, matching adjacent DeFi examples
- aligned its `tsconfig.json` with the repo base config and added DOM libs needed by the vaults.fyi dependency types
- added explicit viem client return types and passed the wallet client chain into `sendTransaction`

Validation:
- `CI=true pnpm install -r --frozen-lockfile`
- `pnpm --filter @turnkey/with-vaultsfyi build`
- `pnpm --filter @turnkey/with-vaultsfyi typecheck`

figitaki:
Closing as superseded by #1440, which landed the vaultsfyi lockfile/build fixes on main and passed SDK js-build. Thanks!

PR #1442, opened 2026-07-17
feat(examples): add defi/atomic-batch-with-aave

cylon56:
## Summary

Adds `examples/defi/atomic-batch-with-aave`: open a **leveraged Aave v3 position and unwind it with one button press** — each side one Turnkey activity, one policy evaluation, one atomic on-chain transaction, via `ACTIVITY_TYPE_ETH_SEND_TRANSACTION_V2` with Gas Station sponsorship (EIP-7702) on Base Sepolia.

- `faucet.ts` — mints test USDC from Aave's faucet as a *sponsored* call, so the wallet needs **no ETH at any point**
- `positions.ts` — prints collateral / debt / health factor / wallet balance so the story reads in the terminal
- `enter.ts` — approve → supply 90 USDC → borrow 20 USDC: 3 calls, one atomic tx; ends with collateral posted *and* debt outstanding
- `exit.ts` — approve → repay(max) → withdraw(max): the emergency exit. The order is load-bearing (collateral can't be withdrawn while debt is outstanding), so sequential execution has a dangerous mid-unwind window — the batch is all-or-nothing
- `exit-wrong-order.ts` — proof, not claim: submits the exit with withdraw first; Turnkey's pre-flight simulation fails the batch as a unit and the position is untouched
- `createPolicy.ts` — scopes a non-root user to **exactly the DeFi interactions used here**: contract allowlist (USDC, Aave Pool, faucet) + selector allowlist (mint/approve/supply/borrow/repay/withdraw)

**Policy enforcement is per-call with all-or-nothing semantics** (verified live): the faucet/enter/exit batches pass; `USDC.transfer` is denied (allowed contract, banned selector); a call to an unlisted address is denied; and a batch mixing one valid call with one bad call is denied as a unit. The non-root user holds valid API keys yet can only faucet, enter, and exit.

README includes a sequential-vs-batch comparison (activities, policy evaluations, dashboard approvals, failure window, gas payer, broadcast path).

Uses `@turnkey/sdk-server`'s `ethSendTransaction` / `getSendTransactionStatus` (new in v7.0.0) — no hand-rolled HTTP. Follows the `with-aave` example conventions.

## Test plan

All scripts run live on Base Sepolia against real Aave v3 (addresses from `@bgd-labs/aave-address-book`):

- faucet (sponsored mint): https://sepolia.basescan.org/tx/0xd39b507f549307fd66fe4c590ef8e662956ff01935fff1f8b20c29f85616fd8d
- enter (approve + supply + borrow, one tx): https://sepolia.basescan.org/tx/0x7b021aaeb273436e927b2569c6d276dd615207870ac68c020d35323b339c379f → collateral $89.98, debt $20.00, HF 3.87
- exit-wrong-order: pre-flight simulation FAILED as expected; position verified untouched
- exit (approve + repay + withdraw, one tx): https://sepolia.basescan.org/tx/0xa3a0df62e41b613586f79e17c990026eeebdbc0388e9da2e5572853a05061e28 → flat, wallet back to ~100 USDC
- createPolicy: run verbatim against a fresh non-root user — allowed calls execute, banned selector/target/mixed batches denied by the policy engine (all cases exercised live)


examples/defi/atomic-batch-with-aave/README.md
```diff
@@ -0,0 +1,133 @@
...
+$ pnpm createPolicy
+```
+
+which installs a policy scoped to **exactly the DeFi interactions this example uses** — the USDC, Aave Pool, and faucet contracts, restricted to the flow's function selectors:
+
+```text
+condition: activity.type == 'ACTIVITY_TYPE_ETH_SEND_TRANSACTION_V2'
+        && eth.tx.from == '<SIGN_WITH>'
+        && eth.tx.to in ['<USDC>', '<AAVE_POOL>', '<FAUCET>']
+        && eth.tx.data[0..10] in ['0xc6c3bbe6', '0x095ea7b3', '0x617ba037',
```
MasterHW, line 120: calldata slicing an acceptable path here, but since goal is to be educational maybe worth calling out the generally preferred ABI upload path here (and in createPolicy.ts)

cylon56: Yeah, makes sense. I actually already did ABI-style policies in the red-button-demo but didn't fully replicate it here

package.json
```diff
@@ -61,6 +61,20 @@
...
+        "GHSA-395f-4hp3-45gv",
+        "GHSA-3jxr-9vmj-r5cp",
+        "GHSA-52cp-r559-cp3m",
+        "GHSA-8x88-c5mf-7j5w",
+        "GHSA-f88m-g3jw-g9cj",
+        "GHSA-gcfj-64vw-6mp9",
+        "GHSA-v245-v573-v5vm",
+        "GHSA-v56q-mh7h-f735",
+        "GHSA-xcpc-8h2w-3j85",
+        "GHSA-xvcm-6775-5m9r"
```
MasterHW: would prefer to see these patched in a separate PR (like #1449) than snuck in here. otherwise good to approve.

cylon56: Ah, gotcha. I was trying to fix the build issue, didn't really someone else was patching. I'll revert

MasterHW (dismissed):

TheUnseen12 (dismissed):
Looks good!

TheUnseen12 (approved):

PR #1448, opened 2026-07-22
feat: Add support for multi-signer transactions with solSendTransacti…

Bijan-Massoumi:
## Summary & Motivation

Adds `solSendTransactionV2` support for Solana transactions requiring multiple Turnkey signers.

- Exposes the v2 endpoint and types across generated SDK clients.
- Accepts ordered signer addresses through `signWiths`.
- Routes the higher-level `solSendTransaction` helper to v2 when `signWiths` is provided.
- Preserves the existing v1 path for single-signer `signWith` requests.
- Updates codegen to associate versioned operations with the correct activity result types.
- Migrates Solana examples to the v2 API.

## How I Tested These Changes

- `pnpm --filter @turnkey/core test -- sol-send-transaction-test.ts --runInBand`
- `pnpm --filter @turnkey/core typecheck`

The unit tests cover both multi-signer v2 routing and legacy single-signer v1 routing.

## Did you add a changeset?

Yes. Added `.changeset/bright-solana-signers.md` with minor bumps for the affected packages.


packages/sdk-browser/scripts/codegen.js
```diff
@@ -127,6 +150,82 @@ function extractLatestVersions(definitions) {
   return latestVersions;
 }
 
+function generateOverrideApiTypes() {
```
amircheikh: So, I know we do this for @turnkey/core but I don't think we should do this for sdk-browser or sdk-server. It's fine to make breaking bumps for those packages, that's what we did for EthSendTransactionV2 also.

sdk-browser is deprecated so we shouldn't really be making feature changes. Having a breaking version bump to change the activity intent is fine. 

sdk-server is pretty much a purely generated client, so any breaking changes to the API correspond to a breaking change to the package.


So I think we should just remove this codegen logic in sdk-server and sdk-browser and just do a breaking version bump

.changeset/bright-solana-signers.md
```diff
@@ -0,0 +1,11 @@
+---
+"@turnkey/core": minor
+"@turnkey/http": minor
+"@turnkey/sdk-browser": minor
+"@turnkey/sdk-server": minor
```
amircheikh: @turnkey/http should be major.

@turnkey/sdk-browser and @turnkey/sdk-server should be major if we remove the codegen versioning logic (other comment I left)

packages/sdk-browser/scripts/codegen.js
amircheikh: This will also force us to update any other examples using sdk-server and solsendtransaction to use the new shape which I think is good

.changeset/bright-solana-signers.md
```diff
@@ -1,11 +1,13 @@
...
+"@turnkey/sdk-server": major
 "@turnkey/sdk-types": minor
 ---
 
 Add Solana send-transaction v2 support for transactions requiring multiple Turnkey signers.
 
-The generated clients now expose separate `solSendTransaction` and `solSendTransactionV2` wrappers. Both submit to `/public/v1/submit/sol_send_transaction`; the v1 wrapper sends `ACTIVITY_TYPE_SOL_SEND_TRANSACTION` with `signWith`, while the v2 wrapper sends `ACTIVITY_TYPE_SOL_SEND_TRANSACTION_V2` with ordered `signWiths`.
+`@turnkey/core` exposes separate `solSendTransaction` and `solSendTransactionV2` wrappers. Both submit to `/public/v1/submit/sol_send_transaction`; the v1 wrapper sends `ACTIVITY_TYPE_SOL_SEND_TRANSACTION` with `signWith`, while the v2 wrapper sends `ACTIVITY_TYPE_SOL_SEND_TRANSACTION_V2` with ordered `signWiths`.
+
+The generated `@turnkey/sdk-server` and `@turnkey/sdk-browser` clients now use the v2 request and response shape through `solSendTransaction`, matching the versioning behavior of other generated activity methods.
```
amircheikh: If you have all the packages in one changeset, they'll all share the same changelog.

I think you should have http, browser and server in the same file. sdk-types should be separate

core should also be separate and I think you should also group @turnkey/react-wallet-kit and @turneky/react-native-wallet-kit into that changeset as well. Even though you didnt edit those packages directly, updating the functions in core will also update them in the wallet kits. 
If you don't give those a changeset, they'll auto resolve to a patch bump when really it should be minor bump just like core.

.changeset/bright-solana-signers.md
```diff
@@ -0,0 +1,9 @@
+---
+"@turnkey/http": major
+"@turnkey/sdk-browser": major
+"@turnkey/sdk-server": major
+---
+
+Add Solana send-transaction v2 support for transactions requiring multiple Turnkey signers.
+
+The generated `@turnkey/sdk-server` and `@turnkey/sdk-browser` clients now use the v2 request and response shape through `solSendTransaction`, matching the versioning behavior of other generated activity methods.
```
amircheikh: I don't think you need to say:
"now use the v2 request and response shape through `solSendTransaction`, matching the versioning behavior of other generated activity methods." 

since the borked versions were never released in the sdk anyways. I also think there should be a before and after of the intent shape.

Take a look at what we did for EthSendTransactionV2:

https://github.com/tkhq/sdk/blob/cd1af93c41a3f41c3c68589cfa6cfe17c1812c2f/.changeset/eighty-tires-guess.md?plain=1

.changeset/bright-solana-signers-wallet-kits.md
```diff
@@ -0,0 +1,9 @@
+---
+"@turnkey/core": minor
+"@turnkey/react-wallet-kit": minor
+"@turnkey/react-native-wallet-kit": minor
+---
+
+Add Solana multi-signer transaction support to the core client and wallet kits.
+
+`@turnkey/core` exposes separate `solSendTransaction` and `solSendTransactionV2` wrappers. Both submit to `/public/v1/submit/sol_send_transaction`; the v1 wrapper sends `ACTIVITY_TYPE_SOL_SEND_TRANSACTION` with `signWith`, while the v2 wrapper sends `ACTIVITY_TYPE_SOL_SEND_TRANSACTION_V2` with ordered `signWiths`.
```
amircheikh: Ok one thing I just realized.

`@turnkey/core` also has a `solSendTransaction` helper function in `core.ts`

The intent params for that function needs to also support both intent versions.

I did something similar for `EthSendTransactionV2`, you can take a look here: https://github.com/tkhq/sdk/pull/1433/changes

I'd also update the changeset to include this change as well as a before and after of the intent shape.
Take a look at the changeset for `EthSendTransactionV2`: https://github.com/tkhq/sdk/blob/cd1af93c41a3f41c3c68589cfa6cfe17c1812c2f/.changeset/petite-lemons-shout.md?plain=1

.changeset/bright-solana-signers-types.md
```diff
@@ -0,0 +1,5 @@
+---
+"@turnkey/sdk-types": minor
+---
+
+Add Solana send-transaction v2 intent and result types with ordered `signWiths`.
```
amircheikh: Before and after would be nice here too.

https://github.com/tkhq/sdk/blob/cd1af93c41a3f41c3c68589cfa6cfe17c1812c2f/.changeset/grumpy-pandas-tell.md?plain=1

amircheikh (approved):

PR #1461, opened 2026-07-31
react-wallet-kit: validate OAuth redirect endpoints

lreading-turnkey:
## Summary & Motivation

Require OAuth popup responses to return to the configured redirect URI before processing the authorization response. This ensures authentication results are handled only at the intended callback endpoint.

## How I Tested These Changes

I added unit tests, but I have not done a complete e2e test. If someone already has some kind of OAuth integration set up just for testing, it'd be awesome if you could help me test!

## Did you add a changeset?

Yes!

### DIFF
--- a/packages/react-wallet-kit/src/utils/oauth/config.ts
+++ b/packages/react-wallet-kit/src/utils/oauth/config.ts
@@ -13,6 +13,11 @@ export const FACEBOOK_GRAPH_URL =
 export const popupWidth = 500;
 export const popupHeight = 600;
 
+export enum OAuthResponseMode {
+  Query = "query",
+  Fragment = "fragment",
+}
+
 /**
  * OAuth provider configuration for unified OAuth flow handling
  */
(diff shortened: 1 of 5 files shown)


amircheikh:
Mind checking to see if we need this fix in react-native-wallet-kit as well?

packages/react-wallet-kit/src/utils/oauth/config.ts
```diff
@@ -13,6 +13,8 @@ export const FACEBOOK_GRAPH_URL =
 export const popupWidth = 500;
 export const popupHeight = 600;
 
+export type OAuthResponseMode = "query" | "fragment";
```
moeodeh3: non-blocking nit: make this an enum

moeodeh3 (dismissed):
LGTM - I don't think RN has the same problem because it uses `InAppBrowser.openAuth()` is an await call. We have this issue on web because we poll and that polling creates a continuous attack/error surface I think 🤔

lreading-turnkey:
> Mind checking to see if we need this fix in react-native-wallet-kit as well?

I took a look, and I tend to agree with @moeodeh3 
RN is using the platform auth-session/deep-link mechanism, which is really a different attack surface. It's validating both OAuth state and the pub-key-bound nonce. We could do a bit more as a defense in depth strategy, but I don't think it is the same type of potential issue. :smile:

moeodeh3 (approved):

amircheikh:
Have a great rest of your day fellas!

PR #1479, opened 2026-08-07
Secrets API: importSecret, exportSecret, and parallel multi-party export proposals

emostov:
## Summary

SDK support for the upcoming Secrets API.

**`@turnkey/crypto`** — `encryptSecretToBundle()` and `decryptSecretBundle()`: verify the signer quorum signature, then HPKE-encrypt/decrypt a secret payload.

**`@turnkey/sdk-server`** (`TurnkeyApiClient`)

- `importSecret({ plaintext, name?, staticProperties? })` → `secretId`
- `exportSecret({ secretId })` → plaintext, the unilateral path. Throws `EXPORT_SECRET_CONSENSUS_NEEDED` when policy requires more approvals.
- `getSecrets()` → metadata only, with `staticProperties` as a `Record<string, string>`.
- Multi-party consensus: `createExportSecretsProposal()` (local-only, computes the activity fingerprint), `submitExportSecrets()`, `awaitExportedSecrets()`.

**`@turnkey/sdk-types`** — error codes `IMPORT_SECRET_ERROR`, `EXPORT_SECRET_ERROR`, `EXPORT_SECRET_CONSENSUS_NEEDED`.

## Design notes

- Multi-agent export relies on two server-side facts: the activity fingerprint is `sha256(raw request body)`, and a duplicate-fingerprint submission from a different user attaches as an approval vote. The SDK therefore serializes the proposal body exactly once and stamps it byte-for-byte — co-signers are order-independent, and a submit race resolves by resubmitting.
- Secrets are UTF-8 strings end to end. No zeroization claims: JS can't reliably zeroize key material, so keys are hex strings dropped after use.
- `init_import_secrets` still goes through the raw `command` helper — it has no generated client method (marked internal in the swagger input; fixed on the next sync).

## Testing

- `pnpm turbo build typecheck test --filter=@turnkey/crypto --filter=@turnkey/sdk-types --filter=@turnkey/sdk-server` clean.
- Exercised end to end against a live org via the examples in #1488 — both the unilateral and two-agent consensus flows.

Companion docs PR: tkhq/docs#764

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/crypto/src/turnkey.ts
```diff
@@ -187,6 +187,170 @@ export const decryptExportBundle = async ({
...
+      : plaintext;
+
+  const targetKeyBuf = uint8ArrayFromHexString(signedData.targetPublic);
+  const encrypted = hpkeEncrypt({ plainTextBuf, targetKeyBuf });
+  plainTextBuf.fill(0);
+
+  return {
+    secretPayload: formatHpkeBuf(encrypted),
+    targetPublicKey: signedData.targetPublic,
+  };
```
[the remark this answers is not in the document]
emostov: Resolved in 93354d89d by removing zeroization from these helpers entirely: they no longer claim to zeroize inputs, so there is no success-path-only cleanup left to move. Best-effort zeroization in JS couldn't honor the contract anyway (transient copies survive).

packages/crypto/src/turnkey.ts
```diff
@@ -187,6 +187,170 @@ export const decryptExportBundle = async ({
...
+      throw new Error('missing "encappedPublic" in payload signed data');
+    }
+
+    const decryptedData = hpkeDecrypt({
+      ciphertextBuf: uint8ArrayFromHexString(signedData.ciphertext),
+      encappedKeyBuf: uint8ArrayFromHexString(signedData.encappedPublic),
+      receiverPriv:
+        typeof embeddedPrivateKey === "string"
+          ? embeddedPrivateKey
+          : uint8ArrayToHexString(embeddedPrivateKey),
```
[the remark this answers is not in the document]
emostov: Resolved in 93354d89d by dropping the zeroization guarantee instead of extending hpkeDecrypt: `embeddedPrivateKey` is now hex-string-only and the docs make no zeroization claim. A byte-based decryption path isn't worth the diff for a guarantee JS can't actually honor.

packages/crypto/src/turnkey.ts
```diff
@@ -187,6 +187,170 @@ export const decryptExportBundle = async ({
...
+ *
+ * @param {EncryptSecretToBundleParams} params - An object containing the following properties:
+ *   - plaintext {string | Uint8Array}: The secret material to encrypt.
+ *   - ingressTargetBundle {string}: A JSON-encoded ServerTargetMsg entry from InitImportSecretsResult.enclaveTargetMessages.
+ *   - organizationId {string}: The expected organization ID to verify against the signed data.
+ *   - dangerouslyOverrideSignerPublicKey {string} [Optional]: Override the default signer public key for verification. Testing only.
+ * @returns {Promise<{secretPayload: string; targetPublicKey: string}>} - The encryption-suite payload for ImportSecretParams.secretPayload and the verified target public key.
+ * @throws {Error} - If signature verification or validation fails.
+ */
+export const encryptSecretToBundle = async ({
```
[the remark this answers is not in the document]
emostov: Deferred, as noted in the PR description: round-trip fixtures need the server-side e2e flow, which isn't in a deployed environment yet. Tests land before this PR is un-drafted.

packages/sdk-server/src/sdk-client.ts
```diff
@@ -205,4 +230,359 @@ export class TurnkeyApiClient extends TurnkeyServerClient {
...
+
+    const proposal = this.createExportSecretsProposal({
+      secrets: [{ secretId: params.secretId }],
+      targetPublicKey: publicKey,
+      organizationId,
+    });
+
+    const result = await this.submitExportSecrets(proposal);
+
+    if (result.status !== "ACTIVITY_STATUS_COMPLETED") {
```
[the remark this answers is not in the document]
emostov: Fixed in 93354d89d: `exportSecret` now polls `get_activity` until the activity reaches a terminal status (configurable `timeoutMs`/`pollingIntervalMs`, defaults 60s/500ms), and still fails fast with `EXPORT_SECRET_CONSENSUS_NEEDED` when consensus is required.

packages/sdk-server/src/sdk-client.ts
```diff
@@ -205,4 +230,359 @@ export class TurnkeyApiClient extends TurnkeyServerClient {
...
+    const deadline = Date.now() + timeoutMs;
+
+    try {
+      while (Date.now() < deadline) {
+        const { activities } = await this.getActivities({
+          organizationId: proposal.organizationId,
+        });
+        const activity = activities.find(
+          (a: any) => a.fingerprint === proposal.fingerprint,
+        );
```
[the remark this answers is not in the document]
emostov: Fixed in 93354d89d: the fingerprint is resolved by walking `list_activities` pages (filtered to `ACTIVITY_TYPE_EXPORT_SECRETS`, pages of 100) until it matches, after which the activity is polled directly by ID via `get_activity`. The public API has no fingerprint filter, so pagination is the closest available primitive.

packages/sdk-server/src/sdk-client.ts
```diff
@@ -205,4 +230,359 @@ export class TurnkeyApiClient extends TurnkeyServerClient {
...
+   * activity, and decrypts the payload. If the export
+   * requires additional approvals this throws with
+   * `EXPORT_SECRET_CONSENSUS_NEEDED`; use `createExportSecretsProposal` /
+   * `submitExportSecrets` for multi-party flows instead.
+   */
+  exportSecret = async (
+    params: ExportSecretParams,
+  ): Promise<string | Uint8Array> => {
+    const organizationId = params.organizationId ?? this.config.organizationId;
+    const { privateKey, publicKey } = generateP256KeyPair();
```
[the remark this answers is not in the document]
emostov: Fixed in 93354d89d by removing the zeroization claim rather than adding a byte-based keygen: the PR no longer promises zeroization anywhere — the single-use hex key is simply dropped after use. Keeps the diff small per the design note about reusing `generateP256KeyPair`.

packages/crypto/src/turnkey.ts
```diff
@@ -187,6 +187,142 @@ export const decryptExportBundle = async ({
...
+    parsedPayload.dataSignature,
+    parsedPayload.data,
+    dangerouslyOverrideSignerPublicKey,
+  );
+  if (!verified) {
+    throw new Error(`failed to verify enclave signature on secret payload`);
+  }
+
+  const signedData = JSON.parse(
+    new TextDecoder().decode(uint8ArrayFromHexString(parsedPayload.data)),
```
figitaki, line 302: AH didn't realize this was how we parse things that come over the wire, we should absolutely be using zod for this stuff or at the minimum putting this behind a try catch but seems like this is pretty standard so not gonna block.

packages/sdk-server/src/sdk-client.ts
```diff
@@ -205,4 +230,398 @@ export class TurnkeyApiClient extends TurnkeyServerClient {
...
+          targetPublicKey: params.targetPublicKey,
+          encryptionSuite: TRANSPORT_ENCRYPTION_SUITE_ENCLAVE_ENCRYPT_V1,
+        })),
+      },
+    };
+    const body = JSON.stringify(request);
+
+    const fingerprint = `sha256:${Buffer.from(
+      sha256(new TextEncoder().encode(body)),
+    ).toString("hex")}`;
```
figitaki, line 427: what is going on here? Is this a new pattern or are we doing this manual construction of finger prints with string injection in other places?

packages/sdk-server/src/sdk-client.ts
```diff
@@ -205,4 +230,398 @@ export class TurnkeyApiClient extends TurnkeyServerClient {
...
+    const response = await fetch(url, {
+      method: "POST",
+      headers: {
+        [stamp.stampHeaderName]: stamp.stampHeaderValue,
+        "Content-Type": "application/json",
+        "X-Client-Version": VERSION,
+      },
+      body,
+      redirect: "follow",
+    });
```
figitaki, line 461: Do we construct all these fetch parameters every time instead of using a helper function? If so not blocking but another surprise finding and something worth consolidating.

packages/sdk-server/src/sdk-client.ts
```diff
@@ -205,4 +230,398 @@ export class TurnkeyApiClient extends TurnkeyServerClient {
...
+      timeoutMs = 60_000,
+      pollingIntervalMs = 500,
+    } = params;
+
+    const deadline = Date.now() + timeoutMs;
+
+    // The public API has no fingerprint filter, so first resolve the
+    // fingerprint to an activity ID by walking list_activities pages, then
+    // poll that activity directly.
+    let activityId: string | undefined;
```
figitaki: We should already have the activity id as it get's returned by `submitExportSecrets` so we could accept that as a parameter here and drop the call to list activities which is a very expensive query.

packages/sdk-server/src/sdk-client.ts
```diff
@@ -205,4 +230,398 @@ export class TurnkeyApiClient extends TurnkeyServerClient {
...
+    while (!activityId && Date.now() < deadline) {
```
figitaki, line 507: So many of these hidden `Date.now()` calls everywhere! Very scary and a bit of an antipattern imo. We should be explicit when we are defaulting to the current time.

packages/sdk-server/src/sdk-client.ts
emostov: The reason we do this is so its easier to lookup the activity without having to be the person who submitted the activity. I'm not sure if there is precedent for this in the sdk but IMO being able to calculate the fingerprint is an internal primitive, if not exernal, that we should have

Claude/Valet/Codex: please check if this is a new pattern or not

packages/crypto/src/turnkey.ts
emostov: Valet make a linear issue for this and figure out what team this should go to triage for

packages/sdk-server/src/sdk-client.ts
emostov: By be explicit do you mean doc comment or something else?

figitaki: Date.now() makes function calls non-deterministic, much harder to test that way. Better to not default and make those parameters required so the caller can determine what the pattern chould be.

packages/sdk-server/src/sdk-client.ts
```diff
@@ -205,4 +230,398 @@ export class TurnkeyApiClient extends TurnkeyServerClient {
...
+    // The export activity requires a 65-byte uncompressed target key.
+    const { privateKey, publicKeyUncompressed } = generateP256KeyPair();
+
+    const proposal = this.createExportSecretsProposal({
+      secrets: [{ secretId: params.secretId }],
+      targetPublicKey: publicKeyUncompressed,
+      organizationId,
+    });
+
+    const submitted = await this.submitExportSecrets(proposal);
```
figitaki, line 359: `submitExportSecrets` already returns an `activityId`, but `exportSecret` discards it and `awaitExportedSecrets` rediscovers the activity through a paginated `list_activities` scan. That makes a successful unilateral export depend on an expensive organization-wide query and creates a timeout path even though we already know exactly which activity to poll. Please pass `submitted.activityId` into the await path (or poll it directly), while retaining fingerprint discovery only for multi-party callers that genuinely do not have an activity ID.

figitaki (changes requested):
Blocking on the activity-ID issue in the unilateral export flow.

packages/sdk-server/src/sdk-client.ts
emostov: Fixed in 89c2815: `awaitExportedSecrets` now takes an optional `activityId`, and `exportSecret` passes `submitted.activityId` straight through, so the unilateral path polls `get_activity` directly and never touches `list_activities`. Fingerprint discovery is kept only for the multi-party case where a co-signer holds the proposal but never submitted it and so has no activity ID.

packages/sdk-server/src/sdk-client.ts
emostov: Done in 89c2815 — same fix as the thread on line 358: `exportSecret` now forwards `submitted.activityId` into `awaitExportedSecrets`, which polls it directly. `list_activities` is only walked when the caller has no activity ID (multi-party co-signer holding just the proposal).

packages/sdk-server/src/sdk-client.ts
emostov: Made `timestampMs` required on `CreateExportSecretsProposalParams` in 89c2815 — that was the only `Date.now()` hiding behind a parameter default, and it is the one that matters: the timestamp is part of the signed bytes, so a hidden default makes the fingerprint non-deterministic and can silently fork consensus. `exportSecret` still accepts an optional `timestampMs` and passes `String(Date.now())` explicitly at the call site.

I left the remaining `Date.now()` calls alone: in `awaitExportedSecrets`/`pollExportSecretsActivity` they measure elapsed time against a deadline rather than feed a value into a request, and `importSecret` builds its `timestampMs` the same way every generated activity method does. Happy to thread a clock/deadline through those too if you want them injectable for tests.

packages/sdk-server/src/sdk-client.ts
emostov: Checked: there is a helper — `TurnkeySDKClientBase.request()` in `__generated__/sdk-client-base.ts:29` builds exactly these fetch params — but it takes an object and `JSON.stringify`s it. `submitExportSecrets` has to POST `proposal.body` byte-for-byte (re-serializing changes the fingerprint and forks consensus), so it cannot use it without a variant that accepts a pre-serialized body. That file is codegen output, so I did not hand-edit it; the consolidation would be a codegen change adding something like `requestRaw(url, body: string)`. Flagging as follow-up rather than doing it in this PR.

packages/sdk-server/src/sdk-client.ts
emostov: Checked the repo as asked: client-side fingerprint construction is a new pattern here. Grepping `packages/` for `fingerprint` outside generated files, the only hits are (a) `activity.fingerprint` read off server responses, and (b) `packages/crypto/src/proof.ts`, which compares a pinned AWS root cert SHA-256 — no existing code computes an activity fingerprint locally, and nothing else string-injects a `sha256:` prefix.

The construction itself matches the server: `sha256:` + hex of SHA-256 over the exact request bytes. Since a fingerprint is what lets a non-submitter look the activity up, I agree it belongs as a primitive; if we keep it, the string-templated prefix is worth pulling into a shared helper (e.g. in `@turnkey/crypto`) rather than living inline in `sdk-client.ts`. Note the fingerprint scan is now only used by callers with no activity ID — the unilateral path polls by ID after 89c2815.

figitaki: down for this to live in a follow up pr but we should expose this as a helper for sure.

figitaki (approved):

PR #1488, opened 2026-08-13
Secrets API examples: api-key-storage and programmable-credential-access

emostov:
Stacked on #1479.

Adds two runnable examples under `examples/key-management/`, mirroring the docs solution pages:

- **`api-key-storage`** ([docs](https://docs.turnkey.com/solutions/key-management/api-key-storage)): imports an exchange API key as a secret with policy-visible static properties, creates a policy scoping retrieval of trade-only keys to a service user, lists the org's secrets, and retrieves the plaintext with `exportSecret`.
- **`programmable-credential-access`** ([docs](https://docs.turnkey.com/solutions/key-management/programmable-credential-access)): multi-agent consensus. Bootstraps `payment-agent` and `browser-agent` users with locally generated P-256 API keys, imports a credit card secret, creates a policy requiring both agents to approve any export, has both agents submit the byte-identical export proposal in parallel (with a conflict retry for the create-activity race), and decrypts on the payment-agent side only.

Both examples typecheck; the lockfile diff is hand-minimized to the two new importers and validates under `--frozen-lockfile`. Not run against a live org (requires a secrets-beta-enabled org).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/key-management/programmable-credential-access/src/index.ts
+++ b/examples/key-management/programmable-credential-access/src/index.ts
@@ -1,0 +1,164 @@
+import * as dotenv from "dotenv";
+import * as path from "path";
+import { Turnkey, TurnkeyApiClient } from "@turnkey/sdk-server";
+import { generateP256KeyPair } from "@turnkey/crypto";
+
+dotenv.config({ path: path.resolve(process.cwd(), ".env.local") });
+
+// Multi-agent consensus for credential access: a browser agent and a payment
+// agent must both sign the same export before the enclave releases a stored
+// credit card, and only the payment agent holds the decryption key.
+// See https://docs.turnkey.com/solutions/key-management/programmable-credential-access
+async function main() {
+  const organizationId = process.env.ORGANIZATION_ID!;
+  const apiBaseUrl = process.env.BASE_URL!;
+
+  const rootClient = new Turnkey({
+    apiBaseUrl,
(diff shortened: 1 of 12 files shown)


examples/key-management/programmable-credential-access/src/index.ts
```diff
@@ -0,0 +1,170 @@
...
+      const result = await agent.submitExportSecrets(proposal);
+      console.log(`${label} submitted: ${result.status}`);
+      return result;
+    } catch (_) {
+      const result = await agent.submitExportSecrets(proposal);
+      console.log(
+        `${label} submitted (after conflict retry): ${result.status}`,
+      );
+      return result;
+    }
```
figitaki, line 146: This catch retries every failure as if it were the documented create-activity race. Authentication, validation, rate-limit, and server errors will trigger an unnecessary duplicate request and may replace the useful original error with a second failure. Could we check that the caught error is the expected Turnkey conflict response before retrying, and rethrow everything else?

figitaki (approved):

examples/key-management/programmable-credential-access/src/index.ts
emostov: Addressed in #1498 — the retry is now gated on `TurnkeyRequestError` with gRPC code 6 (`ALREADY_EXISTS`, verified live: "an activity with the same fingerprint already exists"); all other errors are rethrown.

PR #1495, opened 2026-08-21
crypto: add policy-aware QOS proof verification

sergeytimoshin:
## Summary & Motivation

`@turnkey/crypto.verify()` intentionally stops at a reference-grade proof
chain. It also uses the legacy SHA-256(raw serialized manifest) binding, while
current QOS commits the semantic `VersionedManifest::manifest_hash()` in the
Nitro attestation `user_data` and PCR17.

This draft adds an additive `verifyWithQosPolicy()` API for relying parties
that already distribute independent release trust material. It:

- pins allowed semantic QOS manifest hashes from signed attestation
  `user_data` (not SHA-256 of raw Borsh bytes);
- pins PCR0 through PCR3;
- validates the QOS attestation shape (`SHA384`, module ID, timestamp, no
  nonce);
- requires the complete PCR0-31 bank;
- recomputes and verifies the live manifest/Ephemeral Key commitment in
  PCR17; and
- preserves the existing `verify()` behavior for compatibility.

It also exports `computeQosLiveManifestCommitmentPcr()` and documents that
policy values must not be learned from the proof or service being verified.
The `cbor-js` import is corrected so COSE verification works in the built ESM
package as well as in Jest/CommonJS.

The PCR17 construction mirrors
`qos_nsm::nitro::expected_manifest_commitment_pcr(ManifestCommitmentKind::Live, ...)`
from QOS 0.13.0. The test vector was cross-checked byte-for-byte against that
Rust implementation.

### Draft questions

- This proposal pins the independently approved semantic manifest hash instead
  of introducing a browser-side Borsh/JSON manifest-envelope parser. Is that
  the preferred first API, or should this PR also parse and verify
  manifest-envelope approvals?
- The existing TypeScript fixtures contain only PCR0-15 and predate the live
  PCR17 commitment. The new verifier deliberately rejects them. A current QOS
  live-attestation fixture would allow a positive end-to-end policy test in
  addition to the Rust parity vector and fail-closed legacy-fixture test.
- Should this lower-level policy API remain in `@turnkey/crypto`, or should
  known Turnkey production release policies live in a separate package?

## How I Tested These Changes

- `pnpm --filter @turnkey/crypto test --runInBand` (56 tests passed)
- `pnpm --filter @turnkey/crypto typecheck`
- `pnpm --filter @turnkey/crypto build`
- built ESM import smoke test for `verifyCoseSign1Sig()`
- Prettier check for all changed files
- cross-language PCR17 vector against `qos_nsm = 0.13.0`

## Did you add a changeset?

Yes, a minor changeset for the additive `@turnkey/crypto` API.

### DIFF
--- a/.changeset/tough-proofs-verify.md
+++ b/.changeset/tough-proofs-verify.md
@@ -1,0 +1,5 @@
+---
+"@turnkey/crypto": minor
+---
+
+Add policy-aware QOS proof verification with independently pinned manifest and PCR measurements plus the live PCR17 manifest/Ephemeral Key commitment.
--- a/packages/crypto/README.md
+++ b/packages/crypto/README.md
@@ -69,3 +69,29 @@ const event = JSON.parse(body.toString("utf8"));
 ```
 
 Do not verify a parsed and re-stringified JSON object. Even harmless-looking changes to whitespace or key ordering will change the signed payload.
+
+## Policy-aware QOS proof verification
+
+The existing `verify()` helper checks the App Proof signature, AWS Nitro
+attestation signature and certificate chain, manifest hash binding, and
+Ephemeral Key binding. Use `verifyWithQosPolicy()` when the relying party also
+has independently trusted QOS release measurements:
+
+```ts
+import { verifyWithQosPolicy } from "@turnkey/crypto";
+
+await verifyWithQosPolicy(appProof, bootProof, {
+  allowedManifestSha256: [trustedRelease.manifestSha256],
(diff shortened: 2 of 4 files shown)


emostov:
@sergeytimoshin - thank you for the contribution! To minimize work for you, we just merged an alternate to this #1512 - please let me know if that doesn't meet your needs and if so we can work on a follow up :)

emostov:
Closing for now - but feel free to comment here or reach out to @ zeke on slack

PR #1508, opened 2026-09-02
Add export-path-policies example

andrewkmin:
## Summary & Motivation

$title

This PR adds a new example (`examples/key-management/export-path-policies`) that demonstrates usage of derivation-path-based policy fields for `EXPORT_WALLET_ACCOUNT`. Specifically, we now expose the fields `wallet_account.path_indexes` (BIP32 path indexes with the hardening bit stripped) and `wallet_account.path_hardened` (per-component hardening flags).

The script creates a non-root tester user and a wallet with accounts on three derivation paths, then walks through:

1. baseline: the tester is denied with no policy in place;
2. an ALLOW policy scoped to the hardened subtree `m/8797555'/*'/3'/*'` (via `path_indexes` / `path_hardened`) allows only the in-subtree account
3. accounts _outside_ the subtree stay denied (`OUTCOME_DENY_IMPLICIT` on the path policy);
4. an explicit DENY on `path_indexes[1] == 60` wins over a broad export ALLOW (`OUTCOME_DENY_EXPLICIT`);
5. a policy referencing an unsupported field (`wallet_account.path`) is rejected at creation time.

FYI this script assumes a fresh (parent) organization. An org's existing policies would potentially change the allow/deny outcomes (the example's README calls this out).

A companion docs PR (tkhq/docs#805) documents the new fields and adds policy examples.

## How I Tested These Changes
locally

## Did you add a changeset?
N/A

### DIFF
--- a/examples/key-management/export-path-policies/src/index.ts
+++ b/examples/key-management/export-path-policies/src/index.ts
@@ -1,0 +1,194 @@
+import * as dotenv from "dotenv";
+import * as path from "path";
+import { Turnkey } from "@turnkey/sdk-server";
+import { generateP256KeyPair, decryptExportBundle } from "@turnkey/crypto";
+
+dotenv.config({ path: path.resolve(process.cwd(), ".env.local") });
+
+// Gate wallet account exports on the account's derivation path with the
+// wallet_account.path_indexes / wallet_account.path_hardened policy fields.
+// Use a fresh organization: existing policies can change the results below.
+async function main() {
+  const organizationId = process.env.ORGANIZATION_ID!;
+  const apiBaseUrl = process.env.BASE_URL!;
+  const suffix = Date.now();
+
+  const root = new Turnkey({
+    apiBaseUrl,
+    apiPublicKey: process.env.API_PUBLIC_KEY!,
+    apiPrivateKey: process.env.API_PRIVATE_KEY!,
+    defaultOrganizationId: organizationId,
+  }).apiClient();
+
+  // 0/ Create a non-root tester user. Non-root users can do nothing until a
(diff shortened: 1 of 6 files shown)


examples/key-management/export-path-policies/src/index.ts
```diff
@@ -0,0 +1,199 @@
...
+    let denied = false;
+    try {
+      await exportAccount(address);
+    } catch (error: any) {
+      denied = true;
+      console.log(`Export of ${address} denied as expected: ${error.message}`);
+    }
+    if (!denied) {
+      throw new Error(`export of ${address} should have been denied`);
+    }
```
janjakubnanista: Tiny nit

```ts
try {
  await exportAccount(address);

  throw new Error(`export of ${address} should have been denied`);
} catch (error: unknown) {
  console.log(`Export of ${address} denied as expected: ${error.message}`);
}
```

examples/key-management/export-path-policies/src/index.ts
```diff
@@ -0,0 +1,199 @@
...
+    });
+    rejected = false;
+  } catch (error: any) {
+    console.log(
+      `Policy on wallet_account.path rejected as expected: ${error.message}`,
+    );
+  }
+  if (!rejected) {
+    throw new Error("policy on wallet_account.path should have been rejected");
+  }
```
janjakubnanista: Same as the nit above - you don't need the `let rejected` and `unknown` is preferred to `any`

janjakubnanista (changes requested):

janjakubnanista (approved):

PR #1513, opened 2026-09-04
Traian/mfa scenario4 recovery

t-vila:
## Summary & Motivation

Follow-up to #1507: adds a fourth scenario covering MFA recovery, some improvements in the first three.

**Scenario 4: recovering a passkey-only account**

- Signs in / Signs up with a passkey only
- Losing it means recovering with an email OTP and Google, landing in a session scoped to CREATE_AUTHENTICATORS
- Two negative tests make the design legible: the recovery session cannot sign, and an OTP login naming no session profile falls through to the policy demanding the passkey
- The point: recovery guarded by fewer factors than the front door is just a side entrance

**Fixes to scenarios 1 to 3**

- Confirmation screen after creating policies
- Reset MFA policies and Delete this sub-organization, so scenarios can be re-run on one email
- Scenario 2 gained the sign-message step it was already claiming
- Scenario 3 lost its permanently disabled Create Wallet button, and now shows export errors inside the modal instead of behind it

**Shared code**

- scenarios/mfa.ts holds the MFA handler and the OTP login used by 2, 3 and 4, replacing ~90 duplicated lines per scenario
- The handler reads mfaStatuses instead of assuming a passkey
- formatError walks TurnkeyError.cause, where the server message lives

**Extras:**
- added a delete suborg button within the examples/authentication/oauth and otp-auth/with-backend (useful when testing)

## How I Tested These Changes

locally

## Did you add a changeset?

demo only


hadrelandon (approved):
nice 🔥

PR #1515, opened 2026-09-08
remove hpkeAuthEncrypt

moeodeh3:
## Summary & Motivation

marking this as a minor because this helper should never have been used by public sdk users and only existed for a test


packages/crypto/src/crypto.ts
hadrelandon: should remove the mention of `hpkeAuthEncrypt` here right?

hadrelandon: or make note that it is now removed. I think we need to do something with this comment

moeodeh3: nice catch, removed!

hadrelandon (approved):
🚣

PR #1529, opened 2026-09-17
examples/export-in-node: handle root quorum greater than 1

radusandor:
## Summary & Motivation

`examples/key-management/export-in-node` reads `exportResult.exportBundle` immediately after submitting the export activity. That only works when the activity completes on submission, i.e. when the organization's root quorum is 1 and no consensus policy applies to the export.

With a root quorum of 2 or more, submitting the export counts as the first vote and the activity comes back as `ACTIVITY_STATUS_CONSENSUS_NEEDED`, with the activity ID returned, so `exportBundle` is `undefined` and the example fails inside `decryptExportBundle` with:

```
Error: Error decrypting bundle: SyntaxError: "undefined" is not valid JSON
```

Nothing in that error points at approvals, so the failure is hard to diagnose.

This PR adds a `resolveExportBundle()` helper that:

- returns immediately when the bundle is already present (root quorum of 1 — behavior unchanged)
- prints the activity ID, fingerprint and current vote count, with instructions to approve from the dashboard or via the `approveActivity` endpoint
- polls `getActivity` every 5 seconds for up to 5 minutes, keeping the target key pair in memory so the bundle stays decryptable
- throws with the activity status and failure payload on `FAILED`, `REJECTED` or `AUTHENTICATORS_NEEDED`, instead of silently returning `undefined`

The export bundle can now arrive from two places: the export response itself when the activity completes straight away, or the activity result fetched once the approvals are in. The three export branches therefore also track which result field holds the bundle (`exportWalletResult`, `exportPrivateKeyResult`, `exportWalletAccountResult`).

The README gains a section on organizations with a root quorum greater than 1, including the constraint that the process has to stay alive while approvals are collected: the target private key only exists in that process, so exiting early leaves the bundle permanently undecryptable and requires a fresh export.

## How I Tested These Changes

Ran `pnpm start` against a test organization at root quorum 1, 2 and 3:

- **Quorum 1:** unchanged behavior: exports and decrypts in a single run.
- **Quorum 2 and 3:** the script waits, prints the activity ID and fingerprint, and completes once the activity is approved in the dashboard or via `approveActivity`; the export bundle decrypts correctly.

`pnpm typecheck` and `pnpm -w run prettier-all:check` pass.

## Did you add a changeset?

No. `@turnkey/example-*` is in the `ignore` list in `.changeset/config.json`, and examples are not published to npm.


t-vila (commented):
### Code review

No issues found. Checked for bugs and instruction-file compliance.

t-vila (approved):

PR #1532, opened 2026-09-21
preserve error codes when API key storage fails

t-vila:
## Summary & Motivation

On some iPhones `createApiKeyPair()` generates a key fine, then WebKit refuses to persist it to IndexedDB and throws `DataCloneError`. Making it easier to handle the device-level Keychain fault.

- `API_KEY_STORAGE_UNAVAILABLE` added to `TurnkeyErrorCodes`
- `IndexedDbStamper.createKeyPair` raises it when the store step fails, original kept  as `cause`
- the six `throw new Error(...)` sites in the auth components rethrow the original

**Any** store-step rejection maps to the new code, not a list of `DOMException` names — browsers disagree on what they throw (we've also seen `UnknownError: Connection to Indexed Database server lost`) and the remedy is the same either way.

## How I Tested These Changes

Four unit tests in `packages/core/src/__tests__/indexeddb-stamper-test.ts`. 
No new dependency — the test fakes the four IndexedDB methods `createKeyPair` touches, with `put` as the injection point.

End to end against `examples/authentication/otp-auth/without-backend`, with `IDBObjectStore.prototype.put` patched to throw `DataCloneError`:

`code: API_KEY_STORAGE_UNAVAILABLE   cause: DataCloneError: The object can not be cloned.`

That covers what the unit test can't — the code survives every intermediate `withTurnkeyErrorHandling` wrapper and the modal rethrows the original.

`core` 83 passed, `react-wallet-kit` 67 passed, `pnpm build-all` clean.

### Note for reviewers

The test stubs `@utils` with `jest.mock(..., { virtual: true })` rather than adding a `moduleNameMapper`. `poll-transaction-status-test` and `sol-send-transaction-helper-test`
already rely on `@utils` being unresolvable, so mapping it silently breaks both.
Pre-existing, but worth knowing before anyone "fixes" the alias.

## Did you add a changeset?


.changeset/neat-moles-show.md
```diff
@@ -0,0 +1,7 @@
+---
+"@turnkey/react-wallet-kit": minor
+"@turnkey/sdk-types": minor
+"@turnkey/core": minor
+---
+
+createKeyPair now raises API_KEY_STORAGE_UNAVAILABLE when the device refuses to persist a generated key pair, instead of a generic failure
```
moeodeh3: nit: wrap functions in `

moeodeh3 (dismissed):

moeodeh3 (approved):

PR #1546, opened 2026-09-22
fix(core): decode JWT payloads with base64url

kriss39:
## Summary & Motivation

**Problem.** `parseSession` and `decodeVerificationToken` both split a JWT and decode the payload segment with `atob`:

```ts
const [, payload] = token.split(".");
const decoded = JSON.parse(atob(payload));          // utils.ts:467
...
const json = atob(payloadB64);                      // utils.ts:1389
```

JWT segments are **base64url**, not standard base64. `atob` doesn't translate `-`/`_` back to `+`/`/`, so a perfectly valid, correctly signed token whose payload encoding happens to contain either character fails to decode:

1. `parseSession` — the session can't be parsed, so the token is unusable.
2. `decodeVerificationToken` — the OTP verification token can't be read. This one carries `contact`, the raw user-supplied email or phone, so the encoding is driven by user input rather than by fixed-shape fields.

The file already imports the right helper and uses it correctly a few hundred lines away in `getPublicKeyFromStampHeader`:

```ts
const stampJson = decodeBase64urlToString(stampHeaderValue);   // utils.ts:836
```

**Solution.** Use `decodeBase64urlToString` in both places. One line each, no new imports.

## How I Tested These Changes

Added `src/__tests__/jwt-base64url-test.ts`, building JWT-shaped tokens with `stringToBase64urlString` and a `contact` of `üuser@example.com` — a value that puts a `_` in the payload's base64url form.

On `main` both cases fail with `InvalidCharacterError: The string to be decoded is not correctly encoded.` With the change both decode and round-trip the payload.

`pnpm --filter @turnkey/core exec jest src/__tests__/jwt-base64url-test.ts` → 2 passed. `prettier --check` clean.

Note: the remaining `atob` at `utils.ts:1552` operates on a genuine standard-base64 string, so I left it alone.

## Did you add a changeset?

Yes — `.changeset/quiet-crabs-judge.md`, patch for `@turnkey/core`.


justinformentin:
[decodeBase64urlToString() converts the alphabet and then returns atob() directly](https://github.com/tkhq/sdk/blob/27d316ba2f1db0a9b05e2116d4b63a4258cc82b2/packages/encoding/src/base64.ts#L74-L76). atob() produces a binary/Latin-1 string, not decoded UTF-8.

The test hides that because [stringToBase64urlString() also uses btoa() directly](https://github.com/tkhq/sdk/blob/27d316ba2f1db0a9b05e2116d4b63a4258cc82b2/packages/encoding/src/base64.ts#L13-L17)

The decoder should convert the decoded bytes through TextDecoder, and the test should build its JWT using UTF-8 bytes.

kriss39:
Good catch. The decoder now turns the `atob` byte string into UTF-8 text with `TextDecoder`. I also changed the JWT fixtures to encode JSON through `TextEncoder`, so both token readers are exercised with a real multibyte payload. The direct encoding test covers invalid UTF-8 as well. Focused Jest tests, encoding typecheck, and formatting checks pass.

PR #1550, opened 2026-09-22
fix(core): issue WalletConnect EVM requests on the approved chain

Bruce039:
## Summary & Motivation

`this.ethChain` is pinned once, in the constructor:

```ts
this.ethereumNamespaces = namespaces.ethereumNamespaces;
if (this.ethereumNamespaces.length > 0) {
  this.ethChain = this.ethereumNamespaces[0]!;
}
```

and after that only an explicit `switchChain()` ever changes it. But `buildNamespaces()` sends the EVM chains as **optional** namespaces, so a connecting wallet is free to approve a subset. Pair with `ethereumNamespaces: ["eip155:1", "eip155:137"]` against a wallet that only grants Polygon, and every `personal_sign` / `eth_sendTransaction` still goes out on `eip155:1` — a chain that was never granted for the connected account — until the app happens to call `switchChain`.

The session already says which chain the account is on: `session.namespaces.eip155.accounts[0]` is a CAIP identifier like `eip155:137:0x…`. Today only the address portion is read (`getConnectedEthereum` takes `split(":")[2]`) and the chain part is discarded.

This adds a `getConnectedEthereumChain` sibling that returns the `eip155:<id>` prefix, and resolves the chain per request in `sign()`:

```ts
const ethChain = getConnectedEthereumChain(session) ?? this.ethChain;
```

Falling back to `this.ethChain` keeps behaviour unchanged when there's no session yet, and leaves `switchChain`'s own bookkeeping alone.

## How I Tested These Changes

Added `src/__tests__/wallet-connect-chain-test.ts`: a client with an established session whose account is on `eip155:137`, and a wallet configured with `["eip155:1", "eip155:137"]` — i.e. the granted chain is not the first requested one.

On `main` the assertion fails with the request going to the wrong chain:

```
-   "chain": "eip155:137",
+   "chain": "eip155:1",
```

With the change it targets `eip155:137`.

`pnpm --filter @turnkey/core exec jest src/__tests__/wallet-connect-chain-test.ts` → 1 passed. `prettier --check` clean.

Note: `buildEthProvider` reports `chainInfo.chainId` from `this.ethChain` too, so the chain surfaced through `getProviders()` has the same skew. I left that alone to keep this change to the signing path — happy to extend it if you'd like the descriptor to follow the session as well.

## Did you add a changeset?

Yes — `.changeset/thick-donkeys-shave.md`, patch for `@turnkey/core`.


justinformentin:
[sign() always prefers the chain from accounts[0]](https://github.com/tkhq/sdk/blob/48e1b8030abff83f7feb0cd4dae4479bb88b6298/packages/core/src/__wallet__/wallet-connect/base.ts#L378-L385). That list describes authorized accounts across chains; its first entry is not necessarily the wallet’s currently selected chain.

This conflicts with [switchChain(), which deliberately updates this.ethChain](https://github.com/tkhq/sdk/blob/48e1b8030abff83f7feb0cd4dae4479bb88b6298/packages/core/src/__wallet__/wallet-connect/base.ts#L315-L333). After switching to Polygon, for example, the next signature can still be sent through Ethereum mainnet if mainnet remains the first session account.

The fix should preserve this.ethChain when it is approved and only fall back to an approved session chain when the current/configured chain was not granted.

Bruce039:
Thanks for catching this. The updated resolver keeps `this.ethChain` when the session approved it and falls back to an approved account chain only when it was not granted. I added a regression test with mainnet first in `accounts`, switched to Polygon, and confirmed the next `personal_sign` targets Polygon. Both focused tests and the Prettier check pass.

PR #1551, opened 2026-09-22
fix(core): treat an unparseable stored value as absent

FlashWayne:
## Summary & Motivation

**What's wrong.** `getStorageValue` parses without a guard:

```ts
const item = browserStorage.getItem(sessionKey);
return item ? JSON.parse(item) : undefined;
```

**Why it matters.** The method's type and docs say `T | undefined` — "or undefined if not found" — and callers treat it that way. But an entry that exists and isn't valid JSON makes it throw a `SyntaxError` instead:

- `getSession` / `getActiveSession` sit directly on top of it, so the exception surfaces from every method that reads a session — `getActiveSessionOrThrowIfRequired` has no guard either.
- Nothing on the read path removes the bad entry, so once a key is corrupt every read fails again, across reloads, until something else happens to clear it.
- It's reachable without anything exotic: a quota eviction part-way through a write, a partially-written value, a future SDK version storing an incompatible shape, an extension or user touching `localStorage`, an interrupted AsyncStorage write.

**Fix.** Catch the parse failure and return `undefined` — an unusable value is no more useful than a missing one, and that's the contract the signature already advertises. Applied to both `WebStorageManager` and `MobileStorageManager`, which had the identical line.

I deliberately did not also delete the corrupt entry from the read path: `getStorageValue` is a read, and a self-healing delete would be a surprising side effect. `clearSession` already uses `removeStorageValue` directly, so it isn't blocked by a bad value.

## How I Tested These Changes

Added `src/__tests__/storage-corrupt-json-test.ts`:

- a stored `"{not json"` → resolves to `undefined`,
- a normally stored value → still round-trips.

On `main` the first rejects with `SyntaxError: Expected property name or '}' in JSON at position 1`; the second passes on both, so the happy path is untouched.

`pnpm --filter @turnkey/core exec jest src/__tests__/storage-corrupt-json-test.ts` → 2 passed. `prettier --check` clean.

## Did you add a changeset?

Yes — `.changeset/hip-donuts-shop.md`, patch for `@turnkey/core`.


justinformentin:
It would be nice to extract this into a helper to be re-used.

FlashWayne:
@justinformentin Thanks for the suggestion. I extracted the JSON parsing and fallback behavior into a shared `parseStorageValue` helper and updated both `WebStorageManager` and `MobileStorageManager` to use it in commit 9263e0d. The targeted storage regression tests pass (2/2).

PR #1554, opened 2026-09-22
fix(core): read the Solana public key before signing the stamp

FlashWayne:
## Summary & Motivation

**What's wrong.** `WalletStamper.stamp` produces the two halves of a Solana stamp from two separate reads of the wallet's active account:

```ts
signature = await this.wallet.sign(payload, provider, SignIntent.SignMessage);
...
} else if (isSolanaProvider(provider)) {
  publicKey = await this.wallet.getPublicKey(provider);
}
```

Both `sign` and `getPublicKey` resolve the account themselves (`wallet.accounts[0]`), and the gap between them contains the user's approval — seconds, not microseconds. If the account changes in that window, the stamp carries a public key that doesn't describe the key that produced the signature.

**Why it matters.** It fails closed — the server rejects a stamp whose signature doesn't verify against the embedded key — so this is a reliability bug, not an authentication one. But the failure is opaque: the user approved a signature and gets a rejected request, with nothing pointing at the account switch.

The Ethereum branch doesn't have this problem because it recovers the public key from the signature itself, so the two are bound by construction.

**Fix.** Read the Solana public key before signing and reuse that single value, so both halves of the stamp come from one read.

## How I Tested These Changes

Added `src/__tests__/wallet-stamper-solana-test.ts` with a wallet whose active account flips during `sign()` — the switch-while-approving case. It decodes the resulting stamp header and asserts the embedded `publicKey` is the account named in the signature.

On `main` it fails:

```
Expected: "signing-account-public-key"
Received: "switched-account-public-key"
```

With the change they match.

`pnpm --filter @turnkey/core exec jest src/__tests__/wallet-stamper-solana-test.ts` → 1 passed. `prettier --check` clean.

## Did you add a changeset?

Yes — `.changeset/short-pugs-tease.md`, patch for `@turnkey/core`.


justinformentin:
Two blocking issues:

1. This breaks WalletConnect’s automatic connection flow.
   The new code calls [getPublicKey() before sign()](https://github.com/tkhq/sdk/blob/58ea47c189717c8e1a201a18ec7fcdb7ff9240d3/packages/core/src/__wallet__/stamper.ts#L203-L216). For WalletConnect, sign() establishes or approves a missing session, while getPublicKey() only checks the existing session and throws if none exists. A first signing request that previously opened WalletConnect approval can now fail immediately with “No Solana account to retrieve public key.”
   
2. The account race still exists.
   The change claims both values come from one read, but they do not. The native Solana [getPublicKey() reads wallet.accounts[0]](https://github.com/tkhq/sdk/blob/58ea47c189717c8e1a201a18ec7fcdb7ff9240d3/packages/core/src/__wallet__/web/native/solana.ts#L46-L54), and then [sign() independently reads it again](https://github.com/tkhq/sdk/blob/58ea47c189717c8e1a201a18ec7fcdb7ff9240d3/packages/core/src/__wallet__/web/native/solana.ts#L164-L167). If the account switches between those calls, the stamp still pairs the wrong key and signature—just with the race moved to a different window.
The robust fix is for the Solana signing implementation to capture the account once and return both its signature and public key together. The test should also cover switching between getPublicKey() and sign(), plus the no-existing-WalletConnect-session case.

FlashWayne:
@justinformentin Thanks for catching both issues. Addressed in 21561e0: Solana stamping now uses one `signWithPublicKey` call. The native wallet captures the account passed to signing and returns its key with the signature; WalletConnect approves a missing session before signing and uses one captured address for the request and stamp. Regression tests cover the account switch between the old key lookup and signing calls, plus a first WalletConnect signing request with no session (2/2 passing).

PR #1556, opened 2026-09-22
fix(react-wallet-kit): don't pass undefined to onSuccess on disconnect

0xShadowX:
## Summary & Motivation

```ts
const matchedAccount = connectedWallets
  .flatMap((w) => w.accounts)
  .find((a) => a.address === address);

await disconnectWalletAccount(provider);
onSuccess?.("disconnect", matchedAccount!);
```

`find` returns `WalletAccount | undefined`, and the prop is declared as:

```ts
onSuccess: (type: "connect" | "disconnect", account: WalletAccount) => void;
```

The `!` is the only thing making that compile. When the lookup misses — the `wallets` state can be stale relative to `provider.connectedAddresses` — the consumer's handler is called with `undefined` in a parameter its type says is always present, and a handler that reads `account.address` throws inside the SDK's own callback.

The provider treats the same lookup as a real possibility: its equivalent path guards and throws rather than assuming a hit.

This only calls `onSuccess` when the account was actually found.

The tradeoff, since it is a behaviour change worth naming: in the stale-state case the consumer no longer gets a `"disconnect"` notification at all. I picked that over the alternative — widening `account` to `WalletAccount | undefined` — because widening forces every consumer to handle `undefined` on the `"connect"` path too, where it can't happen. If you'd rather the callback always fire and have the type reflect reality, I'm happy to switch it.

## How I Tested These Changes

`prettier --check` clean.

No test: this is in a modal screen rendered by the client provider, and the package has no component-test harness (`src/tests` holds unit tests for `utils/timers.ts` and the OAuth URL helpers). The issue is visible in the types — `Array.prototype.find` returns `T | undefined` and the prop's parameter is non-optional, with a `!` bridging the two.

## Did you add a changeset?

Yes — `.changeset/silly-seas-clap.md`, patch for `@turnkey/react-wallet-kit`.


justinformentin:
One blocking issue:

When matchedAccount is missing, the new guard skips onSuccess, which is what resolves the [handleConnectExternalWallet() promise](https://github.com/tkhq/sdk/blob/d9f05c4d46a017b3ab5d3d5c2ab211ee944d09e3/packages/react-wallet-kit/src/providers/client/Provider.tsx#L5755-L5766).

So the sequence becomes:
1. Wallet disconnects successfully.
2. matchedAccount is undefined.
3. onSuccess isn’t called.
4. The modal closes.
5. Anyone awaiting handleConnectExternalWallet() waits forever.

The better fix is to always resolve using a discriminated result:
```
type Result =
  | { type: "connect"; account: WalletAccount }
  | { type: "disconnect"; account?: WalletAccount };
  ```
That actually matches [the existing documentation](https://github.com/tkhq/sdk/blob/d9f05c4d46a017b3ab5d3d5c2ab211ee944d09e3/packages/react-wallet-kit/src/providers/client/Types.tsx#L686-L696), which already says the disconnect account may be absent.

0xShadowX:
Thanks for pointing this out. The result now distinguishes connect from disconnect: connect always includes its account, while disconnect can omit it if the local lookup misses. The callback still runs after a successful disconnect, so `handleConnectExternalWallet()` settles instead of hanging. Updated in `968de749`.

PR #1557, opened 2026-09-22
fix(react-wallet-kit): scope the stored OAuth state per attempt

0xShadowX:
## Summary & Motivation

```ts
export const OAUTH_STATE_KEY = "oauth_state";

export function storeOAuthState(state: string) {
  localStorage.setItem(OAUTH_STATE_KEY, state);
}
```

One key, for the whole origin. `localStorage` is shared across tabs, so a second OAuth attempt started before the first finished overwrites the first one's state. When the first tab comes back, `consumeOAuthState` compares its returned state against the second tab's value and throws `INVALID_OAUTH_STATE` — for a flow the user never interfered with.

Two tabs open on the app and a "Sign in with Google" click in each is enough. So is clicking sign-in, getting distracted, and clicking it again in another tab.

The file's comment anticipates the single-tab case — "the user might leave the current browsing context during the OAuth flow" — which is exactly why `localStorage` is the right choice here; it just doesn't account for two contexts being live at once.

This keys the entry by the state string itself, which is already unique per attempt:

```ts
function getOAuthStateKey(state: string): string {
  return `${OAUTH_STATE_KEY}:${state}`;
}
```

`consumeOAuthState(returnedState)` then looks up that attempt's own entry. The CSRF property is unchanged — arguably sharper: a state the SDK never issued has no entry and is rejected as before, and each attempt's entry is removed once consumed.

Two details worth calling out:

- `consumeOAuthState` still falls back to the legacy unscoped key, so a redirect that was already in flight when this version ships can complete rather than failing after the upgrade. `clearAllOAuthData` clears both shapes.
- The PKCE verifier (`${provider}_verifier`) has the same shape of problem — per provider, not per attempt — so two concurrent flows on the *same* PKCE provider still collide there. Fixing that means threading the attempt id from `buildOAuthUrl` through to completion, which felt like a separate change. Happy to follow up if you want it.

## How I Tested These Changes

Added two cases to `src/tests/oauth-test.ts`:

- two states stored before either is consumed → each validates against its own,
- a state that was never issued → still rejected with `INVALID_OAUTH_STATE`.

The first fails on `main` with `OAuth state mismatch`. All 38 pre-existing cases pass unchanged, including the one that asserts the stored state is cleared after a failed validation and the one that seeds the legacy flat key (which now exercises the fallback path).

`pnpm --filter @turnkey/react-wallet-kit exec jest src/tests/oauth-test.ts` → 40 passed. `prettier --check` clean.

## Did you add a changeset?

Yes — `.changeset/eight-cows-repeat.md`, patch for `@turnkey/react-wallet-kit`.


justinformentin:
One blocking issue:

Completing one redirect still invalidates every other concurrent flow.
[clearAllOAuthData() removes all oauth_state:* entries](https://github.com/tkhq/sdk/blob/b1cf8af6b96e21521826287050acc86914f47483/packages/react-wallet-kit/src/utils/oauth/storage.ts#L173-L181) while [completeRedirectOauth() always calls it in finally](https://github.com/tkhq/sdk/blob/b1cf8af6b96e21521826287050acc86914f47483/packages/react-wallet-kit/src/providers/client/Provider.tsx#L732-L734).

After one tab finishes signing in, the app runs clearAllOAuthData(). That function deletes the saved OAuth states for every tab, not just the one that finished.

1. Tab A starts login → saves state A.
2. Tab B starts login → saves state B.
3. Tab A returns and successfully checks state A.
4. Tab A’s cleanup deletes both A and B.
5. Tab B returns → state B is gone → login fails.

This PR fixes the “saving” part but the cleanup immediately undoes the benefit.

The test passes because it only checks states A and B directly. It doesn’t run the real cleanup that occurs after Tab A finishes. The fix is to clean up only the completed attempt’s state, while reserving “delete everything” for an explicit global reset.

0xShadowX:
Fixed in c4c7c36. `completeRedirectOauth()` now clears only the state returned by the completed redirect, leaving other tabs’ pending `oauth_state:*` entries intact. Added a regression test that completes state A and verifies state B still validates. OAuth tests pass (41), and the package typecheck passes.

